PR #175, opened 2020-04-29
chore: Integration backend test

lshaowei18:
## Solution

**Features**:

- Exports loaded app so that it can be imported when testing.
- Set up backend integration test 

## Tests

## Deploy Notes

**New dev dependencies**:

- `jest` : Test framework
- `supertest`: For testing http
- `ts-jest`: allow jest to test TS projects 
- `jest-cli`: For running jest in cli
- `@types/jest`
- `@types/supertest


jeantanzj:
Not actively working on this

PR #238, opened 2020-05-14
chore: add sign in page

lshaowei18:
## Problem

With the new landing page, we need a sign in page.

## Solution

Copied the current landing page into a new component called `SignIn` and render it on `/signin`

## Tests

- [ ] Able to log in from `/signin`
- [ ] `Guide` and `Contact` button opens new tab that has the correct URL


frontend/src/App.tsx
```diff
@@ -21,6 +22,7 @@ const App = () => {
       <AuthContextProvider>
         <Switch>
           <Route exact path="/" component={Landing}></Route>
+          <Route exact path="/signin" component={Login}></Route>
```
pregnantboy, line 25: /login

PR #279, opened 2020-05-19
Archiving

### LINKED ISSUE #110: Archiving stats after send
To reduce query time of logged/finalized messages since the table is not indexed

### LINKED ISSUE #218: Feature: Show stats for # of messages sent on landing page
Change hardcoded number to actual number of messages pulled from the DB
Screenshot 2020-05-12 at 2 49 53 PM

### LINKED ISSUE #270: Inflight messages are not reflected in sending stats
Messages that are sent out and pending response has a sent_at timestamp but without error_code or message_id, and as a result excluded from the sending stats.

With 100 recipients and send rate =1, the number of UNSENT and SENT does not tally to total.

Screenshot 2020-05-19 at 2 57 19 PM

### LINKED ISSUE #271: Resumed job only shows stats from ops table while sending
If a job is stopped and resumed, sent messages are not copied to ops table. By only pulling stats from ops table, it will only show messages that have not previously been sent out.

Screenshot 2020-05-19 at 2 57 52 PM

pregnantboy:
## Problem

- Fetching stats might become slow when number of rows in message table increases. 
- There is also no way to calculate stats once rows they deleted from message table for archival or privacy purposes.

- Inflight messages not reflected in sending stats
- Resume jobs show incorrect sent count
- Progress bar bounces  to zero when paused

Closes #110 
Closes #270 
Closes #271 

## Solution

- 3 new postgres functions
  - `update_stats(campaign_id int)` : calls one of the bottom two functions depending on type
  - `update_stats_email(campaign_id int)`: archives stats for email campaign
  - `update_stats_sms(campaign_id int)`: archives stats for sms campaign

- `statistics` table
  - columns: campaign_id, sent, errored, unsent

- `/campaign/:id/stats` endpoint fixes:
  - check delivered_at for null instead of sent_at to get unsent messages
  - When job_status is STOPPED, read from ops table instead of message table since it hasn't been written back yet
  - during sending, sent count should include archived sent count as well since already sent messages are not enqueued

- new `/stats` endpoint
  - returns global sent count (for frontend use)

**Features**:

- Endpoint to fetch stats for frontpage Closes #218 

## Before & After Screenshots

**AFTER**:

giff

## Tests

-  Check stats for sending and sent campaigns

## Deploy Notes

- Need to deploy workers
- DB migration step to archive stats for all existing campaigns
```sql
DO
$$ 
DECLARE
	rec RECORD;
BEGIN	
 	FOR rec IN SELECT * FROM job_queue ORDER BY campaign_id
	LOOP
		PERFORM update_stats(rec.campaign_id);
		RAISE NOTICE 'Finished updating for project %', rec.id;
	END LOOP ; 
END $$;
```


worker/src/core/resources/update-stats.sql
```diff
@@ -0,0 +1,14 @@
+CREATE OR REPLACE FUNCTION update_stats(selected_campaign_id int) RETURNS VOID
```
jeantanzj, line 1: resources/sql/update-stats.sql

PR #283, opened 2020-05-21
chore: Informative error for unclosed string

### LINKED ISSUE #273: Non-informative error when updating template with single quotes in attributes
Noticed this error in cloudwatch logs

Can be reproduced with single quotes in the mustached attributes
Screenshot 2020-05-19 at 4 53 09 PM

lshaowei18:
## Problem

Closes #273.

## Solution

This is currently covered in test suite, just that error message isn't informative.

## Before & After Screenshots

**BEFORE**:
82306196-51659000-99f1-11ea-83eb-e6218ee7ebcf

**AFTER**:
Screenshot 2020-05-21 at 9 29 25 AM

### DIFF
--- a/backend/src/core/services/template-client.class.ts
+++ b/backend/src/core/services/template-client.class.ts
@@ -115,6 +115,7 @@ export default class TemplateClient {
     } catch (err) {
       logger.error({ message: `${err.stack}` })
       if (err.message.includes('unclosed tag')) throw new TemplateError('There are unclosed curly brackets in the template')
+      if (err.message.includes('unclosed string')) throw new TemplateError(`Invalid character found in keyword. Only alphanumeric characters allowed.`)
       if (err.name === 'Squirrelly Error') throw new TemplateError(err.message)
       throw err
     }
--- a/backend/src/core/services/template-client.class.ts
+++ b/backend/src/core/services/template-client.class.ts
@@ -86,7 +86,7 @@ export default class TemplateClient {
           // only allow alphanumeric template, prevents code execution
           const keyHasValidChars = (key.match(/[^a-zA-Z0-9]/) === null)
           if (!keyHasValidChars) {
-            throw new TemplateError(`Invalid characters in param named {{${key}}}. Only alphanumeric characters allowed.\nKeywords like {{ Person_Name }} are not allowed, but {{ PersonName }} is allowed."`)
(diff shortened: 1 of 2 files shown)


lshaowei18:
Not sure if we should be updating the worker as well, since we are just changing the message.

backend/src/core/services/template-client.class.ts
```diff
@@ -115,6 +115,7 @@ export default class TemplateClient {
     } catch (err) {
       logger.error({ message: `${err.stack}` })
       if (err.message.includes('unclosed tag')) throw new TemplateError('There are unclosed curly brackets in the template')
+      if (err.message.includes('unclosed string')) throw new TemplateError(`Invalid character found in keyword. Only alphanumeric characters allowed.`)
```
jeantanzj, line 118: Maybe something like "Check that your keywords only contain letters and numbers. Keywords like {{ Person's Name }} are not allowed, but {{ PersonsName }} will be allowed."

backend/src/core/services/template-client.class.ts
```diff
@@ -115,6 +115,7 @@ export default class TemplateClient {
     } catch (err) {
       logger.error({ message: `${err.stack}` })
       if (err.message.includes('unclosed tag')) throw new TemplateError('There are unclosed curly brackets in the template')
```
jeantanzj, line 117: Maybe: "Check that all your keywords have double curly brackets around them. A correct example is {{ keyword }}, and incorrect ones are { keyword }, {{ keyword } or { keyword }}.  "

lshaowei18:
Hmmm looks kinda ugly without line breaks
Screenshot 2020-05-21 at 9 13 46 PM

jeantanzj:
Heyyy I think you added the updated message to the wrong error. I suggested changes for both unclosed tag and and unclosed string. 

It’s only a suggestion tho. You can rephrase if you want. I was trying to follow how eslint rules document their rules: eg https://eslint.org/docs/rules/quotes

lshaowei18:
> Heyyy I think you added the updated message to the wrong error. I suggested changes for both unclosed tag and and unclosed string.
> 
> It’s only a suggestion tho. You can rephrase if you want. I was trying to follow how eslint rules document their rules: eg https://eslint.org/docs/rules/quotes

It is pretty good actually, even better if we could have the thumbs up and thumbs down haha

PR #286, opened 2020-05-21
chore: throws 400 when double quote is misused in csv

### LINKED ISSUE #274: Investigate error in cloudwatch logs
Not sure how to reproduce
Screenshot 2020-05-19 at 5 04 57 PM

lshaowei18:
## Problem

The error occurs when a user has double quote that are not being used as `quotes`. 

In a CSV file, double quotes are used as `quotes`.

For example, `hi,my name is` will be separated into two fields. If we want it to be in one field, we have to use the double quotes: `"hi,my name is". 

However, when a unquoted field has double quotes, it results in an error because the parser doesn't know how to handle it.

Error: `404246"N` 
No error: `"404246"N"`

Closes #274.

## Solution

Initially, what I wanted to do was to silence the double quote, meaning `"hello, Tom"` will end up being `"hello, Tom"` instead of `hello, Tom`.

However, this can be confusing to users and it probably creates alot of edge cases that we have to handle.

Instead, what we will throw a 400 and display a meaningful error message.

Im currently not sure what's a good error message and whether we should display examples.

**Improvements**:

- Throws 400 when double quote is misused in a csv field

## Before & After Screenshots

**BEFORE**:
Screenshot 2020-05-21 at 12 13 24 PM

**AFTER**:
Screenshot 2020-05-21 at 12 14 40 PM

## Tests

- [x] `"hello"` does not throw an error
- [x] `hell"o` throws an error

### DIFF
--- a/backend/src/core/services/s3-client.class.ts
+++ b/backend/src/core/services/s3-client.class.ts
@@ -71,6 +71,7 @@ export default class S3Client {
     }
     catch (err) {
       if (err.message.includes('Invalid Opening Quote')) throw new UnexpectedDoubleQuoteError()
+      if (err.message.includes('Invalid Closing Quote')) throw new UnexpectedDoubleQuoteError()
       throw err
     }
   }
--- a/backend/src/core/services/s3-client.class.ts
+++ b/backend/src/core/services/s3-client.class.ts
@@ -5,7 +5,10 @@ import { isEmpty } from 'lodash'
 import config from '@core/config'
 import logger from '@core/logger'
 import { configureEndpoint } from '@core/utils/aws-endpoint'
-import { RecipientColumnMissing, UnexpectedDoubleQuoteError } from '@core/errors/s3.errors'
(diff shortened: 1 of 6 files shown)


backend/src/core/services/s3-client.class.ts
```diff
@@ -38,30 +38,36 @@ export default class S3Client {
...
+          // produces {header1: value1, header2: value2, ...}
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          params.set(row[recipientIndex!], rowWithHeaders) // Deduplication
+        }
       }
+      logger.info({ message: 'Parsing complete' })
+      return Array.from(params.values()) 
+    }
+    catch (err) {
+      if (err.message.includes('Invalid Opening Quote')) throw new UnexpectedDoubleQuoteError()
```
chinying, line 69: Check for `Invalid Closing Quote` as well.

PR #289, opened 2020-05-21
fix: singlerow textarea scrolls with multiple lines

pregnantboy:
## Problem

TextArea component with singleRow flag shows ugly scrollbar when user presses enter or text wraps.

## Solution

Allow TextArea to resize when text wraps but prevent line breaks and `Enter` keypress.

Screenshot 2020-05-21 at 5 52 11 PM

### DIFF
--- a/frontend/src/components/common/text-area/TextArea.tsx
+++ b/frontend/src/components/common/text-area/TextArea.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useRef, useEffect } from 'react'
 import cx from 'classnames'
 import TextareaAutosize from 'react-textarea-autosize'
 import escapeHTML from 'escape-html'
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/text-area/TextArea.tsx
```diff
@@ -11,6 +11,23 @@ const TextArea = ({ highlight, singleRow, placeholder, value, onChange }:
...
+  useEffect(() => {
+    // ignores Enter key presses as it triggers a resize on textarea
+    if (singleRow && textAreaRef.current) {
+      textAreaRef.current.onkeydown = (e) => {
+        if (e.keyCode === 13) {
+          e.preventDefault()
+        }
+      }
+    }
+  }, [textAreaRef.current, singleRow])
```
dvrylc, line 25: hmm since we're already preventing line breaks in the `onChange` event, could we do without this?

pregnantboy: it's a workaround for the weird behaviour exhibited by the autosize textarea library. Even if line breaks are removed, pressing enter will trigger a resize and the height will increase until another key is pressed.

Issue #284, opened 2020-05-21
Set up local storage option for uploading files


jeantanzj:
**Is your feature request related to a problem? Please describe.**

No alternative to S3 currently exists for local development. We cannot test POST /template without an S3 bucket. 

**Describe the solution you'd like**
Set up local storage option for uploading files. Chinying suggested minio

LoneRifle:
I've done some homework, and think that @localstack might be a great drop-in replacement to AWS for local development purposes. The only tweaks we would need to do to the codebase would be to allow the insertion of the `endpoint` into relevant points in the codebase (notably where `awsRegion` comes up).

Note that localstack does not expose ECS APIs

PR #293, opened 2020-05-22
refactor(backend): allow aws endpoint override

LoneRifle:
## Problem

To allow dev work to be self-contained, we need to remove the hard
dependency on hosted services like AWS at runtime. The first step to
this requires us to accept a parameter - an explicit RESTful endpoint -
that will be injected into Postman's AWS JavaScript clients, overriding
`AWS_REGION` in the process.

Doing so will allow developers to get Postman to connect to a specified
mock AWS service when running on their own machines, and would also
help with more invasive testing. This would also allow for external devs
to contribute, since they no longer need to have an AWS account to work
on Postman.

Note that this does not allow overrides for workers, as frameworks that
mock AWS services (eg localstack) do not have a freely available ECS API

Prep work for #284

## Solution

- Add a new parameter in config - `aws.awsEndpoint`, env `AWS_ENDPOINT`
- Create a new utility function - `configureEndpoint()` - that handles
  selection of either endpoint or region for AWS config purposes
- Make use of `configureEndpoint()` in cloudwatch, S3 and SecretsManager
  - change the way we configure WinstonCloudwatch, by specifying
    `awsOptions` instead of `awsRegion`

## Tests

Confirm functionality by running on dev machine without and with `AWS_ENDPOINT` set

## Deploy Notes

N/A - for local dev testing

- `AWS_ENDPOINT` : if specified, overrides `AWS_REGION`. The RESTful endpoint to send all AWS requests

**New scripts**:

- @core/utils/aws-endpoint - used to choose between endpoint and region at runtime

### DIFF
--- a/backend/src/core/config.ts
+++ b/backend/src/core/config.ts
@@ -51,6 +51,11 @@ const config = convict({
       default:  'ap-northeast-1',
       env: 'AWS_REGION',
     },
+    awsEndpoint: {
+      doc: 'The endpoint to send AWS requests to. If not specified, a default one is made with AWS_REGION',
+      default: null,
+      env: 'AWS_ENDPOINT',
+    },
     logGroupName: {
       doc: '	Name of Cloudwatch log group to write application logs to',
       default: 'postmangovsg-beanstalk-prod',
--- a/backend/src/core/utils/aws-endpoint.ts
+++ b/backend/src/core/utils/aws-endpoint.ts
@@ -1,0 +1,19 @@
+import convict from 'convict'
+
+interface AWSEndpointConfig { aws: { awsEndpoint: any, awsRegion: any } }
+
+/**
+ * Provide an Object containing either the endpoint to send AWS requests to,
+ * or if that is absent, the region from which to generate the endpoint. 
+ * This Object is to be injected into AWS JavaScript clients.
+ * 
+ * @param config - a convict object whose schema includes AWSEndpointConfig
+ * @return an object containing either the endpoint or region field
+ */
+const configureEndpoint = <T extends AWSEndpointConfig>(config: convict.Config<T>) => {
+  const endpoint = config.get('aws.awsEndpoint')
+  const region = config.get('aws.awsRegion')
+  return endpoint ? { endpoint } : { region }
+}
+
+export { configureEndpoint } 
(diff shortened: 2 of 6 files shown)


backend/src/core/utils/aws-endpoint.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Provide an Object containing either the endpoint to send AWS requests to,
+ * or if that is absent, the region from which to generate the endpoint. 
+ * This Object is to be injected into AWS JavaScript clients.
+ * 
+ * @param config - a convict object whose schema includes AWSEndpointConfig
+ * @return an object containing either the endpoint or region field
+ */
+const configureEndpoint = <T extends AWSEndpointConfig>(config: convict.Config<T>) => {
```
jeantanzj, line 13: why did you choose to pass the config in, instead of importing it?

backend/src/core/config.ts
```diff
@@ -51,6 +51,11 @@ const config = convict({
       default:  'ap-northeast-1',
       env: 'AWS_REGION',
     },
+    awsEndpoint: {
```
jeantanzj, line 54: I was reading the docs for secrets manager, and cloudwatch. The endpoint property is expected to be a string like 'https://{service}.{region}.amazonaws.com'. Since I'm not familiar with localstack- does this mean we should be adding multiple, specific `awsEndpoint`  for each service here? 

References:
https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SecretsManager.html#constructor-property
https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CloudWatchLogs.html#constructor-property
https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#constructor-property

jeantanzj:
Built on staging. Everything still works

backend/src/core/config.ts
LoneRifle: Not with localstack, or at least, not with _current_ versions of localstack; one endpoint will offer all the services.

backend/src/core/utils/aws-endpoint.ts
LoneRifle: passing it as an argument makes it more explicit that the purpose of the function is to derive something from the config. It also grants the user more control, so that if needed, the config could be changed before calling the function. Lastly, it also helps somewhat with testing, as there's one less thing to worry about when mocking config

PR #301, opened 2020-05-26
chore: Integrate Prettier

dvrylc:
## Problem

Right now we use a variety of coding patterns and styles, some of which make the project more difficult to understand than necessary. Standardizing the code makes the project friendlier to newcomers, especially relevant now that we're open source.

## Solution

This PR enforces Prettier in addition to some ESLint rules across the codebase.

### Notes for reviewer

- There is no shared config – Backend, frontend and worker each have their own ESLint and Prettier setup. I did consider a unified setup at the root level but decided against it because:
    - The ESLint configs are not identical across backend, frontend and worker
    - A unified setup would increase coupling across the repo, making it difficult to move away from a monorepo structure if we ever decide to
    - This is what we already had and aside from less `devDependencies` to install overall, there wasn't much of a reason to change
- The root level has its own Prettier setup (no ESLint)
    - This is for linting the YAML and Markdown files we have at the root level and in `/docs`
    - The root-level Prettier is setup to ignore `/backend`, `/frontend` and `/worker` (refer to the root `.prettierignore`) – This ensures that we are not double-linting at both the root and project level. `/serverless` is also ignored for now. 
- **Majority of this PR is auto-fixed code, the actual changes that need to be reviewed are:**
    - Backend, frontend, worker and root – ESLint config and updated npm scripts
    - Frontend only – Refactored some functions in edf0e5a6ca0b6aa4b7cee551e01776847ec000b7 – The identity of these functions were changing on every render, causing unnecessary `useEffect()` triggers. These warnings were brought up by `react-hooks/rules-of-hooks`.
    - Updated Travis config – Previously only the frontend lint ran on Travis, but I've set it up to lint everything including the root ([sample build](https://travis-ci.com/github/opengovsg/postmangovsg/builds/168088860)). 

## Tests

1. Write some lint-violating code in backend, frontend, worker and root – Each `npm run lint` should attempt to auto-fix or throw an error. I recommend using tsc violations because Prettier violations tend to be auto-fixable ([some ideas](https://github.com/opengovsg/postmangovsg/commit/7307767fbb977aace63f828cadd6d82bd2084b84)). 
2. From here, checkout a new branch and use `git commit --no-verify` to commit and push – Travis should scream at you. Check that each individual Travis job fails with the right set of failures ([sample build](https://travis-ci.com/github/opengovsg/postmangovsg/builds/168137798)). 

## Deploy Notes

New dev dependencies:

- `prettier`
- `eslint-config-prettier`
- `eslint-plugin-prettier`
- `husky`
- `lint-staged`


dvrylc:
> Linting occurs on travis, and since `eslint --fix` is able to fix the issue without manual intervention, it doesn't complain. 

Thanks for catching this, made the necessary changes. 

I chose to leave `lint` as the auto-fixing version because I believe that's what most of us are used to while developing. The no-fix versions are `lint-no-fix` and I've updated the Travis config to call those instead. Wherever possible (backend and worker), the auto-fix scripts are simply `npm run lint-no-fix -- --fix`. 

Regarding more descriptive Prettier errors – 
Prettier doesn't seem to be capable of this because of its internal implementation, which is also why the ESLint Prettier errors are like `Delete ;` instead of actual reasons – because it's just diff output. Don't think there's much we can do here, sadly. 

Also side note –
Found a bug in the Prettier globs – the previous globs were not surrounded by quotes and this resulted in them being expanded by the shell instead of by Prettier ([first yellow box here](https://prettier.io/docs/en/cli.html)), resulting in some files being missed out. Surrounded the globs with quotes and amended all the auto-fix commits to include the missing files.

frontend/src/components/dashboard/create/Create.tsx
```diff
@@ -16,17 +22,17 @@ const Create = () => {
...
       setCampaign(campaign)
       setLoading(false)
     }
-  }
+  }, [id])
 
   useEffect(() => {
     loadProject()
-  }, [id])
+  }, [loadProject])
```
pregnantboy, line 35: can we just use useEffect with id as dependency, and loadProject to take in an id as arg 
useCallback is more appropriate when function is passed to a child component as a prop and you dont want it to rerender each time cos of referential inequality.

pregnantboy: same goes for the other useCallback changes

PR #330, opened 2020-06-01
Generate download link for invalid recipients list

### LINKED ISSUE #302: Download list of invalid recipients for sms
Users want to see the state of their sms messages.

After #211 is released, we can create a read replica and an endpoint that can pull the error_code / received_at / updated_at for the campaign's messages,  returning these as a csv. 
But there's a replication lag so we have to set expectations right.

miazima:
## Problem

User wants to retrieve recipients of all failed messages sent out.

Closes #302

## Solution

- use a read replica db to query as we do not want to overwhelm master db
- implement endpoint to retrieve recipients for a campaign with invalid or error message delivery status
- display invalid recipients download link on campaigns page and SMSDetail, EmailDetail page

## Before & After Screenshots

image
image

## Deploy Notes

- `DB_READ_REPLICA_URI` : URI for read replica db


backend/src/core/services/stats.service.ts
```diff
@@ -84,7 +86,25 @@ const getTotalSentCount = async (): Promise<number> => {
...
+ * Helper method to get error_code, recipients for failed messages from logs table
+ * @param campaignId
+ * @param logsTable
+ */
+const getInvalidRecipients = async (campaignId: number, logsTable: string): Promise<Array<CampaignInvalidRecipient> | undefined> => {
+  // Get read replica instance
+  const sequelize = sequelizeLoader.getSequelizeReadReplicaInstance()
+  // Retrieve message logs with error codes from logs table
+  return await sequelize?.query(
+    `SELECT recipient, sent_at, updated_at, message_id, error_code FROM ${logsTable} \
```
lshaowei18, line 99: Hmmm would it be possible to replace `${logsTable}` with `:logsTable`?

lshaowei18: Also Im just curious if it is possible to use sequelize models instead of raw query to read replica db?

miazima: Including `logsTable` in replacements quotes the string (cos it escapes the string from what I read), hence resulting in syntax error in the query. This seems to be an unresolved issue (see https://github.com/sequelize/sequelize/issues/4494). I also tried `Sequelize.literal(logTable)` to inject the value but that gave an invalid literal value error.

Yet to find a workaround, open for suggestions!

miazima: From what I read, I would need to duplicate the models in order to associate them to another db. Since it's a single query I thought using a raw query would be cleaner.

lshaowei18: > Including `logsTable` in replacements quotes the string (cos it escapes the string from what I read), hence resulting in syntax error in the query. This seems to be an unresolved issue (see [sequelize/sequelize#4494](https://github.com/sequelize/sequelize/issues/4494)). I also tried `Sequelize.literal(logTable)` to inject the value but that gave an invalid literal value error.
> 
> Yet to find a workaround, open for suggestions!

Ah thanks for the explanation. Don't have any good suggestions for now haha

lshaowei18: > From what I read, I would need to duplicate the models in order to associate them to another db. Since it's a single query I thought using a raw query would be cleaner.

Hmmm ok maybe eventually when we have more queries on the read db we can consider models

jeantanzj:
Sorry I've been dragging my feet on this because I'm not sure how we should be handling the callback statuses. Should a message be considered successful only when the "DELIVERED" status is returned? 

I'm also a bit uncomfortable with how the sequelize loader has been implemented. Maybe we should have a dbLoader and a replicaLoader? Honestly I dont know what the best thing to do is, at this point. Will review this within the week.

miazima:
Yeah I came across the sequelize read replica implementation, I didn't do it that way as my understanding was we want to this read replica only for this particular query and the rest of the app reads from the master db. If we are not concerned about the lag time (and its fluctuating nature) for other read queries, we can use this. Not sure if my understanding on the scope of the usage of the read replica is correct?

jeantanzj:
https://sequelize.org/master/class/lib/sequelize.js~Sequelize.html
`options.useMaster - Force the query to use the write pool, regardless of the query type.`

miazima:
> https://sequelize.org/master/class/lib/sequelize.js~Sequelize.html
> `options.useMaster - Force the query to use the write pool, regardless of the query type.`

Ah missed this option, okay will go ahead with this implementation

jeantanzj:
Screenshot 2020-06-18 at 12 48 03 PM
The lag seems minimal. Maybe we can just use the read replica for most queries?

You'll have to set useMaster to true for the sequelize prepared statement queries. I got `"SequelizeDatabaseError: cannot execute INSERT in a read-only transaction\n` when trying to send a campaign, because `SELECT insert_jobs(id);` is actually an insert operation.

jeantanzj:
As discussed, most of the team thinks that it's not a good idea to offload all the reads to a read replica. 

Try setting useMaster to true [by default](https://github.com/opengovsg/postmangovsg/compare/feat/invalid-recipients-list...feat/invalid-recipients-list-2)

backend/src/core/loaders/sequelize.loader.ts
```diff
@@ -20,17 +21,25 @@ import { SmsMessage, SmsTemplate, SmsOp } from '@sms/models'
...
   const dialectOptions = config.get('IS_PROD')
     ? config.get('database.dialectOptions')
     : {}
-  const sequelize = new Sequelize(DB_URI, {
+  const sequelize = new Sequelize({
     dialect: 'postgres',
     logging: false,
     pool: config.get('database.poolOptions'),
+    replication: {
+      read: [parse(DB_READ_REPLICA_URI)],
```
jeantanzj, line 35: I believe you need to set "username" for this. I couldn't get it to work on staging. ie parse should return `{auth, username, host, port, database}` instead of `{auth, user, host, port, database}`

frontend/src/components/common/action-button/ActionButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const ActionButton = (props: any) => {
+  const { className, children, ...otherProps } = props
+  const [toggleDropdown, setToggleDropdown] = useState(false)
+
+  function handleToggleDropdown(
+    event: React.MouseEvent<HTMLDivElement, MouseEvent>
+  ) {
+    event.stopPropagation()
+    setToggleDropdown(!toggleDropdown)
```
xming13, line 14: Hi @miazima, I think using functional update show clearer intent as the value to update is based on the previous state.
`setToggleDropdown(toggleDropdown => !toggleDropdown)`

reference: https://joaoforja.com/blog/why_use_usestate_functional_update_even_when_everything_seems_to_work/

PR #331, opened 2020-06-01
refactor: num_recipients part of campaign details, fix types

pregnantboy:
## Problem

campaign details apis are return campaign object and `num_recipients`

## Solution

unnest object and return `num_recipients` as part of object


jeantanzj:
Pending #337

backend/src/email/services/email.service.ts
```diff
@@ -110,30 +110,34 @@ const setCampaignCredential = (campaignId: number): Promise<[number, Campaign[]]
...
+        include: [
+          {
+            model: JobQueue,
+            attributes: ['status', ['created_at', 'sent_at']],
+          },
+          {
+            model: EmailTemplate,
+            attributes: ['body', 'subject', 'params', 'reply_to'],
+          }],
+      }),
```
jeantanzj, line 131: why was `.get({ plain: true })` moved to the return statement?

pregnantboy: it has to be called on the result not the promise

PR #332, opened 2020-06-01
fix: messages with message_id and error_code double counted

pregnantboy:
## Problem

The `error_code` column is not overwritten on successful retry, due to coalescing of error_code from ops and message table. This means that successfully retried messages adds to the errored count as well.

## Solution
- Set `error_code`, `delivered_at` and `sent_at` on enqueue
- Errored count should not include message with `message_id`

### DIFF
--- a/worker/src/email/resources/sql/enqueue-messages-email.sql
+++ b/worker/src/email/resources/sql/enqueue-messages-email.sql
@@ -9,7 +9,7 @@ BEGIN
 	WITH messages AS 
   (
     UPDATE email_messages m 
-    SET dequeued_at = clock_timestamp(), updated_at = clock_timestamp(), error_code = NULL, delivered_at = NULL, sent_at = NULL
+    SET dequeued_at = clock_timestamp(), updated_at = clock_timestamp(), error_code = NULL, delivered_at = NULL, sent_at = NULL, received_at = NULL
     WHERE m.campaign_id = selected_campaign_id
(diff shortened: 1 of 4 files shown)


worker/src/email/resources/sql/enqueue-messages-email.sql
```diff
@@ -7,20 +7,23 @@ BEGIN
...
-	RETURNING 
-	id,
-	campaign_id, 
-	recipient, 
-	params, 
-	dequeued_at,
-	created_at,
-	updated_at )
+  (
+    UPDATE email_messages m
```
jeantanzj, line 11: Why didn’t you set received at to null as well?
I think all status related columns should be set to null if we retrying

worker/src/email/resources/sql/update-stats-email.sql
```diff
@@ -7,7 +7,7 @@ BEGIN
 WITH stats AS (
   SELECT
     COUNT(*) FILTER (WHERE delivered_at IS NULL) AS unsent,
-    COUNT(*) FILTER (WHERE error_code IS NOT NULL) AS errored,
+    COUNT(*) FILTER (WHERE error_code IS NOT NULL AND message_id IS NULL) AS errored,
```
jeantanzj, line 10: Could you add a comment to this to explain the scenario where both of them can be not null? I.e. when a message is processed by the messaging provider but the callback afterwards indicates failure to send

Issue #320, opened 2020-05-27
Add sentry.io


jeantanzj:
**Is your feature request related to a problem? Please describe.**
I think we've got a team wide subscription to sentry. Might as well add it to track if there are issues on client-side

**Describe the solution you'd like**
Add sentry.

PR #357, opened 2020-06-03
Integrate Sentry on frontend

miazima:
## Problem

Track client side errors through Sentry

Implements frontend error monitoring for #320

## Solution

Integrate Sentry in frontend.

## Deploy Notes

- `REACT_APP_SENTRY_DSN` : Sentry dsn key, explicitly defined as `SENTRY_DSN_PRODUCTION` and `SENTRY_DSN_STAGING` in `amplify.yml`
- `REACT_APP_SENTRY_RELEASE`: A string that represents the current release version, derived from `sentry-cli releases propose-version`

**New dependencies**:

- `@sentry/browser`: Sentry SDK for browsers
- `@sentry/cli`: Sentry CLI

**New scripts**:

- `get-sentry-release`: runs sentry cli to get the proposed release version
- `upload-source-map`: runs `sentry.js` to upload source maps to sentry


amplify.yml
```diff
@@ -20,6 +20,7 @@ frontend:
     preBuild:
       commands:
         - if [ "${AWS_BRANCH}" = "master" ]; then export REACT_APP_BACKEND_URL=${BACKEND_URL_PRODUCTION}; else export REACT_APP_BACKEND_URL=${BACKEND_URL_STAGING}; fi
+        - if [ "${AWS_BRANCH}" = "master" ]; then export REACT_APP_SENTRY_DSN=${SENTRY_DSN_PRODUCTION}; else export REACT_APP_SENTRY_DSN=${SENTRY_DSN_STAGING}; fi
```
miazima, line 23: I need to include the Sentry DSN in the csp headers. This would require exposing `SENTRY_DSN_PRODUCTION` and `SENTRY_DSN_STAGING` as I was't able to use `${REACT_APP_SENTRY_DSN}` to the Content-Security-Policy headers (see https://github.com/aws-amplify/amplify-console/issues/644).

The Sentry DSN key isn't a secret so it isn't critical if it is exposed. (see https://forum.sentry.io/t/dsn-private-public/6297)

amplify.yml
```diff
@@ -20,11 +23,15 @@ frontend:
...
```
jeantanzj, line 26: Can't these be part of the same if else statement  in line 25?

amplify.yml
```diff
@@ -48,8 +61,13 @@ frontend:
...
           value: "same-origin"
         - key: "Content-Security-Policy"
-          value:
-            "connect-src 'self' https://*.postman.gov.sg https://postmangovsg-dev-upload.s3.ap-northeast-1.amazonaws.com/ https://postmangovsg-prod-upload.s3.ap-northeast-1.amazonaws.com https://www.google-analytics.com;
+          value: "connect-src 'self' \
+            https://*.postman.gov.sg https://postmangovsg-dev-upload.s3.ap-northeast-1.amazonaws.com/ \
+            https://postmangovsg-prod-upload.s3.ap-northeast-1.amazonaws.com \
+            https://o399364.ingest.sentry.io/;
+            report-uri \
+            https://o399364.ingest.sentry.io/api/5261015/security/?sentry_key=06ac5522b7354608a16912921d15ef4e \
```
jeantanzj, line 69: why are there two report uri

miazima: There are 2 security header endpoints for staging and production each. This allows us to capture any CSP violations (https://docs.sentry.io/error-reporting/security-policy-reporting/#content-security-policy)

jeantanzj: Yea but now a csp violation on staging gets reported on production as well.  
Btw why do we have 1 project for prod and 1 project for staging instead of 1 project and 2 environments? https://docs.sentry.io/enriching-error-data/environments/

jeantanzj: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri
Maybe we should just remove it. It's very deprecated. 

I attempted to set the uri using the sentry_dsn env var in amplify.yml, but it didn't resolve. report-uri also isn't allowed in the meta tags. Can't see any way around this

PR #374, opened 2020-06-04
feat: Email bounces and delivery status

### LINKED ISSUE #303: Handle bounces and delivery status for emails
Users want to see which emails bounced / were delivered / were read

We need to handle bounces (https://aws.amazon.com/blogs/messaging-and-targeting/handling-bounces-and-complaints/), complaints and delivery statuses. 

Then allow users to download these invalid email addresses, similar to #302

lshaowei18:
## Problem

Closes  #303.

## Solution

The workflow is : 

SES -> SNS -> Lambda -> Database

Since SNS can trigger the lambda directly, the lambda is not on API gateway.

We have to enable original headers so that we can make use of the original message id to identify which message the notification is for.

Screenshot 2020-06-05 at 9 29 54 AM

Currently, we handle 2 types of notifications:
- Successful delivery: Updates received_at in `email_messages`
- Bounce: Updates error_code in `email_messages`

**Features**:

- Updates the database for the successful delivery, bounces and rejects. 

## Tests

- [ ]  bounce notification updates error_code : bounce@simulator.amazonses.com
- [ ] delivery notification updates received_at

## Deploy Notes

_Notes regarding deployment of the contained body of work. These should note any
new dependencies, new scripts, etc._

- `SERVERLESS_EMAIL_DB_STAGING_URL` : url for staging DB
- [ ] `SERVERLESS_EMAIL_DB_PROD_URL`: url for prod DB (Have not added)

## TODOs:
- [ ] Update travis for production deployment
- [ ] Set up db user for production DB
- [ ] Add env var for production 
- [ ] Connect production lambda to SNS
- [ ] Remove SNS -> staging lambda connection


lshaowei18:
Not sure if we should be verifying the AWS signature of the incoming request. 

As far as I know, the lambda is not exposed to the internet, only SNS will be calling it. However the SNS is in a different region, and I am not sure if the request might be intercepted and the content might change.

I have already written the signature code, so shouldn't be too much additional work. Do let me know what you guys think!

jeantanzj:
why do we need two sns topics?

lshaowei18:
> why do we need two sns topics?

We have 3 sns topics for SES now, it was already there when i first visited SES so it might be the default. I believe we can customize with setting up configuration sets, but i have not explored enough to tell if it will affect sending of emails.

Configuration set
`An Amazon SES construct that encapsulates where you want to publish email sending events, and what email sending events you want to publish. When you send an email that you want to use with event publishing, you specify the configuration set to associate with the email.`

I was thinking configuration set is not a priority for now if we are not going to introduce open/click notifications, but happy to explore more.

lshaowei18:
So I have changed SES to 1 topic, `email-status` and `log-email-sns` is subscribed to the topic. 

#400 needs to be merged in first before this since the 4 commits that i added today depends on that. 

Do note that the db role created for the lambda does not have permissions to the black list table.

serverless/log-email-sns/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+        replacements: { recipientEmail, messageId }, type: QueryTypes.UPDATE 
+      })
+  }
+  else {
+    errorCode = "Soft bounce, Amazon SES fails to deliver the email after retrying for a period of time."
+  }
+  
+  console.log('Updating email_messages table.')
+  await dbConnection.query(
+    `UPDATE email_messages SET error_code=:errorCode, received_at=:timeStamp, status='INVALID_RECIPIENT', updated_at = clock_timestamp() WHERE message_id=:messageId`,
```
lshaowei18, line 96: So for soft bounce and permanent bounce i updated the status as `INVALID_RECIPIENT` because I vaguely remember that we will allow resends for `ERROR`. Tbh i am not sure soft bounces should be allowed to resend.

PR #390, opened 2020-06-08
Integrate Sentry on backend

miazima:
## Problem

Track server side errors through Sentry

Implements backend error monitoring for #320

## Solution

Integrate Sentry on the backend

## Deploy Notes

- `REACT_APP_SENTRY_DSN` : Sentry dsn key

**New dependencies**:

- `@sentry/node` : Sentry SDK for nodejs


backend/src/core/loaders/express.loader.ts
```diff
@@ -54,7 +72,14 @@ const expressApp = ({ app }: { app: express.Application }): void => {
     return res.sendStatus(200)
   })
 
+  app.get('/debug-sentry', function mainHandler() {
```
chinying, line 75: Used for testing, to remove before deploying

chinying:
Tests run:
* Not authenticated, hit /debug-sentry error route, event was logged.
* Authenticated with cookie (with frontend), modified a route to deliberately throw an error, Sentry exception got logged
* Authenticated with API key, hit endpoint directly, Sentry exception logged

PR #391, opened 2020-06-09
refactor: Sharding out template logic into local node module

### LINKED ISSUE #363: [Registered] Shard out templating logic
- Move templating logic into it's own local node module for better consistency
- Frontend, backend and workers would npm install this module

lshaowei18:
## Problem

Closes #363.

## Solution

**TemplateClient**
Moved `TemplateClient` into a local node module called `postman-templating`. 

`TemplateClient` no longer has an exported function called `testHydration`, as I felt that `testHydration` was doing too many things and doesn't fall into `TemplateClient's` scope. 

- Parse the csv file

**parseCsv**
I moved this function into `postman-templating` and it can be imported by `import { parseCsv } from `postman-templating`. I didn't put it inside `TemplateClient` as I felt that it is slightly cleaner to exist as a single function. 

This is so that frontend can also use the same function as backend. 

**Improvements**:

- Moved `TemplateClient` and `parseCsv` into a local node module called `postman-templating`
- Refactor `testHydration` into smaller functions
- Backend & worker now uses `postman-templating` module.
- Some of the errors related to templating is now imported from `postman-templating`

## Tests

## TODOs:
- [ ] Move `testHydration` into core `TemplateService` (Not sure about this yet)
- [ ] `npm install` `postman-templating` to make deploying neater (Not sure about this, too)


.travis.yml
```diff
@@ -20,7 +20,11 @@ jobs:
         - cd backend
       install:
         - npm install
+        - cd node_modules/postman-templating
+        - npm install
+        - npm run test
       script:
+        - cd ../../
```
lshaowei18, line 27: I don't really like this portion; makes our deployment process harder to understand. I have to run `npm install` on `node_modules/postman-templating` because it seems like npm install creates a sym link to the folder for local modules. Perhaps I can `npm install` on the folder before I run `npm install` on backend and worker.

backend/package-lock.json
```diff
@@ -8198,6 +8198,5039 @@
         "xtend": "^4.0.0"
       }
     },
+    "postman-templating": {
```
lshaowei18, line 8201: As mentioned, I find it weird to have so many of the module's dependency installed in the package-lock. I don't quite understand how package-lock works and why it is like this, will probably spend more time reading up.

backend/package.json
```diff
@@ -11,7 +11,7 @@
     "postbuild": "npm run copy-assets",
     "start": "node build/server",
     "copy-assets": "copyfiles -u 1 src/assets/* src/**/*.sql build",
-    "test": "jest",
+    "test": "jest --passWithNoTests",
```
lshaowei18, line 14: Added the flag `--passWithNoTests` because I moved the tests to the local module. We can remove this once we have new tests on `backend`.

backend/src/email/services/email-template.service.ts
```diff
@@ -234,16 +235,41 @@ const addToMessageLogs = async (
...
   records: MessageBulkInsertInterface[]
 ): boolean => {
   return records.some((record) => !validator.isEmail(record.recipient))
 }
 
+/**
+ * Attempts to hydrate the first record.
+ * @param records
+ * @param templateBody
+ * @param templateSubject - optional
```
lshaowei18, line 253: I am sure there are ways to reflect optional params but idk which style guide are we following now

backend/src/email/services/email-template.service.ts
```diff
@@ -234,16 +235,41 @@ const addToMessageLogs = async (
...
+ */
+const testHydration = (
```
lshaowei18, line 255: I could abstract `testHydration` into `TemplateService` in `core`, but I find that it is cleaner this way, since SMS doesn't have `templateSubject`.

modules/postman-templating/tsconfig.json
```diff
@@ -0,0 +1,47 @@
...
+  "compileOnSave": true,
+  "compilerOptions": {
+    /* Basic Options */
+    "target": "es6",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
+    "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
+    "sourceMap": true,                        /* Generates corresponding '.map' file. */
+    "outDir": "./build",                      /* Redirect output structure to the directory. */
+    "rootDir": "./src",                       /* Specify the root directory of input files. Use to control the output directory structure with --outDir. */
+    "downlevelIteration": true,               /* Provide full support for iterables in 'for-of', spread, and destructuring when targeting 'ES5' or 'ES3'. */
+    "declaration": true,
```
lshaowei18, line 20: This flag allows TS to automatically create type declaration files.

lamkeewei:
Just noting down what we discussed with the team. 

These were the options that were proposed:
1. npm install from separate repo (postmangovsg-js-tools?)
2. npm pack and copy tarball into respective folders 
3. Publish to npm registry
4. Build and push Docker image in Travis and use Dockerrun.aws.json on beanstalk 
5. Separate branch (sdk branch)
6. [lerna](https://github.com/lerna/lerna) to manage multi-package repo

Of which, we will doing **option 4**. Meanwhile, I'll also take a look at the issues with the large package-lock.json diff.

modules/postman-templating/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "postman-templating",
+  "version": "1.0.0",
+  "description": "Templating for Postman",
+  "main": "build/index.js",
+  "scripts": {
+    "start": "tsc && node build/index.js",
+    "build": "rimraf build && tsc",
+    "prepare": "npm run build",
```
lamkeewei, line 9:
```suggestion
    "preinstall": "npm run build",
```

`prepare` doesn't run when _another_ module installs `postman-templating` as a dependency. `preinstall` will ensure that another package installing this will run the build step. Found the info at [npm docs](https://docs.npmjs.com/misc/scripts). Might help us avoid the manual `npm install`.

backend/package-lock.json
lamkeewei: I've tried running `npm dedupe` and it managed to reduce the number of lines changed to slightly less than 2000. From my understanding, what it does is that it tries to surface the common dependencies higher in dependency tree. Can take a look at [npm-dedupe](https://docs.npmjs.com/cli/dedupe).

lamkeewei: As a side note, most of the babel-related dependencies introduced seems to be coming from `jest`. There's an issue on their repo about ongoing work to reduce the number/size of their dependencies.

.travis.yml
```diff
@@ -20,9 +20,36 @@ jobs:
...
+        - echo $PWD
         - npm run lint-no-fix
+        - npm run build
         - npm run test
+      before_deploy:
+        - TAG=$REPO:travis-$TRAVIS_BRANCH-$TRAVIS_COMMIT-$TRAVIS_BUILD_NUMBER
```
lamkeewei, line 34: For consistency with workers, I think it would be better to put this into a separate shell script (`deploy.sh`).

lshaowei18: Hmmmm so i tried using shell script for before_deploy and it doesn't seem to work.

lshaowei18: We could just put everything in a shell script and run deploy, which I have not tried yet

backend/Dockerfile
```diff
@@ -4,12 +4,18 @@ FROM node:12-alpine
...
-COPY ./package* ./
-RUN npm ci
+WORKDIR /usr/home/postmangovsg/backend
+
+COPY modules /usr/home/postmangovsg/modules
+
+RUN cd /usr/home/postmangovsg/modules/postman-templating && npm install && npm run build
+
+COPY backend/package* ./
+RUN npm install
```
lshaowei18, line 14: Had to npm install here because of this bug: https://npm.community/t/npm-ci-fails-with-transitive-local-packages/4059

lshaowei18:
Had to use `npm install` instead of `npm ci` in Dockerfile because of : https://npm.community/t/npm-ci-fails-with-transitive-local-packages/4059
https://stackoverflow.com/questions/57825322/npm-ci-fails-on-circleci-with-no-such-file-or-directory-for-symlinked-modules

worker/Dockerfile
```diff
@@ -4,12 +4,18 @@ FROM node:12-alpine
 RUN apk update && apk upgrade && apk add --no-cache --virtual builds-deps build-base \
   g++ make python tini
 
-WORKDIR /usr/home/postmangovsg/
-COPY ./package* ./
+WORKDIR /usr/home/postmangovsg/worker
+
+COPY modules /usr/home/postmangovsg/modules
+
+RUN cd /usr/home/postmangovsg/modules/postman-templating && npm install && npm run build
```
jeantanzj, line 11: @lamkeewei why do we still need to install and build here, if the build runs on postinstall?

lshaowei18: Maybe we don't have to, I added this previously. Let me check if it is redundant.

lshaowei18: Hmmm so it seems like the original
`RUN cd /usr/home/postmangovsg/modules/postman-templating && npm install && npm run build`
Can be changed to 
`RUN cd /usr/home/postmangovsg/modules/postman-templating && npm install`

When we build on travis it seems like the `node_modules` in `modules/postman-templating` isn't copied into docker. 

As a result, when I try to run `npm ci` on `backend` in `Dockerfile`, it throws an error saying that `rimraf: not found` in `postman-templating`. 

Running `npm install` instead doesn't throw the error. 

I found this on stackoverflow: 
`npm ci (named after Continuous Integration) installs dependencies directly from package-lock.json and uses package.json only to validate that there are no mismatched versions. If any dependencies are missing or have incompatible versions, it will throw an error.`

https://travis-ci.com/github/opengovsg/postmangovsg/jobs/350810755

Do let me know if i got anything wrong; still not 100% sure what I am talking about

lamkeewei: I tried this out and is facing the same issue too. `npm install` works fine while `npm ci` gives the same error about `rimraf`. Probably got to find out more about the difference between `npm ci` and `npm install`.

On a separate note, I was able to drop the entire line: `RUN cd /usr/home/postmangovsg/modules/postman-templating && npm install && npm run build` and just do the`npm install` in `/usr/home/postmangovsg/workers`. 

My worker Dockerfile looks like this: 

```dockerfile
FROM node:12-alpine

# required for node-gyp
RUN apk update && apk upgrade && apk add --no-cache --virtual builds-deps build-base \
  g++ make python tini

WORKDIR /usr/home/postmangovsg/worker

COPY modules /usr/home/postmangovsg/modules
COPY worker/package* ./
RUN npm install

COPY worker/src ./src
COPY worker/tsconfig.json ./
RUN npm run build
RUN npm prune --production

EXPOSE 4000
ENTRYPOINT [ "tini", "--" ]
CMD ["npm", "start"]
```

You can test out the build locally by running `docker build -t worker -f worker/Dockerfile .` from the project root. If you have errors about missing modules, you can try removing `postman-templating/package-lock.json` and regenerating it with `npm install`.

lshaowei18: Hmmm so I was talking to chinying and he says that `npm ci` is preferred because `npm install` will install the `node_modules` according to the `package.json` and the version of the packages might change unexpectedly.

`npm ci` installs according to the `package-lock.json` so we know that whatever that is installed locally is what is installed on production.

lshaowei18:
Replaced with #436

PR #399, opened 2020-06-12
feat: implement encryption decryption service using SubtleCrypto

### LINKED ISSUE #362: [Registered] Encryption/Decryption service
Service that provides the following helper functions
- symmetric encryption (AES-GCM) - for transforming csv
- symmetric decryption - for decryption of payload on public facing page

miazima:
## Problem

Implement helper functions that are needed to encrypt and decrypt data.

Closes #362 

## Solution

Use in built browser api, `SubtleCrypto`, to implement encryption and decryption using AES-GCM mode.

*Todo*
- Working on supporting IE11 as IE11 does not support promises and uses a event driven approach


miazima:
Also requires.a polyfill for `TextEncoder` for IE11. (see https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder), not added currently

frontend/src/services/crypto.service.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export async function encryptData(payload: string, password: string) {
+  try {
+    const initializationVector = window.crypto.getRandomValues(
+      new Uint8Array(12)
+    )
+    const algorithm = {
+      name: ENCRYPTION_METHOD,
+      iv: initializationVector,
+    }
```
pregnantboy, line 43:
```suggestion
      tagLength: 128,
    }
```

frontend/src/services/crypto.service.ts
```diff
@@ -0,0 +1,94 @@
...
+  )
+
+  return { cipher: decodedCipher, iv: decodedIv }
+}
+
+export async function decryptData(ciphertext: string, password: string) {
+  try {
+    const { cipher, iv } = decodeCipherText(ciphertext)
+
+    const algorithm = { name: ENCRYPTION_METHOD, iv }
```
pregnantboy, line 79: add in `tagLength: 128` for ie 11 compat

frontend/src/services/crypto.service.ts
```diff
@@ -0,0 +1,94 @@
...
+    'raw',
+    pwHash,
+    ENCRYPTION_METHOD,
+    false, // extractable option
+    ['encrypt', 'decrypt'] // possible actions that can be done with the key
+  )
+  return key
+}
+
+function encodeCipherToString(buffer: ArrayBuffer, iv: Uint8Array) {
```
pregnantboy, line 18: why is cipher arraybuffer converted to base64 but iv to hex?
how about using a library for encoding/decoding? https://www.npmjs.com/package/byte-base64

PR #406, opened 2020-06-13
Display invalid stats

pregnantboy:
## Features:
- Get unsent, sent, error and invalid count from ops table based on status column
- Display invalid recipient count on frontend

Screenshot 2020-06-13 at 11 39 26 PM

### DIFF
--- a/backend/src/core/services/stats.service.ts
+++ b/backend/src/core/services/stats.service.ts
@@ -1,6 +1,6 @@
-import { fn, col, cast } from 'sequelize'
+import { fn, cast } from 'sequelize'
 import { Statistic, JobQueue } from '@core/models'
-import { CampaignStats } from '@core/interfaces'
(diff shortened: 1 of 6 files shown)


backend/src/core/services/stats.service.ts
```diff
@@ -74,10 +79,12 @@ const getCurrentStats = async (
 
     // Sent count must be added to archived sent count since sent messages are
     // not enqueued to ops table if a project is stopped and resumed
+    // Since sent and invalid are not retried, they have to be added to archived stats
     return {
       error: opsStats.error,
       unsent: opsStats.unsent,
       sent: opsStats.sent + archivedStats.sent,
+      invalid: opsStats.invalid + archivedStats.invalid,
```
jeantanzj, line 87: comment that this is to account for synchronous bounces / telegram

PR #411, opened 2020-06-14
chore: Shard out template module

lshaowei18:
## Problem

I felt that #391 is just massive and I am very uncomfortable with that, so this is part of my effort in trying to break it down into smaller PRs.

This PR is to be merged into `refactor-template-module` which is a newly created feature branch. 

## Solution

This new node module will consist of: 

`template-client`: Every function that is in the class, except for `testHydration` and its helper function. `testHydration` is not included because it is only for backend and I think that it doesn't fall under the scope of this class.

`parse-csv`: A function that we will need to use in both backend and frontend.

`CSVParams`: A type that was in S3 Client, helpful to make `parse-csv` clearer.

There are also various errors that related to templating, which can be found in `errors.ts`.

### Usage
We can import the function/class/types from the module by `import { TemplateClient } from 'postman-templating'`.

### Type Declaration
With `"declaration": true` in `tsconfig.json`, type declaration files are automatically generated in the build folder.


modules/postman-templating/src/parse-csv.ts
```diff
@@ -0,0 +1,55 @@
...
+import { 
+  RecipientColumnMissing,
+  UnexpectedDoubleQuoteError
+} from './errors'
+/**
+ * Ensures that the recipient column exists and converts headers to lowercase
+ * Deduplicates the csv by overriding the same recipient with newer records
+ * @param readStream
+ */
+export const parseCsv = async (readStream: NodeJS.ReadableStream): Promise<Array<CSVParams>> => {
```
lamkeewei, line 13: I'm not super familiar with Typescript but will using NodeJS types in this function have any effect in terms of being able to use this function in the browser?

lamkeewei:
I see that `papaparse` has been added in the frontend for checking for a valid CSV file, not sure if we want to consolidate on using that on the backend also.

modules/postman-templating/src/parse-csv.ts
lshaowei18: That's a good point. I don't think it can be usable in the browser. At the same time, the code for the frontend has not been written yet and i'm not sure what is the requirement for that.

lshaowei18: I was thinking we can change it to a different library if it doesn't work on frontend or just use `parseCsv` for backend & worker. Maybe a little short sighted, but I think we can figure out what to do at a later stage when someone is working on it.

lshaowei18:
> I see that `papaparse` has been added in the frontend for checking for a valid CSV file, not sure if we want to consolidate on using that on the backend also.

Ah you mean to use papaparse for the `parseCsv` function?

lamkeewei:
> Ah you mean to use papaparse for the `parseCsv` function? I would prefer to push this in first before we refactor and use `papaparse` in `parseCsv` if that's the case. Less room for errors and easier to review.

You have a good point there. We can relook at this when implementing the frontend templating. Other than that lgtm.

lshaowei18:
> > Ah you mean to use papaparse for the `parseCsv` function? I would prefer to push this in first before we refactor and use `papaparse` in `parseCsv` if that's the case. Less room for errors and easier to review.
> 
> You have a good point there. We can relook at this when implementing the frontend templating. Other than that lgtm.

Thanks for the review, will merge it into the feat branch now.

PR #415, opened 2020-06-15
[Template Module] refactor:  test hydration

lshaowei18:
## Problem

`testHydration` was doing too many things. Functions should ideally have a single responsibility.

## Solution

- Parse the csv file

### DIFF
--- a/backend/src/core/services/template.service.ts
+++ b/backend/src/core/services/template.service.ts
@@ -1,15 +1,21 @@
 import { v4 as uuid } from 'uuid'
+import { difference, keys } from 'lodash'
+
 import S3 from 'aws-sdk/clients/s3'
 import { Transaction } from 'sequelize/types'
 
 import config from '@core/config'
 import logger from '@core/logger'
+import { isSuperSet } from '@core/utils'
+import { MissingTemplateKeysError } from '@core/errors/template.errors'
 import { configureEndpoint } from '@core/utils/aws-endpoint'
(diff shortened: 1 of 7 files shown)


backend/src/core/services/template.service.ts
```diff
@@ -1,15 +1,21 @@
...
 import logger from '@core/logger'
+import { isSuperSet } from '@core/utils'
+import { MissingTemplateKeysError } from '@core/errors/template.errors'
 import { configureEndpoint } from '@core/utils/aws-endpoint'
 import { jwtUtils } from '@core/utils/jwt'
 import { Campaign } from '@core/models'
 import { CsvStatusInterface } from '@core/interfaces'
 
 const MAX_PROCESSING_TIME = config.get('csvProcessingTimeout')
+type CSVParams = { [key: string]: string }
```
lamkeewei, line 17: Can this type be imported from the `postman-templating` module instead of redeclaring again here?

lshaowei18: Yup it can. I will make another PR after this one is in that replaces the `templateClient`, `CSVParams`, the errors types, as well as uninstalling the libraries that are no longer in use for worker and backend.

lamkeewei:
lgtm. Good approach on the decoupling of the S3, CSV parsing and templating logic. Will need to clean up unused libraries and methods (mostly CSV parsing) in a later PR.

PR #417, opened 2020-06-16
chore: Improve Telegram schema

dvrylc:
## Problem

The current Telegram schema has a couple of missing constraints and inconsistent types. 

## Solution

- `telegram_messages` and `telegram_ops`
  - Make `campaign_id` and `recipient` not null
  - Add `status` column
- `telegram_ops`
  - Change `recipient` datatype to `number / BIGINT` – this matches the datatype of `telegram_id` in `bot_subscribers` and `telegram_subscribers`
- `telegram_subscribers`
  - Set `phone_number` to primary key
  - Set unique constraint on `telegram_id` – there should never be more than one mapping for any phone number or Telegram ID

### DIFF
--- a/backend/src/telegram/models/telegram-subscriber.ts
+++ b/backend/src/telegram/models/telegram-subscriber.ts
@@ -8,9 +8,15 @@ import { Column, Model, Table, DataType } from 'sequelize-typescript'
 export class TelegramSubscriber extends Model<TelegramSubscriber> {
   @Column({
     type: DataType.STRING,
+    primaryKey: true,
+    unique: true,
   })
   phoneNumber!: string
 
-  @Column({ type: DataType.BIGINT })
+  @Column({
+    type: DataType.BIGINT,
+    primaryKey: true,
+    unique: true,
+  })
   telegramId!: number
 }
--- a/backend/src/telegram/models/telegram-subscriber.ts
+++ b/backend/src/telegram/models/telegram-subscriber.ts
@@ -1,4 +1,5 @@
-import { Column, Model, Table, DataType } from 'sequelize-typescript'
+import { Column, Model, Table, DataType, HasMany } from 'sequelize-typescript'
+import { BotSubscriber } from '.'
 
 @Table({
   tableName: 'telegram_subscribers',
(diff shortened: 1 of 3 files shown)


backend/src/telegram/models/telegram-subscriber.ts
```diff
@@ -8,9 +8,15 @@ import { Column, Model, Table, DataType } from 'sequelize-typescript'
 export class TelegramSubscriber extends Model<TelegramSubscriber> {
```
lamkeewei, line 8:
```suggestion
export class TelegramSubscriber extends Model<TelegramSubscriber> {
  @HasMany(() => BotSubscriber, { as: 'bot_subscriber' })
```

I added this association also while implementing the backend. Sorry, forgot to create a PR at that point.

backend/src/telegram/models/telegram-subscriber.ts
```diff
@@ -8,9 +8,15 @@ import { Column, Model, Table, DataType } from 'sequelize-typescript'
...
   @Column({
     type: DataType.STRING,
+    primaryKey: true,
+    unique: true,
```
lamkeewei, line 12: Not sure if we still need to set `unique` here given that primary keys are already unique.

backend/src/telegram/models/telegram-subscriber.ts
```diff
@@ -8,9 +8,15 @@ import { Column, Model, Table, DataType } from 'sequelize-typescript'
...
   })
   phoneNumber!: string
 
-  @Column({ type: DataType.BIGINT })
+  @Column({
```
lamkeewei, line 16: I'm thinking that the constraint on `phoneNumber` might be enough to ensure uniqueness for the `phoneNumber` - `telegramId` mapping.

backend/src/telegram/models/telegram-subscriber.ts
dvrylc: I added `unique` here because the primary key is compound. Considering this situation:

```
phone_1: telegram_1
phone_1: telegram_2
```

The primary key unique constraint would still allow this, but to us it's invalid. 

But now I'm re-thinking if we even need a compound primary key here. I think a possible alternative is to set one of the columns as the primary key and the other as unique - this indexes both columns anyway, so the compound primary key just seems like an extra index for no benefit?

lamkeewei:
lgtm

PR #420, opened 2020-06-16
[Template Module] chore: Deployment changes to backend & worker

lshaowei18:
## Problem

Deployment to backend/worker/frontend will have to change to accommodate the local `postman-templating` module.

### backend:
- Build docker image on travis
- Push build to ECR (Elastic Container Registry)
- Deploy zip file with `Dockerrun.aws.json` in it to EB (Elastic Beanstalk)
- EB will pull image from ECR

### worker:
- Changes to the deploy script and builds on the root folder instead of the worker folder.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,14 +71,15 @@ jobs:
         - npm install
       script:
         - npm run lint-no-fix
+        - cd ../
       deploy:
         - provider: script
-          script: ./deploy.sh postmangovsg-workers staging-sending staging-logger
+          script: ./worker/deploy.sh postmangovsg-workers staging-sending staging-logger
           on:
             branch: $STAGING_BRANCH
             condition: "$DEPLOY_WORKER = true"
         - provider: script
-          script: ./deploy.sh postmangovsg-workers prod-sending prod-logger
+          script: ./worker/deploy.sh postmangovsg-workers prod-sending prod-logger
           on:
             branch: $PRODUCTION_BRANCH
             condition: "$DEPLOY_WORKER = true"
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,25 @@ jobs:
       script:
         - npm run lint-no-fix
         - npm run test
+      before_deploy:
+        - cd ../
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -23,6 +23,25 @@ jobs:
...
+        - export TAG=travis-$TRAVIS_COMMIT-$TRAVIS_BUILD_NUMBER
+        - export ECR_REPO=$REPO-backend
+        - export PATH=$PATH:$HOME/.local/bin
+          # Install tools for deployment
+        - pip install --user awscli #For docker
+
+          # Login to AWS ECR, credentials defined in $AWS_ACCESS_KEY_ID and $AWS_SECRET_ACCESS_KEY
+        - $(aws ecr get-login --no-include-email --region $AWS_DEFAULT_REGION)
+        - docker build -f ./backend/Dockerfile -t $ECR_REPO:$TAG .
+        - docker tag $ECR_REPO:$TAG $ECR_REPO:$TRAVIS_BRANCH
```
lamkeewei, line 37: Not too sure if this additional tag is required. In our `Dockerrun.aws.json` we are running `$ECR_REPO:$TAG`, so this tagged image is never used anywhere else.

lshaowei18: Ah that's true. Tbh i don't really understand docker tags; I just copied Go's code. Will take another look at it.

lshaowei18: I think `$ECR_REPO:$TRAVIS_BRANCH` will be useful when there are multiple branches that are pushing to ECR; it makes it slightly easier to differentiate which branch pushed the image. 

Since tags are just a way for us to refer to images, seems like there is no harm in adding an extra tag.

lamkeewei: Yep that makes sense. I think it's useful as a quick pointer to the latest build on a particular branch. Also given that additional tags do not copy anything, there's no downside to this.

lamkeewei:
lgtm. Note on having to uncomment out the lines to build the templating module in future PR.

Issue #407, opened 2020-06-13
Buttons not disabled on click


pregnantboy:
**Describe the bug**
Some buttons are not disabled on click, allowing the same api call to be made multiple times.
 e.g. Pause campaign, resume campaign,  save template, etc..

PR #422, opened 2020-06-16
fix: Prevent duplicate campaign creation

dvrylc:
## Problem

Clicking the submit button twice (i.e. double clicking) during campaign creates duplicate campaigns.

Part of #407, not sure if we wanna tackle all in this PR?

## Solution

Disable button and event handler when a campaign creation is already in progress.

### DIFF
--- a/frontend/src/components/dashboard/create/create-modal/CreateModal.tsx
+++ b/frontend/src/components/dashboard/create/create-modal/CreateModal.tsx
@@ -15,8 +15,13 @@ const CreateModal = () => {
   const history = useHistory()
   const [selectedChannel, setSelectedChannel] = useState(ChannelType.SMS)
   const [name, setName] = useState('')
+  const [isCreating, setIsCreating] = useState(false)
+
   async function handleCreateCampaign() {
     try {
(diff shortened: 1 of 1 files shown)


frontend/src/components/dashboard/create/create-modal/CreateModal.tsx
```diff
@@ -15,8 +15,13 @@ const CreateModal = () => {
   const history = useHistory()
   const [selectedChannel, setSelectedChannel] = useState(ChannelType.SMS)
   const [name, setName] = useState('')
+  const [isCreating, setIsCreating] = useState(false)
+
   async function handleCreateCampaign() {
     try {
+      if (isCreating) return
+      setIsCreating(true)
```
miazima, line 23: Should we set this back to false in the `catch`?  In the case that `createCampaign` throws an error, user would not be able to try again.

dvrylc: Oh yes missed that, thanks!

dvrylc:
Closing in favor of #430.

PR #430, opened 2020-06-19
feat: Automatic input and button disabling when executing async functions

### LINKED ISSUE #407: Buttons not disabled on click

dvrylc:
## Problem

Some of our inputs and buttons are not disabled when an async function call is fired, allowing unwanted duplicate calls to be made (e.g. by double-clicking a button). 

Closes #407.

## Solution

This PR introduces a local state variable `asyncLoading` in `PrimaryButton` and `TextInputWithButton` that signifies if an async function execution is ongoing. The input and button are disabled based on `asyncLoading` to prevent unwanted user interaction while waiting for a response. 

The default behavior is to re-enable the input/button after the async function execution completes, but there are quite a few cases where the component is unmounted right after the response is received – for example when navigating away after clicking a button. Trying to re-enable an unmounted component would throw an exception. I've introduced a ref `isMounted` that is set to `false` during unmount, ensuring that we try to re-enable the component only if it's still mounted. 

This "disable-while-awaiting" behavior was previously implemented in several parent components throughout the app and are now not necessary – these have been removed. 

## Tests

1. Turn on network throttling in devtools and set it to `Slow 3G`
2. Go through the usual flows, ensuring that any async operation triggered by `PrimaryButton` or `TextInputWithButton` cannot be fired more than once

### DIFF
--- a/frontend/src/components/common/text-input-with-button/TextInputWithButton.tsx
+++ b/frontend/src/components/common/text-input-with-button/TextInputWithButton.tsx
@@ -1,56 +1,69 @@
-import React from 'react'
+import React, { useState, useRef, useEffect, MutableRefObject } from 'react'
 import cx from 'classnames'
 import styles from './TextInputWithButton.module.scss'
 import { PrimaryButton, TextInput } from '../'
 
-const TextInputWithButton = (props: {
-  value: string
-  onChange: Function
-  onClick: Function
-  children: React.ReactNode
-  type?: string
+interface TextInputWithButtonProps
+  extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'onChange'> {
+  onChange: (value: string) => void
+  onClick: () => void | Promise<void>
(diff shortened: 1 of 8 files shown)


frontend/src/components/common/text-input-with-button/TextInputWithButton.tsx
```diff
@@ -1,56 +1,69 @@
...
+
+  const asyncSubmit = async (e: React.FormEvent) => {
     e.preventDefault()
+
+    setAsyncLoading(true)
+    await Promise.resolve(onClick())
+    // Only enable if still mounted
+    if (isMounted.current) {
+      setAsyncLoading(false)
+    }
```
pregnantboy, line 47: same as above

PR #431, opened 2020-06-22
[Template Module] chore: Use postman-templating module in worker and backend

lshaowei18:
## Solution

Installed `postman-templating` on both worker and backend.

Replaced `TemplateClient`, `TemplateError` and remove the relevant code for worker and backend with `postman-templating`.

Uninstalled `squirrelly` for both worker and backend.

Removed `parseCsv`, related errors, tests, dependencies from `postman-templating` as we decided that it is out of the scope for this feat, could always refactor later on.

Introduced a new type called `CSVParams`, using it to replace CSVParamsInterface.

## TODO
- [x] Move `parseCsv` out of `s3.client` and into core services
- [x] Uninstall `xss` on backend


lshaowei18:
Marking this as draft, will rebase with feature branch and configure deployments.

backend/src/core/services/s3-client.class.ts
```diff
@@ -37,9 +38,7 @@ export default class S3Client {
    * Deduplicates the csv by overriding the same recipient with newer records
    * @param readStream
    */
-  async parseCsv(
-    readStream: NodeJS.ReadableStream
-  ): Promise<Array<CSVParamsInterface>> {
+  async parseCsv(readStream: NodeJS.ReadableStream): Promise<Array<CSVParams>> {
```
lamkeewei, line 41: Would it be easier if we move `parseCsv` out of `S3Client` at this point so that it will be easier when we do get around to refactoring it?

lshaowei18: That's a good idea. Will put it inside core services.

backend/src/core/services/parse-csv.service.ts
```diff
@@ -0,0 +1,65 @@
...
+  } catch (err) {
+    if (err.message.includes('Invalid Opening Quote'))
+      throw new UnexpectedDoubleQuoteError()
+    if (err.message.includes('Invalid Closing Quote'))
+      throw new UnexpectedDoubleQuoteError()
+    throw err
+  }
+}
+
+export const parseCsvService = {
```
lamkeewei, line 63:
```suggestion
export const ParseCsvService = {
```
Should probably capitalize to be consistent with the other services (e.g. `ApiKeyService`, `AuthService`)

backend/src/core/types/s3.types.ts
```diff
@@ -0,0 +1 @@
+export type CSVParams = { [key: string]: string }
```
lamkeewei, line 1: Minor nitpick, should probably rename file to `parse-csv.types.ts`

lshaowei18: Thanks that's a good catch I forgot about this file.

lamkeewei:
lgtm. 👍

PR #436, opened 2020-06-24
Refactor template module

### LINKED ISSUE #363: [Registered] Shard out templating logic

lshaowei18:
## Problem

Closes #363.

## Solution

**TemplateClient**
Moved `TemplateClient` into a local node module called `postman-templating`. 

`TemplateClient` no longer has an exported function called `testHydration`, as I felt that `testHydration` was doing too many things and doesn't fall into `TemplateClient's` scope. 

- Parse the csv file

**Improvements**:

- Moved `TemplateClient` and relevant errors into a local node module called `postman-templating`
- Backend & worker now uses `postman-templating` module.
- Some of the errors related to templating is now imported from `postman-templating`

**TODOs**:
- [x] clean up optional for `testHydration`
- [ ] Look into node module cache
- [ ] Check if `CSVParams` is needed
- [x] Change `@types` to dev dependencies


lshaowei18:
Closing this for now, since this is quite off from what we want. Will look into multipart upload + papaparse chunking

modules/postman-templating/src/template-client.ts
```diff
@@ -37,13 +27,13 @@ export default class TemplateClient {
    * @param templateBody - template body
    * @param params - dict of param variables used for interpolation
    */
-  parseTemplate(
+  parseTemplate = (
```
pregnantboy, line 30: why change to arrow function?

pregnantboy:
@lshaowei18 @lamkeewei 
Changes:
- Dev dependencies in postman-templating that are required for build is moved to dependencies (will explain later)
- Travis copies modules folder into current directory (backend, work, etc..) before deploy
- The whole folder gets zipped up and uploaded to the bucket as per normal
- On docker build, the modules folder is copied to the parent folder, so the dependency file path in package.json is still correct
- docker run npm ci
- This will also install the shared modules and build it since it is under the required array in package.json. The reason why it couldnt be built before was because typescript and rimraf was in devDependencies and not installed.

modules/postman-templating/src/template-client.ts
lshaowei18: Sorry probably a mistake. Will change it back.

pregnantboy:
@lamkeewei @lshaowei18 
- using npm ci instead of npm install for backend. this will cache .npm folder instead of node_modules. Caching node_modules will break the build as it will not run postinstall on the shared modules. 
ref: https://stackoverflow.com/questions/42521884/should-i-have-travis-cache-node-modules-or-home-npm
ref: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#caching-with-npm
- set `edge: true` so that modules folder will be zipped. or else it will only zip files in `git ls-files`. 
ref: https://docs.travis-ci.com/user/deployment-v2/providers/elasticbeanstalk/#controlling-which-files-are-included-in-the-zip-archive

Issue #364, opened 2020-06-03
[Registered] Csv transform service


pregnantboy:
- parses csv
- imports templating module and hydrates rows
- encrypts hydrated messages with password
- generates password hash
- output csv should only contain `recipient`, `encrypted_payload` and `password_hash` columns
- upload stream to s3

**explore using workers, `papaparse` supports this out of the box**

jeantanzj:
Can we look into sharing CSV parsing code with backend? Backend currently using csv-parse.

jeantanzj:
Will look into parsing row by row and streaming the upload. Will stub the encryption service in this PR if not available yet

lshaowei18:
@lamkeewei mentioned that frontend is using `papaparse` to parse the csv. Will start looking into it while the templating module PR is being reviewed.

lshaowei18:
> @lamkeewei mentioned that frontend is using `papaparse` to parse the csv. Will start looking into it while the templating module PR is being reviewed.

During a sync up on 22 June we have decided that we will not include the parsing of csv into templating module PR and would refactor later down the line after the PR is merged. Related to #432.

lshaowei18:
Password management is quite scary to me so I have been reading a little about how to store passwords. It seems to me like a sufficient approach would be to generate a random salt for each password and hashing the password with the random salt. When a user logins, we can take the password and hash it with the random salt that is generated and stored as part of the password hash and verify if the provided password is valid.

Google recommends a multi layered strategy: 
`Have the client computer hash the password using a cryptographically secure algorithm and a
unique salt provided by the server. When the password is received by the server, hash it again with a different salt that is unknown to the client. Be sure to store both salts securely.`

Does seem like they just use two salts instead of a random salt for each user.

I am still learning and trying to understand more about this, so if I am wrong please correct me!

lshaowei18:
Started to look into the parsing of csv and here are my assumptions: 

- Reason why we want to stream the file is because we are afraid that the file is too big for us to store it in the browser
- We will parse the whole CSV file, and then loop through each row to template it.
- Loops through the templated messages and encrypt it

If that's the case, I think we can share the `parseCsv` between backend and frontend. 

Initially, I thought we will template it while parsing.

lshaowei18:
So I have prototyped and here's a general view of what I envision this service to do:

- User creates a password protected email campaign
- User fills in the email subject and body, clicks on upload file
[...]

PR #459, opened 2020-07-01
Upload component for Protected Email 

### LINKED ISSUE #365: [Registered] New UploadRecipients Component for registered mail
Additional features for component
- New textarea for `content template`
- Preview first templated unencrypted content
- Uploads file and call new endpoint `email/upload/complete/registered` endpoint

Pending UI design

miazima:
## Problem

New upload component for protected email to support recipient template creation, previewing recipient template and encrypting message

Closes #365 

**Features**:

- Include `protectedlink` as a required column
- Extract params from message B template before triggering sample csv download
- new upload component `ProtectedEmailRecipients` 
- Interface for user to enter recipient template, upload file, preview template and then encrypt message
- once user gets feedback on successful encryption, user can click next which executes upload to s3 (stubbed)

## Before & After Screenshots

image
image
image


frontend/src/services/protectedemail.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  preview?: string
+}
+
+export async function extractTemplateParams(
+  template: string
+): Promise<Array<string>> {
+  return Promise.resolve(['recipient', 'password', 'secretmsg'])
+}
+
+export async function validateCsv(
```
lshaowei18, line 14: Hmmmm I didn't know we were going to do this, but it does seem like a good idea if we don't mind parsing the file twice. 

What I mean is that in this function, we could parse the whole file using `papaparse` and make sure that `recipient`, `password` and any params used in the template is in each row of the csv and that there are no parsing related errors.

If there is any error in this stage, we will not proceed to encrypt the content, hash the password, or upload the file.

Once we are sure that the params have no errors, we could proceed with the uploading. I have written some thoughts on what the uploading would be like. #364

frontend/src/components/dashboard/create/email/ProtectedEmailRecipients.tsx
```diff
@@ -0,0 +1,295 @@
...
+  params,
+  onNext,
+}: {
+  csvFilename: string
+  numRecipients: number
+  params: Array<string>
+  onNext: (changes: Partial<EmailCampaign>, next?: boolean) => void
+}) => {
+  const containerRef = useRef<HTMLAnchorElement>(null)
+  const [body, setBody] = useState('')
```
lshaowei18, line 39: Is `body` the content template?

miazima: yes, shall rename it `content`

lshaowei18:
Hi @miazima , if you haven't started work on `extractTemplateParams` and `validateCsv` i will work on it with a separate PR would that be ok? Sorry it's kinda messy

miazima:
@lshaowei18 okay sure

frontend/src/services/protectedemail.service.ts
```diff
@@ -0,0 +1,30 @@
+export interface CsvStatus {
+  csvFilename?: string
+  csvError?: string
```
lshaowei18, line 3: Hmmm i'm not sure if the `csvError` should be a string or an array of strings because there could be multiple rows with missing params. @miazima what do you think? Should we display all the errors or allow them to download the errors? 

We could stick with a generic string first for now, and open an issue so that the scope of this work don't become too big.

frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -70,13 +72,19 @@ const EmailTemplate = ({
...
+        <b>Note:</b> Recipient is a required column in the CSV file.{' '}
+        {protect && (
+          <>
+            Include <b>{'{{ protectedlink }}'}</b> to insert the unique
+            protected link for the recipient.
+          </>
+        )}
       </p>
       <p>
         Example
```
pregnantboy, line 90: the example is irrelevant for protected mail..

frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -70,13 +72,19 @@ const EmailTemplate = ({
         value={subject}
         onChange={setSubject}
       />
-      <h4>Message</h4>
+      <h4>{protect ? 'Message A' : 'Message'}</h4>
```
pregnantboy, line 75: Message A may make no sense to the user. Will prefer something like Plaintext message, or something.. TBD

frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -70,13 +72,19 @@ const EmailTemplate = ({
...
       <p>
         To personalise your message, include keywords that are surrounded by
         double curly braces. The keywords in your message template should match
         the headers in your recipients CSV file.
         <br />
-        <b>Note:</b> Recipient is a required column in the CSV file.
+        <b>Note:</b> Recipient is a required column in the CSV file.{' '}
+        {protect && (
+          <>
+            Include <b>{'{{ protectedlink }}'}</b> to insert the unique
```
pregnantboy, line 84: wording has to imply that protectedlink is a compulsory field. handleSaveTemplate should check for this as well..(along with server-side checks)

frontend/src/components/common/sample-csv/SampleCsv.tsx
```diff
@@ -8,15 +8,22 @@ import styles from './SampleCsv.module.scss'
...
   params,
   defaultRecipient,
+  elementRef,
 }: {
   params: Array<string>
   defaultRecipient: string
+  elementRef?: RefObject<HTMLAnchorElement>
 }) => {
-  const RECIPIENT_HEADER = 'recipient'
+  const RECIPIENT_KEYWORDS = ['recipient', 'password']
```
pregnantboy, line 17: this will affect both protected and non-protected campaigns

frontend/src/components/dashboard/create/email/ProtectedEmailRecipients.tsx
```diff
@@ -0,0 +1,295 @@
...
+          value={content}
+          onChange={setContent}
+        />
+
+        <div className="separator"></div>
+      </>
+    )
+  }
+
+  async function onDownloadSample(
```
pregnantboy, line 84: a bit too hackish. would prefer passing the template into sampleCsv as a param, together with the protected flag

frontend/src/services/protectedemail.service.ts
```diff
@@ -0,0 +1,30 @@
+export interface CsvStatus {
```
pregnantboy, line 1: call it protected.service.ts, should be easily extendable to sms or other channels

frontend/src/components/dashboard/create/email/ProtectedEmailRecipients.tsx
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  async function onEncryptCsv() {
+    if (!tempFile) {
+      setErrorMessage('Error encrypting file')
+      return
+    }
+    setIsCsvEncrypting(true)
+    const encryptedFile = await encryptCsv(tempFile)
+    setEncryptedFile(encryptedFile)
```
pregnantboy, line 137: there is no encrypted file

PR #462, opened 2020-07-01
[Registered] feat: Multipart upload to s3 

### LINKED ISSUE #364: [Registered] Csv transform service

lshaowei18:
## Problem

For registered mail, we do not want to store the hydrated message as well as the password in plain text.

Closes #364

## Solution

When the user uploads a csv file, the following happens in the browser: 
- We get the template from backend
- Start multipart upload
- The csv file will be parsed in chunks, the default chunk size is 10mb.
- For each chunk:
  - Transform all of the params into this format `recipient,encryptedPayload,passwordHash`
  - Join all the transformed data into a single string
  - Get presigned url from backend
  - Upload the single string of data onto s3 with the presigned url
- Once the whole file is parsed, we complete the multipart upload

### Multipart upload to s3
The need to keep track of etags and part number makes multipart upload more complex than a normal upload.

Imagine we have 3 parts that we have uploaded. In order to complete the upload using aws sdk s3's `completeMultipartUpload`, we need to provide the following information about the parts: 

` MultipartUpload: {
   Parts: [
      {
     ETag: "\"d8c2eafd90c266e19ab9dcacc479f8af\"", 
     PartNumber: 1
    }, 
      {
     ETag: "\"d7z1fhgd13aw57s9hj64fhaccagbm51a\"", 
     PartNumber: 2
    }
   ] 
`

Part number: `A part number uniquely identifies a part and also defines its position within the object being created. If you upload a new part using the same part number that was used with a previous part, the previously uploaded part is overwritten.`

ETag: `An opaque identifier assigned by a web server to a specific version of a resource found at a URL.`

The etag provided for us whenever we make a successful upload to s3 with a presigned url and is unique for each individual part upload.

Part number is something that we have to keep track of and we have to ensure that they are in sequence and cannot be repeated. Initially i wanted to keep track of the part number on the backend, but it means that I have to use a db to keep track of the state. So I ended up keeping track of it in the frontend.

### Papaparse 
We are not making use of the `workers` and asynchronous parsing because I couldn't find a way to keep track and assign the correct part number to each chunk if we were to make use of these two features. 

I am currently keeping track of the part number using a variable. The problem is that when it is asynchronous, it creates a race condition and as a result the part number is not incremented properly.

If we find that the process is too slow and wants to make it faster, we could keep track of the part number in backend and we might be able to make use of `workers` and asynchronous parsing.

To enable synchronous parsing while using an async function on chunk, we need to make use of `parser.pause` and `parser.resume`.

### Transaction id
Initially i was passing `uploadId` and `s3Key` between frontend and backend, but after discussing with chinying, we think that it is safer to sign it with JWT. 

**Features**:

Backend:
- `/campaign/{campaignId}/email/upload-start-multipart`: Starts a multipart upload
- `/campaign/{campaignId}/email/upload-multipart-url`: Get a presigned url to upload a single part
- `/campaign/{campaignId}/email/upload-complete-multipart`:  Complete multipart

Frontend:
- Ability to parse files in synchronous chunks and multipart upload to s3.

TODOs:

- [ ] research about readable stream on frontend
- [x] Add JWT for uploadId and s3Key
- [ ] Install templating module and template the message


backend/src/email/middlewares/email.middleware.ts
```diff
@@ -92,9 +95,98 @@ const previewFirstMessage = async (
...
+    await s3Client.completeMultipartUpload({
+      transactionId,
+      partCount,
+      etags,
+    })
+
+    // Not sure what i should return here
+    return res.json({
+      transaction_id: transactionId,
+    })
```
miazima, line 178: I think we can return a 200 here, this function could be extracted out to be part of a larger function that handles the completion of upload as well as updating the campaign and db

backend/src/email/middlewares/email.middleware.ts
```diff
@@ -92,9 +95,98 @@ const previewFirstMessage = async (
...
+      transaction_id: transactionId,
+      part_count: partCount,
+      etags,
+    } = req.body
+
+    await s3Client.completeMultipartUpload({
+      transactionId,
+      partCount,
+      etags,
+    })
```
miazima, line 173: return the 's3Key` here to be able to read the file and extract the records

backend/src/email/routes/email-campaign.routes.ts
```diff
@@ -560,4 +582,149 @@ router.post(
...
+ *           description: Bad Request
+ *         "401":
+ *           description: Unauthorized
+ *         "500":
+ *           description: Internal Server Error
+ */
+router.post(
+  '/upload-complete-multipart',
+  celebrate(completeMultipartValidator),
+  EmailMiddleware.completeMultipart
```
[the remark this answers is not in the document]
lshaowei18: Hmmm that's true. Should we opt for multiple middlewares? Keeps our function smaller and hopefully more readable.

For example:
`
  router.post(
  '/upload-complete-multipart',
  celebrate(completeMultipartValidator),
  EmailMiddleware.completeMultipart,
  EmailTemplateMiddleware.updateEncryptedMessages
)`

backend/src/email/middlewares/email-template.middleware.ts
```diff
@@ -255,16 +262,103 @@ const deleteCsvErrorHandler = async (
...
+    // - download from s3
+    const s3Client = new S3Client()
+    const fileContent = await s3Client.getCsvFile(s3Key)
+
+    const records = UploadService.getProtectedRecordsFromCsv(
+      +campaignId,
+      fileContent
+    )
+
+    if (EmailTemplateService.hasInvalidEmailRecipient(records)) {
```
lshaowei18, line 309: It seems alittle too late to check the emails here, because the file would have been encrypted and uploaded by then. Could we move this check to the frontend?

PR #467, opened 2020-07-01
Fix normal links in Telegram templates

### LINKED ISSUE #465: [Telegram] Normal links in message template broken
**Describe the bug**
Previous fix for telegram inline mentions broke the parsing of normal links. 

**To Reproduce**
1. Create a Telegram campaign
2. Create a template with a normal link (`<a href='https://google.com'>Hello</a>`)
3. Send a test message
4. Error will occur 

**Expected behavior**
Normal links should be supported.

lamkeewei:
## Problem

Normal links were being stripped by the previous xss options for Telegram. 

Closes #465 

## Solution
**Bug Fixes**:
- Use `onTagAttr` instead of `safeAttrValue`. I wrongly assumed previously that `safeAttrValue` default behaviour will just use the default `safeAttrValue` function when nothing is returned. 
    - Reference: [docs](https://github.com/leizongmin/js-xss/tree/v1.0.6#customize-the-handler-function-for-attributes-of-matched-tags) for `onTagAttr`
- Add CSS to show styles for `pre` and `code` blocks. 
- Replace `\n` with `<br />` in recipients page preview to show line breaks in preview. 

## Tests
Create a Telegram campaign and send a test message with this template to try out all the supported tags:
```
<b>bold</b>, bold
<i>italic</i>, italic
<u>underline</u>, underline
<s>strikethrough</s>, <strike>strikethrough</strike>, <del>strikethrough</del>
<b>bold <i>italic bold <s>italic bold strikethrough</s> <u>underline italic bold</u></i> bold</b>
<a href="https://www.google.com/">inline URL</a>
<a href="tg://user?id=837238105">inline mention of a user</a>
<code>inline fixed-width code</code>
<pre>pre-formatted fixed-width code block</pre>
<pre><code class="language-python">pre-formatted fixed-width code block written in the Python programming language</code></pre>
```

### DIFF
--- a/backend/src/core/config.ts
+++ b/backend/src/core/config.ts
@@ -5,6 +5,7 @@
 import convict from 'convict'
 import fs from 'fs'
 import path from 'path'
+import xss from 'xss'
 const rdsCa = fs.readFileSync(path.join(__dirname, '../assets/db-ca.pem'))
 /**
  * To require an env var without setting a default,
@@ -362,21 +363,16 @@ const config = convict({
           pre: [],
           a: ['href'],
         },
-        onTagAttr: (
+        safeAttrValue: (
           tag: string,
           name: string,
-          value: string,
-          isWhiteAttr: boolean
+          value: string
         ): string | void => {
           // Handle Telegram mention as xss-js does not recognize it as a valid url.
-          if (
-            isWhiteAttr &&
-            tag === 'a' &&
-            name === 'href' &&
-            value.startsWith('tg://')
-          ) {
-            return `href="${value}"`
+          if (tag === 'a' && name === 'href' && value.startsWith('tg://')) {
+            return value
           }
+          return xss.safeAttrValue(tag, name, value, xss.cssFilter)
         },
         stripIgnoreTag: true,
       },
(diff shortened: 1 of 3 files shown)


backend/src/core/config.ts
```diff
@@ -362,14 +362,20 @@ const config = convict({
...
-          if (tag === 'a' && name === 'href' && value.startsWith('tg://')) {
-            return value
+          if (
+            isWhiteAttr &&
+            tag === 'a' &&
+            name === 'href' &&
+            value.startsWith('tg://')
+          ) {
+            return `href="${value}"`
           }
```
dvrylc, line 379:
```suggestion
          }
          return xss.safeAttrValue(tag, name, value, xss.cssFilter)
```

I actually agree that `safeAttrValue` is more suitable for this use case since we're modifying the value and not the attribute itself. 

The package [exports the default `safeAttrValue`](https://github.com/leizongmin/js-xss/blob/1e3bda7871635bf7c35b7ee4cb5c84842cc14e2c/lib/index.js#L26), so I thought we could stick with the original implementation and add a fallback on the default. But this effectively breaks our ability to pass in customized `cssfilter` options - we don't use it but not sure if we're ok with that?

lamkeewei: Yeah that definitely looks cleaner. I think it's unlikely we will need to use a customized `cssfilter` for Telegram. Will change to use `safeAttrValue` and merge in. Thanks!

PR #468, opened 2020-07-01
chore: log jobs using id

jeantanzj:


jeantanzj:
Adding a note here that this change was made to enable us to send messages with duplicate recipients

PR #470, opened 2020-07-01
chore: copy modules in during travis build, and copy out in dockerfile

pregnantboy:
Todo:
- [x] ~Set up eslint for shared modules~ will do it another pr
- [x] Move packages required to build to `dependencies `
- [x] ~Update dockerfiles to npm ci --production for modules~ no longer needed


modules/postman-templating/src/errors.ts
```diff
@@ -4,4 +4,4 @@ export class TemplateError extends Error {
     Object.setPrototypeOf(this, new.target.prototype) // restore prototype chain
     Error.captureStackTrace(this)
   }
-}
```
lshaowei18, line 7: This is a little weird; what changes are made to this file?

pregnantboy: just linted it, the ⛔️  means there no new line at the end of file

pregnantboy:
> Lgtm, thanks for cleaning up the travis and dockerrun, are we still doing tarball?

nope

pregnantboy:
@lshaowei18 @lamkeewei 
Changes:
- Dev dependencies in postman-templating that are required for build is moved to dependencies (will explain later)
- Travis copies modules folder into current directory (backend, work, etc..) before deploy
- The whole folder gets zipped up and uploaded to the bucket as per normal
- On docker build, the modules folder is copied to the parent folder, so the dependency file path in package.json is still correct
- docker run npm ci
- This will also install the shared modules and build it since it is under the required array in package.json. The reason why it couldnt be built before was because typescript and rimraf was in devDependencies and not installed.

PR #480, opened 2020-07-03
chore: remove slashes in amplify

jeantanzj:
## Problem

Amplify could not render our app when our amplify.yml file's new lines were appended with slashes.

## Solution

Do not lint the amplify.yml file

Maybe there's a better rule for this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "typescript": "^3.8.3"
   },
   "lint-staged": {
-    "*.{yml,md},!amplify.yml": "prettier --write"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -45,7 +45,7 @@
     "typescript": "^3.8.3"
   },
   "lint-staged": {
-    "*.{yml,md}": "prettier --write"
+    "*.{yml,md},!amplify.yml": "prettier --write"
```
dvrylc, line 48: Running prettier via lint-staged still respects the `.prettierignore` so we don't have to explicitly exclude it

PR #487, opened 2020-07-06
[Registered] chore: Validating CSV

### LINKED ISSUE #475: [Registered] Validate CSV file
We decided that it would be simpler we parse the whole csv file to look for parsing errors as well as missing params first, before hydrating and transforming it.

Refer to #459 & #364  for more context.

lshaowei18:
## Problem

Encrypting and hashing each row can take up quite alot of time for big files. 

If the csv file turns out to be malformed, the time taken will be wasted. Thus, we decided that we should validate the csv as much as possible before encrypting and hashing.

Closes #475

## Solution

There are two things that needs to be present for the csv:
- The keys that are used in the template: `Hi {{name}}, you scored {{score}} for your O levels.`, the keys `name` and `score` must be present.
- Keys that we need for sending, encryption, and hashing. Currently, they are `recipient` and `password`.

The csv file will parsed twice in `validateCsv`. 

The first time is to validate and ensure all the keys ar in the csv headers. This is a cheap and fast way to find out if the csv file is somewhat valid, since we are just checking the headers.

The second time we will parse the whole file without the headers. We will check that all the required keys are in each row.

### DIFF
--- a/frontend/src/services/validate.service.ts
+++ b/frontend/src/services/validate.service.ts
@@ -20,7 +20,14 @@ export async function validateCsv(
 ): Promise<CsvStatus> {
   try {
     await validateHeaders(file, templateParams)
-    await validateRows(file, templateParams)
+    const numRows = await validateRows(file, templateParams)
+    return Promise.resolve({
+      csvError: '',
+      numRecipients: numRows,
+      // TODO: Change this to an actual preview
+      preview: 'This is a preview of message B',
+      csvFilename: file.name,
+    })
   } catch (err) {
     // if there is an error, return a csv status with the error message
     return Promise.resolve({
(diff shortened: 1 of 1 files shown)


frontend/src/services/validate.service.ts
```diff
@@ -0,0 +1,137 @@
...
+        // If there is errors, append to the errors array
+        if (step.errors.length != 0) {
+          reject(`Error parsing file.`)
+        }
+        // Check params
+        const params = step.data
+
+        try {
+          const paramKeys = keys(params)
+          checkTemplateKeys(paramKeys, templateParams)
```
lshaowei18, line 97: Ideally i would like to template it instead of just checking if the params contain the keys. I am also kinda worried about the performance, since for each row we are doing `keys(params) and then finding out the difference between each array.

lshaowei18:
Merged into  feature/protected-email-upload

PR #495, opened 2020-07-07
[Registered] chore: Refactor backend upload service

lshaowei18:
## Solution

Moved the multi-upload functions into another file.


lshaowei18:
@pregnantboy 
`Refactor upload.service.ts to upload-multi.service.ts (extends/exports upload.service.ts)`
Not sure if this is what you are looking for, let me know if i understood it wrongly.

backend/src/core/services/upload.service.ts
```diff
@@ -14,6 +14,13 @@ import { Campaign } from '@core/models'
 import { CsvStatusInterface } from '@core/interfaces'
 import { CSVParams } from '@core/types'
 
+import {
```
lshaowei18, line 17: Not sure if this is a good pattern, was also considering creating a new service called `MultiUploadService`

lshaowei18:
Too many changes to `registered-mail`, will reference this and make another PR.

PR #503, opened 2020-07-08
Verify password endpoint for protected campaigns

### LINKED ISSUE #368: [Registered]  Verify password hash route
New backend route:
- verifies password hash matches (Recommend using https://nodejs.org/api/crypto.html#crypto_crypto_timingsafeequal_a_b to mitigate timing attacks)
- returns encrypted payload

TBC: store read count

miazima:
## Problem

The password entered by the user on to decrypt the protected message needs to be verified by the server before sending over the encrypted payload. 

Closes #368 

**Features**:

- `verifyPasswordHash` endpoint verifies the password hash and returns the encrypted payload on success, or returns `401 Unauthorized` on mismatch
- compare the input password hash with the password hash stored in the `protected_messages` table
- use `crypto.timingSafeEqual` function to compare the hashes to prevent timing attacks


backend/src/email/routes/email-campaign.routes.ts
```diff
@@ -637,4 +644,49 @@ router.get('/stats', EmailStatsMiddleware.getStats)
  */
 router.get('/export', EmailStatsMiddleware.getFailedRecipients)
 
+/**
+ * @swagger
```
lshaowei18, line 648: Hmmm swagger documentation doesn't seem to match. Will the `passwordHash` be in the body or params?

backend/src/core/middlewares/protected.middleware.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @param res
+ * @param next
+ */
+const verifyPasswordHash = async (
+  req: Request,
+  res: Response,
+  next: NextFunction
+): Promise<Response | void> => {
+  try {
+    const { campaignId, uuid, passwordHash } = req.params
```
lshaowei18, line 16: I think we should be getting `passwordHash` from `req.body`

backend/src/core/routes/index.ts
```diff
@@ -78,6 +79,7 @@ const router = Router()
 router.use('/ping', ping)
 router.use('/auth', authenticationRoutes)
 router.use('/stats', statsRoutes)
+router.use('/protect', protectedCampaignRoutes)
```
lshaowei18, line 82: hmmm are we doing `/protect` or `/p`? Could we use `config.get('protectedPath')`?

PR #508, opened 2020-07-09
chore: halt only on hard bounce

jeantanzj:
## Problem

Soft bounces should not halt the campaign as they do not contribute to the bounce rate computed by SES.

## Solution

Halt only when error_code=Hard bounce.


jeantanzj:
I released this separately so that the users could resume their campaigns

PR #509, opened 2020-07-09
[Registered] CSV Transform service updates

jeantanzj:
For your perusal - you might want to just copy paste relevant parts.

Main changes
- Chunk size calculation
- Get all presigned urls on /upload/start instead of getting a url every chunk
- Uploading a buffer instead of string


lshaowei18:
Thanks Jean. I merged `csv-transform-service` into `registered-mail` so will copy and make a new PR

frontend/src/services/protect-csv.service.ts
```diff
@@ -36,6 +36,49 @@ async function transformRows(
...
+      },
+      complete: function (results) {
+        // We can also check for recipient and password here.
+        // results.data will contain 1 row of results because we aborted on the first step
+        const row = Object.entries(results.data) as Array<[string, string]>
+        for (const [field, value] of row) {
+          if (/^[a-zA-Z0-9\s-_'"/]+$/.test(field)) {
+            // field is of an acceptable format
+            const regexp = new RegExp(`\\s+${field}\\s+`, 'gi')
+            const numReplacements = template.match(regexp)?.length || 0
```
lshaowei18, line 74: Should we use the templating module instead of regexp? For example, the template could be `hi {{recipient}}, your secret code is {{ password }}` and the regex will only catch `password` and not `recipient`.

PR #510, opened 2020-07-10
Fix eslint warning

xming13:
## Problem

This PR fixes eslint warnings in existing frontend code. It does not close any issue.

## Solution

I make changes based on the recommendation in
https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md
https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/alt-text.md.

Also,
- updated `ProgressPane` to use appropriate tag and attribute for a11y
- added `type="button"` to all `button` element for consistency.

**Improvements**:

- a11y improvement
- No more compiled with warnings! :)

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image


frontend/src/styles/app.scss
```diff
@@ -70,7 +70,49 @@ h5 {
   }
 }
 
-a {
+// reset default button style
+button {
```
jeantanzj, line 74: @pregnantboy would you rather that the `a` links be converted into a button component? Seems like the style you were going for.

PR #517, opened 2020-07-13
Get encrypted payload from backend

miazima:


frontend/src/services/crypto.service.ts
```diff
@@ -116,3 +116,13 @@ export async function decryptData(
     throw new Error(`Error decrypting data: ${error.message}`)
   }
 }
+
+export async function deriveHashedPassword(
```
pregnantboy, line 120:
```suggestion
export async function hashPassword(
```

frontend/src/services/crypto.service.ts
```diff
@@ -116,3 +116,13 @@ export async function decryptData(
...
+  password: string,
+  salt: string
+): Promise<string> {
+  const cryptoKey = await deriveKey(password, salt)
+  const key = await exportKey(cryptoKey)
+  const hash = sha256(key)
```
pregnantboy, line 126:
```suggestion
  const hash = await sha256(key)
```

frontend/src/services/decrypt-mail.service.ts
```diff
@@ -1,4 +1,14 @@
-import { decryptData } from './crypto.service'
+import axios from 'axios'
+import { decryptData, deriveHashedPassword } from './crypto.service'
+
+async function getEncryptedPayload(
+  id: string,
+  hashedPassword: string
+): Promise<string> {
+  return axios.post(`/protect/${id}`, { hashedPassword }).then((response) => {
```
pregnantboy, line 8: async await

frontend/src/services/decrypt-mail.service.ts
```diff
@@ -1,4 +1,14 @@
...
+    return response.data?.encryptedPayload
```
pregnantboy, line 9: snake case

PR #525, opened 2020-07-13
chore: Prevent href link from being sanitized by xss

### LINKED ISSUE #516: Href gets sanitized when the link is a keyword
**Describe the bug**

In the email template, when you have the href as a keyword like `<a href="{{protectedlink}}">link</a>`, it gets sanitised into `<a href>link</a>`.

**To Reproduce**
recording

**Expected behavior**
Should retain the keyword. This affects registered mail.

lshaowei18:
## Problem

Closes #516 

## Solution

- Make use of `safeAttrValue` and prevent keywords in href to be sanitised.

- Allow https links and keywords in img src

recording (1)

Backend new dependencies:
- `xss`: To be able to call the default `safeAttrValue` from the lib

### DIFF
--- a/backend/src/core/config.ts
+++ b/backend/src/core/config.ts
@@ -355,6 +355,22 @@ const config = convict({
           img: ['src', 'alt', 'title', 'width', 'height'],
         },
         stripIgnoreTag: true,
+        // Overriding the default safeAttrValue
+        safeAttrValue: (tag: string, name: string, value: string): string => {
+          const keywordRegex = /{{\s*?\w+\s*?}}/g
+          // Do not sanitize keyword when it's a href link, eg: <a href="{{protectedlink}}">link</a>
+          if (tag === 'a' && name === 'href' && value.match(keywordRegex)) {
+            return value
+          }
+
+          // Do not sanitize keyword when it's a img src, eg: <img src="{{protectedImg}}">
+          if (tag === 'img' && name === 'src' && value.match(keywordRegex)) {
+            return value
+          }
+          // The default safeAttrValue does guard against some edge cases
+          // https://github.com/leizongmin/js-xss/blob/446f5daa3b65e9e8f0e9c71276cf61dad73d1ec3/dist/xss.js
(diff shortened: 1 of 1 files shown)


backend/src/core/config.ts
```diff
@@ -332,6 +332,16 @@ const config = convict({
           img: ['src', 'alt', 'title', 'width', 'height'],
         },
         stripIgnoreTag: true,
+        safeAttrValue: (
+          tag: string,
+          name: string,
+          value: string
+        ): string | void => {
+          // Do not sanitize keyword when it's a href link, eg: <a href="{{protectedlink}}">link</a>
```
jeantanzj, line 340: Could you do the same for img src?

backend/src/core/config.ts
```diff
@@ -332,6 +332,32 @@ const config = convict({
...
+              } catch (_) {
+                return
+              }
+              return value
+            }
+            // Allow keywords
+            if (value.startsWith('{{')) {
+              return value
+            }
+          }
```
pregnantboy, line 359: return value

backend/src/core/config.ts
```diff
@@ -332,6 +332,32 @@ const config = convict({
           img: ['src', 'alt', 'title', 'width', 'height'],
         },
         stripIgnoreTag: true,
+        safeAttrValue: (
+          tag: string,
+          name: string,
+          value: string
+        ): string | void => {
```
pregnantboy, line 339: should always return a string

backend/src/core/config.ts
```diff
@@ -332,6 +332,32 @@ const config = convict({
...
+          // Do not sanitize keyword when it's a href link, eg: <a href="{{protectedlink}}">link</a>
+          if (tag === 'a' && name === 'href' && value.startsWith('{{')) {
+            return value
```
pregnantboy, line 342: can we use the regex  for highlighting

lshaowei18:
Replaced by #551

PR #526, opened 2020-07-14
[Registered] chore: Template A checks

lshaowei18:
## Problem

Multiple `protectedlink` would throw an error:
recording (4)

Template A's subject was allowed to have keywords:
recording (5)

## Solution

**Improvements**:

- Added checks for subject to ensure it has no keywords. (We could allow them to add `protectedlink`)
- Allow multiple mentions of `protectedlink` in `checkTemplateBody`

## After Screenshots

**AFTER**:
Multiple `protectedlink`:
recording (2)

Keywords in subject:
recording (3)

### DIFF
--- a/backend/src/core/services/protected.service.ts
+++ b/backend/src/core/services/protected.service.ts
@@ -70,17 +70,21 @@ const storeProtectedMessages = async (
 const checkTemplateVariables = (body: string): void => {
   const { variables } = templateClient.parseTemplate(body)
 
+  const unique = [...new Set(variables)]
+
   const essential = ['protectedlink']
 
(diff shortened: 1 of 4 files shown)


backend/src/core/services/protected.service.ts
```diff
@@ -66,27 +66,44 @@ const storeProtectedMessages = async (
 }
 /**
  * Verifies that the template for protected campaigns has the compulsory keywords
+ * The template should not contain any other keywords other than the compulsory ones.
  */
-const checkTemplateVariables = (body: string): void => {
+const checkTemplateBody = (body: string): void => {
```
pregnantboy, line 71: can just use the same function to check both body and subject

PR #542, opened 2020-07-14
feat: Force custom Telegram keyboards to be shown

dvrylc:
## Summary

From the comment in `generatePadding`: 

```
 * If a user has previously hidden/dismissed the custom keyboard, some Telegram clients (e.g. iOS)
 * continue hiding the custom keyboard sent with subsequent messages if there are no changes
 * detected in the button labels. A random-length whitespace string is padded to the button labels
 * to force clients to always show the custom keyboard. This has no visible effect to users because
 * Telegram trims whitespace before it reaches the user.
```

## Tests

1. Run the handler locally
2. Send `/start` then hide the custom keyboard
3. Continue sending `/start` – Each time a response is received, the custom keyboard should be brought up again.


lamkeewei:
Tested on Android and the padding doesn't seem to be trimmed though...

PR #543, opened 2020-07-14
feat: Use full bot token instead of bot ID in Telegram webhook URLs

dvrylc:
## Problem

We currently use bot ID as the path parameter for bot webhook URLs. Anyone who knows the bot ID (which is not meant to be a secret) can start sending updates to our handler. 

Although an attacker can't make use of this because we have additional checks, ideally our handler should only be handling updates from trusted sources (i.e. from Telegram only and not strangers). 

## Solution

- Extend the path parameter to include the full bot token instead of just the bot ID. The bot token is a secret that is only known to the bot owner. 
- Verify that the incoming bot token is valid and not revoked with a call to Telegram's `getMe` endpoint. 

## Tests

1. Save a new bot credential in the app
    - Your bot's webhook URL should be set to `telegramOptions.webhookUrl` + the full bot token. Verify this via `GET https://api.telegram.org/bot<token>/getWebhookInfo`. 
2. Run the handler locally and go through the bot subscription flow.
    - The handler should be able to handle the new path parameter
    - You should see `credentials: Bot token verified.` in the handler logs - this means that the received token has been validated by Telegram. 

## Deploy Notes

- API Gateway: replace the current `/{botId}` resource with `/{botToken}`
- Execution role: 
    - Travis: Delete `$PRODUCTION_TELEGRAM_HANDLER_ROLE` and `$STAGING_TELEGRAM_HANDLER_ROLE` since the Lambda can now use the generic execution role
    - Lambda: Set the execution role to the generic one, then delete the Telegram-related roles
- Update all current bots to use the new webhook URL scheme

### DIFF
--- a/serverless/telegram-handler/src/credentials.ts
+++ b/serverless/telegram-handler/src/credentials.ts
@@ -18,7 +18,7 @@ export const verifyBotIdRegistered = async (
   botId: string,
   sequelize: Sequelize
 ): Promise<void> => {
-  const { exists: botIdExists } = await sequelize.query(
+  const botIdExists = await sequelize.query(
     `SELECT 1 AS exists FROM credentials WHERE name = :botId;`,
     {
       replacements: {
--- a/serverless/telegram-handler/src/credentials.ts
+++ b/serverless/telegram-handler/src/credentials.ts
@@ -1,5 +1,6 @@
 import AWS from 'aws-sdk'
 import { Sequelize } from 'sequelize-typescript'
+import crypto from 'crypto'
 
 import config from './config'
 import { Logger } from './utils/logger'
@@ -63,8 +64,15 @@ export const getBotTokenFromId = async (botId: string): Promise<string> => {
  * Compares two bot tokens and throws an error if they do not match.
  */
 export const verifyBotToken = (token1: string, token2: string): void => {
-  if (token1 !== token2) {
+  // `crypto#timingSafeEqual` throws an error if the input buffers are of different length
+  // and returns false if they are of the same length but different contents. Handle both cases.
+  try {
+    if (!crypto.timingSafeEqual(Buffer.from(token1), Buffer.from(token2))) {
+      throw new Error()
+    }
+
+    logger.log('Bot token verified')
+  } catch (err) {
     throw new Error('Bot token mismatch')
   }
-  logger.log('Bot token verified')
 }
--- a/serverless/telegram-handler/src/credentials.ts
+++ b/serverless/telegram-handler/src/credentials.ts
@@ -58,3 +58,13 @@ export const getBotTokenFromId = async (botId: string): Promise<string> => {
 
   return botToken
 }
+
+/**
+ * Compares two bot tokens and throws an error if they do not match.
+ */
+export const verifyBotToken = (token1: string, token2: string): void => {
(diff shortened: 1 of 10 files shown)


serverless/telegram-handler/src/credentials.ts
```diff
@@ -58,3 +58,13 @@ export const getBotTokenFromId = async (botId: string): Promise<string> => {
 
   return botToken
 }
+
+/**
+ * Compares two bot tokens and throws an error if they do not match.
+ */
+export const verifyBotToken = (token1: string, token2: string): void => {
+  if (token1 !== token2) {
```
lamkeewei, line 66: Should probably use a [timing safe comparison method](https://nodejs.org/api/crypto.html#crypto_crypto_timingsafeequal_a_b) to check equality of bot token. [Reference on timing attacks](https://codahale.com/a-lesson-in-timing-attacks/).

lamkeewei:
I'm thinking if we should implement some kind of monitoring/alerting. In particular, I think the following two scenarios might be worth monitoring: 
1. `bot_id` that are not registered in Postman
2. `bot_token` that does not match the stored token

Given that these two scenarios would be unlikely to occur during normal operations, it could be useful to give us an early warning to check things out. We can potentially use Sentry to capture these events.

serverless/telegram-handler/src/credentials.ts
dvrylc: Learned something new, thanks for pointing this out!

lamkeewei:
Given this scenario: 
1. Bot token is compromised by the user (leaked on internet, etc) 
2. User gets a new bot token from BotFather but has yet to update Postman with the new token (delete old token and insert new token)

During this window, the adversary can potentially still use the old bot token to update/override phone numbers. Because we only check against what’s stored in our SecretsManager and not verify with the Telegram Bot API for the validity of the token. While the eventual `ctx.reply` would fail because the API call to Telegram Bot API will be rejected, the phone number has already been upserted. 

We can potentially solve this by catching the error for `ctx.reply` and rolling back the database transaction for the upsert.

dvrylc:
> This is because we only check against what’s stored in our SecretsManager and not verify with the Telegram Bot API for the validity of the token. 

Perhaps we should do this instead? Right after instantiating the Telegraf instance, we fire off a call to `getMe` before actually handling the update. If the call fails, we can drop the update and avoid getting to a stage where we have to rollback.

lamkeewei:
> Perhaps we should do this instead? Right after instantiating the Telegraf instance, we fire off a call to `getMe` before actually handling the update. If the call fails, we can drop the update and avoid getting to a stage where we have to rollback.

Yep that will work too. Was thinking of saving an extra API call. If we're doing this, do we still need to fetch and compare the stored token from SecretsManager?

dvrylc:
As discussed during sync-up: 

> If we're doing this, do we still need to fetch and compare the stored token from SecretsManager?

Replaced the Secrets Manager check with a call to `getMe`. Have also removed the `aws-sdk` dependency and config since nothing else uses it. `.travis.yml` has also been updated to deploy the Lambda with the generic execution role. 

> We can potentially use Sentry to capture these events.

Good idea, we'll tackle the Sentry-related changes in a separate PR.

PR #544, opened 2020-07-15
chore: Retry getting s3 object in backend

### LINKED ISSUE #379: Retry getting s3 object
Related to https://github.com/opengovsg/postmangovsg/pull/371

S3 supports atomic writes only when a new object is inserted. When overriding a file of the same name, the subsequent PUT will be eventually consistent. The new version of the file cannot be retrieved at that point. Presumably, this occurs when the same pre-signed url is used to upload multiple files, causing them to override the previous version.

The current stop-gap fix throws an error when the file cannot be retrieved, causing the user to have to re-upload their file.

lshaowei18:
## Problem

Closes #379

## Solution

Made use of `async-retry` to retry only when s3 fails to find the specified key. There are other errors that could be thrown from `parseCsv`, so I had to check the error code.

Alternatively, I can do the retry on the frontend by reading the error code given by the backend after calling `upload/complete`. If it is the error message related to this, I will retry using axios.

Not sure which approach is better tbh. 

## Tests

- [ ] When an invalid s3 key is provided, backend will retry before throwing 500.
- [ ] Other parsing errors thrown by parseCsv will not result in any retries.

## Deploy Notes

New dependencies for backend.

**New dependencies**:

- `async-retry` : For async retries operations

**New dev dependencies**:

- `@types/async-retry`

### DIFF
--- a/backend/src/core/services/s3-client.class.ts
+++ b/backend/src/core/services/s3-client.class.ts
@@ -6,6 +6,7 @@ import { configureEndpoint } from '@core/utils/aws-endpoint'
 
 import { CSVParams } from '@core/types'
 import { ParseCsvService } from '@core/services'
+import { CSVNotFoundError } from '@core/errors'
 
 const FILE_STORAGE_BUCKET_NAME = config.get('aws.uploadBucket')
 
@@ -36,7 +37,7 @@ export default class S3Client {
    * @param s3Key
    */
   async getCsvFile(s3Key: string): Promise<Array<CSVParams>> {
-    return await retry(
+    return retry(
       async (bail) => {
         try {
           const downloadStream = this.download(s3Key)
@@ -47,11 +48,11 @@ export default class S3Client {
             bail(e)
             return []
           }
-          throw e
+          throw new CSVNotFoundError()
         }
       },
       {
-        retries: 5,
+        retries: 3,
         minTimeout: 1000,
         maxTimeout: 3 * 1000,
         factor: 1,
--- a/backend/src/core/services/s3-client.class.ts
+++ b/backend/src/core/services/s3-client.class.ts
@@ -1,4 +1,5 @@
 import S3 from 'aws-sdk/clients/s3'
+import retry from 'async-retry'
 
 import config from '@core/config'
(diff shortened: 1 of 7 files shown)


backend/src/core/services/s3-client.class.ts
```diff
@@ -30,13 +31,31 @@ export default class S3Client {
...
+    return await retry(
+      async (bail) => {
+        try {
+          const downloadStream = this.download(s3Key)
+          const fileContents = await ParseCsvService.parseCsv(downloadStream)
+          return fileContents
+        } catch (e) {
+          if (e.code !== 'NoSuchKey') {
+            bail(e)
+            return []
```
lshaowei18, line 48: Not sure if this is a good pattern to be using. `retry` will eventually throw an error, even though i am returning an empty array. Initially i wanted to just return void, but this would mean i have to change the signature of the function.

lshaowei18:
Out of curiosity I tried to implement it on the frontend by returning and checking specific error message so that i can retry. The branch name is `frontend-retry-s3`

backend/src/core/services/s3-client.class.ts
jeantanzj: This is ok. If it can't get the csv file after retrying, then it should fail and require the user to re-upload.

jeantanzj:
> Out of curiosity I tried to implement it on the frontend by returning and checking specific error message so that we can choose when to retry. The branch name is `frontend-retry-s3`

I looked at that branch. If you prefer to retry on frontend, perhaps we could replace the global axios object instead (frontend/src/config.ts)?

lshaowei18:
> > Out of curiosity I tried to implement it on the frontend by returning and checking specific error message so that we can choose when to retry. The branch name is `frontend-retry-s3`
> 
> I looked at that branch. If you prefer to retry on frontend, perhaps we could replace the global axios object instead (frontend/src/config.ts)?

Tbh I'm not too sure which approach is better. What do you think?

Hmmm yup we could replace the global axios object; I wasn't exactly sure what the library was doing and was afraid that the retries will be set for all of the other requests.

PR #547, opened 2020-07-15
Replace CSVParse with papaparse

jeantanzj:
## Problem

For consistency between backend and frontend, replace CSVParse with papaparse. 

## Notes

The main difference between the parsed output of CSVParse and Papaparse is that Papaparse is able to parse a file where some columns are quoted, and some are not. CSVParse would throw an error parsing the header 
```
recipient,New York City,40°42'46"N,74°00'21"W\`,test
```
 due to extra quotes, but Papaparse would parse that as five columns
```
["recipient","New York City", "40°42'46\"N","74°00'21\"W`","test"]
```

However, in this scenario, if the csv file only had three columns, 
```
recipient,New York City,40°42'46"N,74°00'21"W\`,test
a,b,c
d,e,f
```
Papaparse would store an error of type `FieldMismatch`, code `TooManyFields`.
Similarly, if the csv file had more columns than headers, Papaparse would store an error of type `FieldMismatch`, code `TooFewFields`.


jeantanzj:
Superseded

PR #548, opened 2020-07-15
Refactor failed recipients check in Campaigns component

### LINKED ISSUE #535: Loading of campaign list is slow
**Describe the bug**
Loading each page of campaigns is slow

for loop is awaiting each promise in series
https://github.com/opengovsg/postmangovsg/blob/20d58642ecdf4cbcf1f4a98df427155e25e56bb9/frontend/src/components/dashboard/campaigns/Campaigns.tsx#L60-L63

**Suggestion**
Return hasFailedRecipient property in same api call that fetches campaign list

**Screenshots**
Screenshot 2020-07-14 at 5 55 37 PM

miazima:
## Problem

Loading of campaign list was slow - after calling the campaign list endpoint to get the list of campaigns, the stats endpoint was called to determine the existence of failed recipients for each campaign. This was blocking the display of the campaign list on the frontend hence resulting in slower loading times for campaign list.

Closes #535

## Solution

Return `has_failed_recipients` as a boolean with the get campaign list endpoint and remove the calls to the stats endpoint from Campaigns component.

### DIFF
--- a/frontend/src/classes/Campaign.ts
+++ b/frontend/src/classes/Campaign.ts
@@ -27,6 +27,7 @@ export class Campaign {
   isCsvProcessing: boolean
   statusUpdatedAt: Date
   protect: boolean
+  hasFailedRecipients: boolean
 
   constructor(input: any) {
     this.id = input['id']
@@ -40,6 +41,7 @@ export class Campaign {
     this.sentAt = input['sentAt']
     this.statusUpdatedAt = input['statusUpdatedAt']
     this.protect = input['protect']
+    this.hasFailedRecipients = input['has_failed_recipients']
   }
 
   getStatus(jobs: Array<{ status: string }>): Status {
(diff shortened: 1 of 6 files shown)


frontend/src/classes/Campaign.ts
```diff
@@ -40,6 +41,7 @@ export class Campaign {
     this.sentAt = input['sentAt']
     this.statusUpdatedAt = input['statusUpdatedAt']
     this.protect = input['protect']
+    this.hasFailedRecipients = input['has_failed_recipients']
```
lamkeewei, line 44: The backend returns the following payload:
```
{
   "campaigns":[
      {
         ...
         "statistic":{
            "has_failed_recipients":true
         }
      }
   ],
   "total_count":1
}
```

So `input['has_failed_recipients']` will be undefined.

miazima:
Closing this in favour of #580 which is based off this

PR #551, opened 2020-07-15
chore: add xss options into postman-templating module

### LINKED ISSUE #532: Extend postman-templating to include xss options
Currently the xss options are copied in multiple files.

Consolidate the xss options as constants in `postman-templating`. For example, one could `const templateClient = new TemplateClient(EMAIL_XSS)` where `EMAIL_XSS` is imported from `postman-templating`.

**Additional context**
Will also need to take note on how we are overriding `safeAttrValue` in `registered-mail` and `telegram`  and consider if there is any side effects.

lshaowei18:
## Problem

Closes #532, #516 

## Solution

To use the xss options :
`import { TemplateClient, XssEmailOption } from 'postman-templating'`
`const client = new TemplateClient(XssEmailOption)`

Initially we installed `xss` on backend and worker to use the default `safeAttrValue` from `xss`, now that we moved the options into postman-templating we can uninstall `xss`.

### DIFF
--- a/modules/postman-templating/test/template-client.test.ts
+++ b/modules/postman-templating/test/template-client.test.ts
@@ -1,6 +1,7 @@
 import { TemplateClient } from '../src/template-client'
 import { TemplateError } from '../src/errors'
-import xss from 'xss'
+
+import { XssEmailOption, XssSmsOption, XssTelegramOption } from '../src/xss-options'
 
 describe('template', () => {
   let templateClient: TemplateClient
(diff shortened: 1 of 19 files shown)


modules/postman-templating/test/template-client.test.ts
```diff
@@ -1,6 +1,7 @@
 import { TemplateClient } from '../src/template-client'
 import { TemplateError } from '../src/errors'
-import xss from 'xss'
+
+import { XssEmailOption, XssSmsOption, XssTelegramOption } from '../src/xss-options'
```
pregnantboy, line 4: can we capitalize and camel_case these constants

PR #558, opened 2020-07-16
fix: Unhandled promise rejection in multipart upload

lshaowei18:
## Problem

Unhandled promise rejections in `protectAndUploadCsv` causes browser to crash.

## Solution

I don't completely understand what is going on; it seems like within a Promise if any of the code throws an error and it is not handled by the Promise, it will be dealt as any other uncaught errors by the JS engine.

## Before & After Screenshots

**BEFORE**:
uncaught_error

**AFTER**:
recording (6)

### DIFF
--- a/frontend/src/services/protect-csv.service.ts
+++ b/frontend/src/services/protect-csv.service.ts
@@ -106,25 +106,34 @@ export async function protectAndUploadCsv(
         const rows = chunk.data
 
         // transform into rows of data into {recipient,payload,passwordHash,id}
-        const data = await transformRows(template, rows, partNumber)
+        try {
+          const data = await transformRows(template, rows, partNumber)
 
-        // Upload the single string onto s3 based through the presigned url
-        const etag = await uploadPartWithPresignedUrl({
-          presignedUrl: presignedUrls[partNumber - 1],
-          data,
-        })
-        etags.push(etag)
+          // Upload the single string onto s3 based through the presigned url
+          const etag = await uploadPartWithPresignedUrl({
+            presignedUrl: presignedUrls[partNumber - 1],
+            data,
+          })
+          etags.push(etag)
+        } catch (e) {
+          reject(e)
+        }
+
         parser.resume()
       },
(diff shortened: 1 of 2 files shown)


frontend/src/services/protect-csv.service.ts
```diff
@@ -106,25 +106,34 @@ export async function protectAndUploadCsv(
...
+          const data = await transformRows(template, rows, partNumber)
+
+          // Upload the single string onto s3 based through the presigned url
+          const etag = await uploadPartWithPresignedUrl({
+            presignedUrl: presignedUrls[partNumber - 1],
+            data,
+          })
+          etags.push(etag)
+        } catch (e) {
+          reject(e)
```
jeantanzj, line 119: To prevent the parser from continuing to parse,
```javascript
reject(e)
parser.abort()
```

frontend/src/services/protect-csv.service.ts
```diff
@@ -106,25 +106,34 @@ export async function protectAndUploadCsv(
...
-        await completeMultiPartUpload({
-          campaignId,
-          filename: file.name,
-          transactionId,
-          partCount: partNumber,
-          etags,
-        })
-        resolve()
+        try {
+          await completeMultiPartUpload({
```
jeantanzj, line 126:
```javascript
complete: async function(results) {
try{
if(!results.meta?.aborted){
...
}
}catch(e){
reject(e)
}
}
```

frontend/src/services/upload.service.ts
```diff
@@ -216,16 +216,20 @@ export async function uploadPartWithPresignedUrl({
...
+      new Blob(data, { type: contentType }),
+      {
+        withCredentials: false,
+        timeout: 0,
+      }
+    )
+    return response.headers.etag
+  } catch (e) {
+    errorHandler(e, 'Error uploading part with presigned url')
+  }
```
pregnantboy, line 232: can u also add a try catch to the beginMultipartUpload function

PR #569, opened 2020-07-16
chore: support underscore for keywords in template

### LINKED ISSUE #565: [Registered] Template B unable to handle capitalised keyword
**Describe the bug**
When the template contains capitalised keyword like `{{ADDRESS1}}`, it fails the validation of CSV, despite the csv having the header `ADDRESS1`.

**To Reproduce**

Use a csv that contains `ADDRESS1` as a header.

recording (7)

### LINKED ISSUE #566: Support for keyword with underscore in templates
Currently, underscores in keywords are not supported, ie: `{{POSTAL_CD}}`. For the MOE's use case, they have over 100+ column headers with underscores and it will be a pain for them to change. It is also quite ugly if we change things to `{{postmanCd}}`. 

Allow underscores in keywords.

lshaowei18:
## Problem

Closes #566 

## Solution

Add underscore to the the regex that is validating the template keys.

## After Screenshot

**AFTER**:
recording (8)

### DIFF
--- a/modules/postman-templating/src/template-client.ts
+++ b/modules/postman-templating/src/template-client.ts
@@ -93,11 +93,11 @@ export class TemplateClient {
             )
           }
 
-          // only allow alphanumeric template, prevents code execution
-          const keyHasValidChars = key.match(/[^a-zA-Z0-9]/) === null
+          // only allow alphanumeric, underscore template, prevents code execution
+          const keyHasValidChars = key.match(/[^a-zA-Z0-9\_]/) === null
           if (!keyHasValidChars) {
             throw new TemplateError(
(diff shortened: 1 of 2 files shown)


modules/postman-templating/src/template-client.ts
```diff
@@ -93,11 +93,11 @@ export class TemplateClient {
             )
           }
 
-          // only allow alphanumeric template, prevents code execution
-          const keyHasValidChars = key.match(/[^a-zA-Z0-9]/) === null
+          // only allow alphanumeric, underscore template, prevents code execution
+          const keyHasValidChars = key.match(/[^a-zA-Z0-9\_]/) === null
```
pregnantboy, line 97:
```suggestion
          const keyHasValidChars = key.match(/\W/) === null
```

PR #573, opened 2020-07-16
feature: Normalise SMS phone numbers with libphonenumber-js

### LINKED ISSUE #546: Standardise and use libphonenumber-js for SMS
Telegram feature uses `libphonenumber-js` to validate and normalise phone numbers to [E.164](https://www.twilio.com/docs/glossary/what-e164) format. Will be good to standardise and use this in the SMS also. 

Use phone number normalisation (#486)

lamkeewei:
## Problem

Closes #546 

## Solution
- Uses `PhoneNumberService` to normalise phone numbers before insertion into messages table (similar to Telegram)
- Remove appending of country code in `TwilioClient` for both `backend` and `worker`

## Tests
1. Run unit tests added in previous PR to confirm the validation and normalisation logic
2. Create an SMS campaign with the following recipients:
```
recipient
91234567
14128018888
+8613910998888
6580123456
```
3. Check in `sms_messages` table that all the messages should be normalised  
4. Upload a recipient list with the following numbers. Each of them should raise an error.
```
recipient
612345678   (Invalid length)
8123456     (Invalid length)
11234567    (Number not starting with 6, 8 or 9) 
```

## Deploy Notes
- Deploy workers after merge
- Removed `DEFAULT_COUNTRY_CODE` environment variable in `worker` config

### DIFF
--- a/backend/src/sms/services/sms-template.service.ts
+++ b/backend/src/sms/services/sms-template.service.ts
@@ -1,10 +1,12 @@
 import { difference, keys, chunk } from 'lodash'
 import { Transaction } from 'sequelize'
 
+import config from '@core/config'
 import logger from '@core/logger'
 import { isSuperSet } from '@core/utils'
-import { HydrationError } from '@core/errors'
+import { InvalidRecipientError, HydrationError } from '@core/errors'
 import { Campaign, Statistic } from '@core/models'
+import { PhoneNumberService } from '@core/services'
 import { TemplateClient, XSS_SMS_OPTION } from 'postman-templating'
 
 import { SmsTemplate, SmsMessage } from '@sms/models'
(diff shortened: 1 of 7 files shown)


backend/src/sms/services/sms-template.service.ts
```diff
@@ -226,11 +228,24 @@ const addToMessageLogs = async (
...
+/**
+ * Validate that recipient is a valid phone number and format it.
+ */
+const validateAndFormatNumber = (
   records: MessageBulkInsertInterface[]
-): boolean => {
-  const re = /^\+?[0-9]+$/
-  return records.some((record) => !re.test(record.recipient))
+): MessageBulkInsertInterface[] => {
+  return records.map((record) => {
```
chinying, line 237: Hmm I get what you're doing here, not sure if it may end up being confusing at some point because the individual `record` objects already get mutated even if you don't return them (line 247) in this map function.

lamkeewei: Oh right, the unintended side effects could be confusing. Would it be better if I returned a new instance of record instead of modifying it in place?

chinying: Yeah let's do that. 
The scenario I'm thinking of is:
```
let records = a
// validateAndFormatNumber gets invoked, mutating records
const formattedRecords = validateAndFormatNumber(records)

// I intended to use records as 'a' here, but it got mutated
someOtherFunction(records)
```

PR #576, opened 2020-07-17
Fix/update stats

### LINKED ISSUE #514: Stats are not updated by callbacks
**Describe the bug**
We knew about this issue. The dashboard should show the latest stats when callbacks are returned. 

**To reproduce**
Send a large number of messages. When callbacks return before the campaign is complete, stats will be updated. When callbacks return after the campaign is complete, stats is not updated.

**Possible solutions**
- Callback updates stats (this is io heavy)
- Manual refresh button
- Refresh on export
- ?

chinying:
## Problem

Closes #514 

## Solution

* Added a `refresh-stats` endpoint, that updates that `statistics` table with the entries from the `<channel>_messages` table. Also added a manual refresh button the frontend stats dashboard that allows this endpoint to be called. 
I went with a separate endpoint instead of a query param in the end cloudflare didn't allow for a rate limiting rule with just a query param. I've also opted to carry out rate limiting at the Cloudflare level over express and nginx as it wasn't API server instance specific (ie. I don't have to handle stuff like "did this IP hit my other API instance already"), and I just simply don't feel like messing with nginx if I can avoid it. 

* Stats also get updated from the messages table when `/export` is run.

## Before & After Screenshots

**AFTER**:
How the refresh stats button appears on the stats dashboard:
Screenshot 2020-07-20 at 1 19 04 PM

## Deployment Notes
* Turn on rate limiting rules for prod, on `/refresh-stats` and `/export`

### DIFF
--- a/backend/src/core/interfaces/campaign.interface.ts
+++ b/backend/src/core/interfaces/campaign.interface.ts
@@ -59,6 +59,7 @@ export interface CampaignStatsCount {
   unsent: number
   sent: number
   invalid: number
+  updatedAt?: Date
 }
 
 export interface CampaignInvalidRecipient {
--- a/backend/src/core/services/stats.service.ts
+++ b/backend/src/core/services/stats.service.ts
@@ -246,4 +246,5 @@ export const StatsService = {
   getNumRecipients,
   setNumRecipients,
   getFailedRecipients,
+  updateStats,
 }
--- a/backend/src/core/services/stats.service.ts
+++ b/backend/src/core/services/stats.service.ts
@@ -250,6 +250,7 @@ const getFailedRecipients = async (
   campaignId: number,
   logsTable: any
 ): Promise<Array<CampaignInvalidRecipient> | undefined> => {
+  await updateStats(+campaignId)
   const data = await logsTable.findAll({
     raw: true,
     where: {
--- a/backend/src/core/services/stats.service.ts
+++ b/backend/src/core/services/stats.service.ts
@@ -1,76 +1,19 @@
-import { fn, cast, Transaction, Op, QueryTypes } from 'sequelize'
+import { fn, cast, Transaction, Op } from 'sequelize'
 import { Statistic, JobQueue, Campaign } from '@core/models'
 import {
   CampaignStats,
   CampaignStatsCount,
   CampaignInvalidRecipient,
 } from '@core/interfaces'
-import { MessageStatus, ChannelType } from '@core/constants'
-import logger from '@core/logger'
-
-/**
- * Calls update_stats SQL function for respective channels.
- * fails silently
- * @param campaignId
- */
-const updateStats = async (campaignId: number): Promise<void> => {
-  const campaign = await Campaign.findOne({
-    where: { id: campaignId },
-  })
-  if (!campaign) return
-
-  logger.info(`updateStats invoked for campaign ${campaignId}`)
-
-  switch (campaign.type) {
-    case ChannelType.Email: {
-      await Statistic.sequelize?.query(
-        'SELECT update_stats_email(:campaign_id);',
-        {
-          replacements: { campaign_id: campaignId },
-          type: QueryTypes.SELECT,
-        }
-      )
-      break
-    }
-    case ChannelType.SMS: {
-      await Statistic.sequelize?.query(
-        'SELECT update_stats_sms(:campaign_id);',
-        {
-          replacements: { campaign_id: campaignId },
-          type: QueryTypes.SELECT,
-        }
-      )
-      break
-    }
-    case ChannelType.Telegram: {
-      await Statistic.sequelize?.query(
-        'SELECT update_stats_telegram(:campaign_id);',
-        {
-          replacements: { campaign_id: campaignId },
-          type: QueryTypes.SELECT,
-        }
-      )
-      break
-    }
-    default: {
-      logger.error(`Invalid channel type found for campaign ${campaignId}`)
-    }
-  }
-}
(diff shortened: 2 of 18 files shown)


backend/src/core/services/stats.service.ts
```diff
@@ -1,19 +1,102 @@
...
+      )
+      break
+    }
+    default: {
+      logger.error(`Invalid channel type found for campaign ${campaignId}`)
+    }
+  }
+}
+
+const consolidateStats = async (): Promise<void> => {
```
jeantanzj, line 74: Is this unused? why?

chinying: Oh unused for now. This was the query I wanted to use to fetch campaigns that stats that have not been updated for the last 36 hours. 

Actually let me go ahead and remove this, since we now want a job to be scheduled for X hours after the campaign is sent.

backend/src/core/services/stats.service.ts
```diff
@@ -1,19 +1,76 @@
...
+ */
+const updateStats = async (campaignId: number): Promise<void> => {
+  const campaign = await Campaign.findOne({
+    where: { id: campaignId },
+  })
+  if (!campaign) return
+
+  logger.info(`updateStats invoked for campaign ${campaignId}`)
+
+  switch (campaign.type) {
```
jeantanzj, line 24: Could we place these in their respective *channel*-stats.service.ts please

backend/src/core/interfaces/campaign.interface.ts
```diff
@@ -59,6 +59,7 @@ export interface CampaignStatsCount {
   unsent: number
   sent: number
   invalid: number
+  updatedAt?: Date
```
jeantanzj, line 62: Why optional

backend/src/sms/middlewares/sms-stats.middleware.ts
```diff
@@ -12,8 +12,9 @@ const getStats = async (
   next: NextFunction
 ): Promise<Response | void> => {
   const { campaignId } = req.params
+  const { refresh } = req.query
   try {
-    const stats = await SmsStatsService.getStats(+campaignId)
+    const stats = await SmsStatsService.getStats(+campaignId, refresh)
```
jeantanzj, line 17: Could we break this into
```javascript
if(refresh) await SmsStatsService.refreshStats(campaignId)
const stats = await SmsStatsService.getStats(campaignId)
```
Do you recall our query param `?reallySend=true`? This alarmingly reminds me of it, because it's not clear immediately what the effect of the `refresh` parameter is.

frontend/src/components/common/progress-details/ProgressDetails.tsx
```diff
@@ -201,6 +200,16 @@ const ProgressDetails = ({
           </tr>
         </tbody>
       </table>
+
+      <div className={styles.statsLastUpdated}>
```
jeantanzj, line 204: Could we only show this button when the campaign is complete, since we poll the stats while the campaign is in progress?

backend/src/email/routes/email-campaign.routes.ts
```diff
@@ -576,7 +586,7 @@ router.post(
  *        "500":
  *           description: Internal Server Error
  */
-router.get('/stats', EmailStatsMiddleware.getStats)
+router.get('/stats', celebrate(statsValidator), EmailStatsMiddleware.getStats)
```
jeantanzj, line 589: Remember to set a ratelimit on this route pattern with ?refresh=true.

backend/src/core/services/stats.service.ts
chinying: The reason why it's here is because if we ever have a consolidation script, you want to call all channels anyway, and you're just moving the switch case elsewhere?

I'm thinking from the POV of a cron job, if we need to find all campaigns where we need to run this updateStats function, will still need a `switch ... case` on the channel type to know which function to call right?

backend/src/sms/middlewares/sms-stats.middleware.ts
chinying: Alright, I can. Although arguably this reeks more of a param / variable naming issue

backend/src/core/interfaces/campaign.interface.ts
chinying: Because the return type here https://github.com/opengovsg/postmangovsg/blob/f65fa61481636350bffdd9bc90108567a1163792/backend/src/core/services/stats.service.ts#L76
doesn't have an `updatedAt`

backend/src/core/services/stats.service.ts
jeantanzj: The entire purpose of structuring our code base into core, email, sms, telegram is that core doesnt need to know anything about types. That being said, we should also extract getTwilioCredential and getTelegramCredential from backend/src/core/services/credential.service.ts

backend/src/sms/middlewares/sms-stats.middleware.ts
jeantanzj: See https://martinfowler.com/bliki/FlagArgument.html

jeantanzj:
The moment that refresh stats is clicked pm, the export button starts spinning. I think it's an unintended side effect? 
Screenshot 2020-07-29 at 7 54 04 AM

chinying:
> And maybe refresh stats button should disappear after X days have passed?
Hmm I think this is a nice to have but not necessary for now, since we're rate limiting anyway. And I will want the option to be there in case callbacks somehow come back after X days, until we can determine what is the proper duration this ought to be.

PR #583, opened 2020-07-20
feat: Append unsubscribe url to email message body

### LINKED ISSUE #579: [Unsubscribe] Append unsubscribe links during email send
- Before sending every email, generate the unsub link using the channel type and `message_id` and prepend/append it to the message body

lshaowei18:
## Problem

Closes #579

## Solution

Initially I thought that appending unsubscribe url through the workers wasn't a very good idea because:
- User won't be able to see it in the test message (We don't use the workers to send the test message)
- Single responsibility; workers should focus on sending messages and not editing messages

However, after considering other ways to achieve this like adding it to every email template & storing the url as a param, I find that the other ways are significantly more complicated.

We could append a test url to the body for the test message if we wanted to.

**Features**:

- Calculate hash, create unsubscribe url and append it to the hydrated message body.

**Improvements**:

- Added an interface named `Message` to make `message-worker/index.ts` neater

Screenshot 2020-07-22 at 3 19 30 PM

### DIFF
--- a/worker/src/core/loaders/message-worker/email.class.ts
+++ b/worker/src/core/loaders/message-worker/email.class.ts
@@ -67,7 +67,7 @@ class Email {
       .then((result) => map(result, 'get_messages_to_send_email'))
   }
 
-  calculateHmac(campaignId: number, recipient: string) {
+  calculateHash(campaignId: number, recipient: string): string {
     const version = config.get('unsubscribeHmac.version')
     return crypto
       .createHmac(
@@ -80,7 +80,7 @@ class Email {
 
   generateUnsubLink(campaignId: number, recipient: string): URL {
     const version = config.get('unsubscribeHmac.version')
-    const hmac = this.calculateHmac(campaignId, recipient)
+    const hmac = this.calculateHash(campaignId, recipient)
     const link = new URL(`/unsubscribe/${version}`, config.get('frontendUrl'))
     link.searchParams.append('c', campaignId.toString())
     link.searchParams.append('r', recipient)
--- a/worker/src/core/loaders/message-worker/email.class.ts
+++ b/worker/src/core/loaders/message-worker/email.class.ts
@@ -56,6 +56,7 @@ class Email {
       body: string
       subject: string
       replyTo: string | null
+      campaignId: number
     }[]
   > {
     return this.connection
--- a/worker/src/core/loaders/message-worker/email.class.ts
+++ b/worker/src/core/loaders/message-worker/email.class.ts
@@ -96,7 +96,7 @@ class Email {
       For more information, please visit our <a href="https://guide.postman.gov.sg/faqs/faq-recipients" style="color:#edae49" target="_blank">site</a>. 
     </p>
     <p style="font-size:12px;color:#818284;line-height:2em">
-      If you wish to unsubscribe from similar emails from your sender, please click <a href="${unsubUrl}" style="color:#edae49" target="_blank">here</a
+      If you wish to unsubscribe from similar emails from your sender, please click <a href="${unsubUrl}" style="color:#edae49" target="_blank">here</a>
       to unsubscribe and we will inform the respective agency.
     </p>
     `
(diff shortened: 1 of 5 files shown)


lamkeewei:
> Initially I thought that appending unsubscribe url through the workers wasn't a very good idea because:
> 
> * User won't be able to see it in the test message (We don't use the workers to send the test message)
> * Single responsibility; workers should focus on sending messages and not editing messages
> 
> However, after considering other ways to achieve this like adding it to every email template & storing the url as a param, I find that the other ways are significantly more complicated.

I think you've raised really valid points on doing the appending of the link on the workers. Do you have a rough sense what is the work required if we were to do this on the backend instead?

lshaowei18:
> > Initially I thought that appending unsubscribe url through the workers wasn't a very good idea because:
> > 
> > * User won't be able to see it in the test message (We don't use the workers to send the test message)
> > * Single responsibility; workers should focus on sending messages and not editing messages
> > 
> > However, after considering other ways to achieve this like adding it to every email template & storing the url as a param, I find that the other ways are significantly more complicated.
> 
> I think you've raised really valid points on doing the appending of the link on the workers. Do you have a rough sense what is the work required if we were to do this on the backend instead?

Hmmm the solution i can think of is to add the text to the template and an unique parameter to the params after we parse the csv:
1) Add `<br>Click here to unsubscribe: {{unsubscribeUrl}}` to the email's template.
2) Derive the unsub url and add it as a new parameter to the params. `unsubscribeUrl` cannot be used by the user in his/her csv.

The main problem with this approach is that the frontend checks, the parsing logics, will have to change because user cannot include `unsubscribeUrl` as a parameter. Will require alot of changes as well as decisions.

Some questions about the approach:
- Can the user provide a template that has`{{unsubscribeUrl}}`?
- What happens when the user have a column in their csv called `unsubscribeUrl`? Which value do we take or should we throw an error?

lshaowei18:
Currently i'm just appending plain text; when i try to add html the css in the html do not show in the email client. Not sure what is going on; will continue to look into it.

lshaowei18:
> Currently i'm just appending plain text; when i try to add html the css in the html do not show in the email client. Not sure what is going on; will continue to look into it.

So the css wasn't working because before we send out the email message we hydrate and xss filter the message. 

I believe we have two options here if we want to add styling: 
- Relax the xss filter to allow `<hr>` and `style` on `<p>`.  This would affect frontend, backend and worker. It would also mean that user inputs can have the whitelisted options.
- Append the html only after we hydrate the body.

I'm slightly in favour of the latter, and will update the PR to show the code looks like. Do let me know what you guys think.

worker/src/core/loaders/message-worker/email.class.ts
```diff
@@ -66,21 +67,54 @@ class Email {
       .then((result) => map(result, 'get_messages_to_send_email'))
   }
 
-  sendMessage({
+  calculateHmac(campaignId: number, recipient: string) {
```
lamkeewei, line 70: Should rename to `calculateHash`?

worker/src/core/loaders/message-worker/email.class.ts
```diff
@@ -66,21 +67,54 @@ class Email {
...
+    const hmac = this.calculateHmac(campaignId, recipient)
+    const link = new URL(`/unsubscribe/${version}`, config.get('frontendUrl'))
+    link.searchParams.append('c', campaignId.toString())
+    link.searchParams.append('r', recipient)
+    link.searchParams.append('h', hmac)
+    return link
+  }
+
+  appendUnsubToMessage(msg: string, unsubUrl: string): string {
+    return `${msg} <br> <hr> <p style="font-size:12px;color:#818284;line-height:2em">
```
lamkeewei, line 92: Petty would like to change the copywriting to the following:

[Postman.gov.sg](http://postman.gov.sg/) is a mass messaging platform used by the Singapore Government to communicate with stakeholders. For more information, please visit our [site](https://guide.postman.gov.sg/faqs/faq-recipients). 

If you wish to unsubscribe from similar emails from your sender, please click [here]() to unsubscribe and we will inform the respective agency.

lshaowei18:
> lgtm. Added a small fix for a broken link tag (`</a`).

Thanks for the fix!

PR #585, opened 2020-07-20
Improve export error list component

### LINKED ISSUE #482: Improve messaging on stats page
Users have a hard time figuring out the actions required for the stats page. A possible alternative is to include a simple phrase but the UI is getting progressively cluttered as we add more instructions. A tooltip could solve this pain point.

The specific feedback was
> Not sure what the export button is meant to do. I only found out that it is a list of bounced email after I downloaded it

### LINKED ISSUE #549: Improve design for campaigns table
## Feature description 
Discussed with Sarah and she suggested the following improvements to campaigns table:
1. Show the export button even when the export is not available to improve discoverability. There will be three states for the button: Preparing (before time limit is reached), Ready and No error. Refer to image below for what the three states will look like.
image
2. Increase width of campaigns table to 1040px 

## Mockup
image

miazima:
## Problem

The export interface is not intuitive to users, the general `Export`` label has caused confusion to some users who did not know what the function of the feature was until they downloaded the csv file.

Closes #549 and closes #482 

**Features**:

- Implement 3 states to the Export component - No Error, Loading and Ready
- Loading state is while waiting for `EXPORT_LINK_DISPLAY_WAIT_TIME` to elapse after campaign has been sent
- Improve instructions on Campaign stats page
- Minor changes to campign list css

## Before & After Screenshots

image


lshaowei18:
Hmmm if a campaign is paused, would the desired behavior be that it will just stay as a spinner with error list?

miazima:
> Hmmm what happens if a campaign is halted/paused after a minute?
On the stats page, when a campaign is halted/paused the export button is disabled until the campaign status eventually becomes `LOGGED` from `STOPPED`, and then the export button is enabled again.

chinying:
Hmm do you want to consider using alt text as well? You can afford to be more descriptive with alt text

frontend/src/components/common/export-recipients/ExportRecipients.tsx
```diff
@@ -48,19 +55,42 @@ const ExportRecipients = ({
...
     } finally {
       setDisabled(false)
     }
   }
 
+  function renderExportButton() {
+    if (exportStatus === CampaignExportStatus.NoError) {
+      return (
+        <span
+          className={cx(styles.noError, { [styles.alignLabel]: alignLabel })}
```
pregnantboy: is this alignLabel prop necessary?

backend/src/core/services/campaign.service.ts
```diff
@@ -76,6 +76,17 @@ const listCampaigns = ({
           ['updated_at', 'status_updated_at'],
         ],
       },
+      {
+        model: Statistic,
+        attributes: [
+          [
+            literal(
+              'CASE WHEN (errored + invalid) > 0 THEN True ELSE False END'
```
pregnantboy, line 84:
```suggestion
              '(errored + invalid) > 0'
```

pregnantboy: can u also change the literal above to ```cred_name IS NOT NULL```

miazima: This returns `{ statistics: null }` when the campaign is not logged, so I did this to conform it to `{ statistics: 
{ has_failed_recipients: true } }`, do you think returning null is clearer for a campaign which is not logged?

frontend/src/components/common/export-recipients/ExportRecipients.tsx
miazima: This is used to align the `No error` status in the campaign list component, since it doesn't have an icon and the other two statuses contain icons, I used a margin to align the text similar to the other statuses

backend/src/core/services/campaign.service.ts
pregnantboy: null case has been handled by this right?
```
getFailedRecipients(statistic: { has_failed_recipients: boolean }): boolean {
    return statistic && statistic.has_failed_recipients
  }
```

frontend/src/components/common/export-recipients/ExportRecipients.tsx
pregnantboy: Seems misaligned when there are no other statuses.

Screenshot 2020-07-23 at 10 37 51 PM

pregnantboy: On the other hand, centre-aligned seems ok for any scenario
Screenshot 2020-07-23 at 10 42 24 PM

pregnantboy: @sarahsalim thoughts?

backend/src/core/services/campaign.service.ts
miazima: yeah true, okay will replace the literal

frontend/src/components/common/export-recipients/ExportRecipients.tsx
sarahsalim: @pregnantboy @miazima Sure, centre-aligned is alright.

pregnantboy:
One last quirk 
button

frontend/src/components/common/export-recipients/ExportRecipients.tsx
```diff
@@ -3,23 +3,30 @@ import cx from 'classnames'
...
+  exportStatus,
+  iconPosition,
 }: {
   className?: string
   campaignId: number
   campaignName: string
   status: Status
   sentAt: Date
+  exportStatus: CampaignExportStatus
+  iconPosition: string
```
pregnantboy, line 26:
```suggestion
  iconPosition: 'left' | 'right'
```

pregnantboy:
Button should be disabled when unavailable

Issue #541, opened 2020-07-14
Prepare environment for MOM


jeantanzj:
- [ ] Check that MOM user has verified their email, and the DKIM/SPF records are set up
- [x] New db, same db instance
- [ ] New EB - modified SES_FROM
- [ ] New workers - modified SES_FROM
- [ ] Add parameter to email callback to differentiate between prod and MOM
- [ ] Should I care about twilio callback and telegram callback?

PR #590, opened 2020-07-21
feat: support separate environment for email callback

jeantanzj:
## Problem

Addresses https://github.com/opengovsg/postmangovsg/issues/541
SES can only publish events to one topic. We need a way to differentiate between an event that is meant to update the production db, or to update the db that we spun up separately for MOM. 

## Solution
The gist is that I added an environment variable to the X-SMTPAPI header to differentiate between events from different environments.

A potential solution was to pass two different database connection strings to the current lambda handling email callbacks. The advantage is that the same event only triggers one lambda. 

However, I chose to setup two lambdas instead because it is far easier to extend this to multiple environments (if we wanted production-03, production-04) without changing code. 
The disadvantage is that every SES event now triggers two lambdas. Pricewise, it's negligible (20 cents per mil requests). I'm slightly concerned about the number of concurrent executions (the max number of concurrent executions in the past week is 75 for the email lambda, 400 for the twilio lambda, and 2 for telegram), but we can still handle it for now. 

## Deploy notes
- [x] Add a subscription to this lambda (SES-SNS) for V1 USEast-1
- [x] Add a subscription to this lambda (SES-SNS) for V2 Sydney
- [x] Add a constraint on the new db to prevent creation of sms and telegram campaigns. 
```
CREATE FUNCTION disallow_sms_telegram_campaigns()
  RETURNS trigger AS
$func$
BEGIN
	IF (NEW."type" = 'SMS' or NEW."type"='TELEGRAM') THEN
      RAISE EXCEPTION 'SMS and TELEGRAM campaigns are not allowed in this environment';
   END IF;
   RETURN NEW;
END
$func$  LANGUAGE plpgsql;

CREATE TRIGGER disallow_type_before_insert_campaigns
BEFORE INSERT ON "campaigns"
FOR EACH ROW EXECUTE PROCEDURE disallow_sms_telegram_campaigns();
```
- [ ] When this branch is merged into master, cut momgovsg from master and treat it as a release branch from then on. 

## Tests
Test on mom.postman.gov.sg and api-mom.postman.gov.sg
**Be careful about sending many email messages** as the new lambda's code has not been deployed to production! The production lambda will try to update the status of messages using the message id. Fortunately, the first few email_messages in production are test messages, so we can still test a few messages, and truncate the email_messages table as necessary to keep the id number low.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,17 @@ jobs:
           bucket_name: "postmangovsg-elasticbeanstalk-appversion"
           on:
             branch: $PRODUCTION_BRANCH
+        - provider: elasticbeanstalk
+          edge: true
+          skip_cleanup: true
+          access_key_id: $AWS_ACCESS_KEY_ID
+          secret_access_key: $AWS_SECRET_ACCESS_KEY
+          region: $AWS_DEFAULT_REGION
+          app: "postmangovsg"
+          env: "postmangovsg-production-02"
+          bucket_name: "postmangovsg-elasticbeanstalk-appversion"
+          on:
+            branch: $PRODUCTION_02_BRANCH
     - name: worker
       before_install:
         - cd worker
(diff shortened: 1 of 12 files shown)


jeantanzj:
Given that we are using the same SES creds for two environments, should we adjust the rate down?

jeantanzj:
**Update: The assumption above was incorrect.** 
I can publish to two separate topics.

PR #591, opened 2020-07-21
Trim empty lines

pregnantboy:
## Problem

Templated messages may leave multiple empty lines if values for the keywords are blank. 

## Solution

Give users the option to trim empty lines when creating the template. (Only for protected messages now)
Basically combines consecutive linebreak tags `<br>` into one.

**Improvements**:

New checkbox component for frontend.

## Before & After Screenshots

Screenshot 2020-07-22 at 1 55 10 AM

**WITHOUT TRIMMING**
Screenshot 2020-07-22 at 1 59 26 AM
**WITH TRIMMING**
Screenshot 2020-07-22 at 1 59 35 AM

## Tests

Can be tested on `/test/template` page


jeantanzj:
Explain again, why `<b></b>` instead of the standard `<p></p>`?  
I think I would not know how to create new lines - perhaps we should get someone outside of Postman to try this component without explaining to them how it works.

pregnantboy:
<p> has a 1rem bottom margin which causes the empty line to be larger

jeantanzj:
> `<p>` has a 1rem bottom margin which causes the empty line to be larger

My main issue is that since <b></b> is not a conventional way of creating a new line, someone who googles it will not be able to understand its effect. It's an inline element. Could we consider other block level elements?

pregnantboy:
@jeantanzj changed to p tags

pregnantboy:
Merged with #609

Issue #578, opened 2020-07-17
[Unsubscribe] Unsubscribe confirmation page


lamkeewei:
`/unsubscribe/{version}?c=${campaign_id}&r=${recipient}&?h=${hmac}`
- ~~Check for design from Sarah~~ (Done)
- Non-protected route
- Version is in url params to maintain consistency with protected email
- Query backend on load to validate hmac and unsub status (status = `SUBSCRIBED`, otherwise user has already unsubscribed)
- Show confirm unsubscribe button if validation succeeds and message has not been unsubscribed to yet
- Update backend when confirmation button is pressed, show an error if fail

**Screen Designs**
Designs and assets are up on Zeplin. 

_Main unsubscribe page_
<img width="1440" alt="RequestUnSub" src="https://user-images.githubusercontent.com/3666479/88181840-fc006600-cc61-11ea-9ecb-ba512c2532f3.png">

_If user chooses "I'd rather stay"_
<img width="1440" alt="StaySubscribed" src="https://user-images.githubusercontent.com/3666479/88181850-01f64700-cc62-11ea-9e69-82bb5712d481.png">

_If user choose to unsubscribe_
<img width="1440" alt="ConfirmUnSub-2" src="https://user-images.githubusercontent.com/3666479/88181854-0458a100-cc62-11ea-9fd2-59d016305913.png">

PR #592, opened 2020-07-22
feat: Unsubscribe confirmation page

lshaowei18:
## Problem

closes: #578

## Solution

The main features for the page are: 
- ~When someone visits a url starting with `/unsubscribe/v1`, we make sure that the url is valid by checking the query through making a call to the backend. If the url is invalid, we will redirect the person to our homepage for now (we could have a 404 error page)~
- ~Once we verify that the url is valid, we can then decide what message to show on the page, depending on whether the person has unsubscribed or not.~
- When the unsubscribe button is clicked, we will make a post request to `/unsubscribe`, if it fails we will display the error message. If it is successful, we will hide the button and inform the user the unsub was successful.

Sarah's design: https://app.zeplin.io/project/5e7d5ae3d4169893da733825/screen/5f183d09defb6687560ab41b
alt

I think the current design is good (I really like the sad and happy face), but I am concerned that it might be confusing for the user. Having the `I'rather stay` button to be more eye catching than the `unsubscribe me` button is weird because the goal of the user is to unsubscribe. We should be making it as [easy](https://uxdesign.cc/empathy-driven-unsubscribe-experience-e55f22d9b5a7)  as possible to unsubscribe. 

[LinkedIn](https://uxdesign.cc/empathy-driven-unsubscribe-experience-e55f22d9b5a7) way of doing it feels like a better experience: 

linkedin1
linkin2

~Currently in this PR, the implementation looks like:~

recording (11)

~I would prefer this, but I have also done Sarah's design in another branch, `unsub-frontend-alt`.~
Implemented Sarah's design in this PR.


frontend/src/components/unsubscribe/Unsubscribe.tsx
```diff
@@ -0,0 +1,121 @@
...
+      return
+    } catch (e) {
+      setValid(false)
+    }
+  }
+
+  useEffect(() => {
+    // Remove the question mark infront
+    // ?c=32&r=shaowei%40open.gov.sg&h=82f3 -> c=32&r=shaowei%40open.gov.sg&h=82f3
+    const query = location.search.substr(1)
```
lshaowei18, line 70: Not sure if this is a good practice

frontend/src/components/unsubscribe/Unsubscribe.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const { c, r, h } = querystring.parse(query)
```
lshaowei18, line 71: Having to remove the `?` and checking if they are of type `string` seems weird to me.

jeantanzj:
```javascript
a = 'https://postman.gov.sg/v1/unsubscribe?c=aaa&r=bbb&h=ddd&h=eee'
g = querystring.parse(url.parse(a).query)
isValid = Object.values(g).every(x=>typeof x === 'string')
```

lshaowei18: Thanks that's alot more elegant than my implementation

frontend/src/components/unsubscribe/Unsubscribe.tsx
```diff
@@ -0,0 +1,142 @@
...
+    hash,
+    version,
+  }: {
+    campaignId: number
+    recipient: string
+    hash: string
+    version: string
+  }) {
+    try {
+      const isUnsub = await isUserUnsubscribed({
```
jeantanzj, line 52: why do we need to check this? the POST/PUT to unsubscribe should be idempotent

lshaowei18: It is mainly for UX, to check if the user is already unsubscribed. If he is unsubscribed, we won't show the confirmation button

lamkeewei:
Let me get @sarahsalim opinion on this. 

However, I can understand the intention of the original design. For any mailing list, the goal for the mailing list owner will be to retain the subscriber. Hence, emphasising on "I'd rather stay" helps nudge the user towards that desired action. But I guess in Postman's case, it is not so clear cut as we send both marketing type campaigns and transaction/notification type campaigns.

lshaowei18:
> Let me get @sarahsalim opinion on this.
> 
> However, I can understand the intention of the original design. For any mailing list, the goal for the mailing list owner will be to retain the subscriber. Hence, emphasising on "I'd rather stay" helps nudge the user towards that desired action. But I guess in Postman's case, it is not so clear cut as we send both marketing type campaigns and transaction/notification type campaigns.

Yup that's fair, understand where you are coming from. Happy for either design to go in.

frontend/src/components/unsubscribe/Unsubscribe.tsx
lamkeewei: @lshaowei18 we've discussed during sync-up just now and we will remove the GET call. 

Even if the users have an invalid URL (invalid hash, etc) or have already unsubscribed previously, we will always show the unsub page with the unsub button first. The validation will only happen on POST to `/unsubscribe`. I'll make the backend changes in a follow up PR.

frontend/src/components/unsubscribe/Unsubscribe.tsx
```diff
@@ -0,0 +1,111 @@
...
+    try {
+      const query = new URL(window.location.href).searchParams.toString()
+      const params = querystring.parse(query)
+
+      // Check to make sure that the search params are all strings and not []strings
+      const isValid = Object.values(params).every((v) => typeof v === 'string')
+      if (!isValid) {
+        throw new Error(
+          'Search params should all be strings and not array of strings.'
+        )
```
miazima, line 30: This error message would not be understandable by users, I think it would be sufficient to throw a general error something like `Invalid subscribe request` or `Error occured during unsubscription`

frontend/src/components/unsubscribe/Unsubscribe.tsx
```diff
@@ -0,0 +1,111 @@
...
+      await unsubscribeRequest({
+        campaignId: +campaignId,
+        recipient: recipient as string,
+        hash: hash as string,
+        version,
+      })
+      setUnsubscribed(true)
+    } catch (err) {
+      setErrorMsg(err.message)
+    }
```
miazima, line 44: When a request with invalid query params is rejected by the Joi validator for the `/unsubscribe` request, the error message returned is `"c" must be a number`. We can simplify the error displayed here like I mentioned above.

frontend/src/components/unsubscribe/Unsubscribe.tsx
```diff
@@ -0,0 +1,111 @@
...
+        </>
+      )
+    }
+  }
+
+  return (
+    <div className={styles.outer}>
+      <div className={styles.inner}>
+        <>
+          <img src={appLogo} />
```
lamkeewei, line 98: Minor issue. Add alt tags to this image and the one below. The rest of the warnings for missing alt tags were fixed in #586.

PR #595, opened 2020-07-22
feat: Telegram help command

### LINKED ISSUE #593: [Telegram] Help command
If users are confused while talking to the Telegram bot, there's no easy way for them to get help.

Introduce a `/help` command that provides simple explanations (tbc) and a way for them to contact us if necessary. 

Ideas for contacting us: 

- Postman Support account
- Form/email

Notes: 

- Register `/help` on the backend when saving bot credentials

dvrylc:
## Problem

From the issue: 
> If users are confused while talking to the Telegram bot, there's no easy way for them to get help.

Closes #593.

## Solution

Screen Shot 2020-07-22 at 5 17 56 PM

Add a new `/help` command that:
- Shows a short summary of the commands available (similar to BotFather)
- Provides a way for recipients to contact us via a support form

## Tests

Already tested on staging, to test locally: 

1. Make sure the new env vars are set, then run the handler locally
2. Send a `/help`, you should receive the message shown above
3. Storing your bot token via the frontend should also add a `/help` option to the suggestion list when you type `/`

## Deploy Notes

- `TELEGRAM_BOT_CONTACT_US_URL`: Support form URL
- `TELEGRAM_BOT_GUIDE_URL`: Recipient guide URL

Also remember to update all current bots with the new command.

### DIFF
--- a/serverless/telegram-handler/.env-example
+++ b/serverless/telegram-handler/.env-example
@@ -1,7 +1,7 @@
 # Required
 DB_URI=postgres://localhost:5432/postmangovsg_dev
 TELEGRAM_BOT_CONTACT_US_URL=https://go.gov.sg/postman-contact-us-recipient
-TELEGRAM_BOT_GUIDE_URL=https://guide.postman.gov.sg/faqs/faq-recipients
+TELEGRAM_BOT_GUIDE_URL=https://go.gov.sg/postman-recipient-guide
 # SENTRY_DSN=https://123.ingest.sentry.io/456
 
 # Optional, required in dev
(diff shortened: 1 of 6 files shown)


serverless/telegram-handler/.env-example
```diff
@@ -1,5 +1,7 @@
 # Required
 DB_URI=postgres://localhost:5432/postmangovsg_dev
+TELEGRAM_BOT_CONTACT_US_URL=https://go.gov.sg/postman-contact-us-recipient
+TELEGRAM_BOT_GUIDE_URL=https://guide.postman.gov.sg/faqs/faq-recipients
```
chinying, line 4: I know this is an example file, and I presume we'll be using this value as well when we deploy. Will it make sense for this to be a go link as well? In the event that we change our guide.

dvrylc: Ah good point, I initially went with this because it's consistent with frontend (Go links for forms, normal links for the rest) but the use case here is a bit different. Will make the changes.

PR #598, opened 2020-07-23
Unsubscribe email digest serverless

### LINKED ISSUE #580: [Unsubscribe] Cron job to send weekly unsubscribe digest to users
- Sweep `unsubscribe` table for `sent_at = NULL`, send users with at least one message a digest of unsub requests (recipient emails), grouped by campaign
	- Once complete, update `sent_at` 
- Use read replica when generating report

**Open questions**
- How should we execute the cron jobs? 
	- AWS Lambda: Potential for a timeout due to max 15 minutes execution time	
	- Elastic Beanstalk Worker: Each deploy will affect in progress jobs
	- ECS Scheduled Tasks: *To be evaluated*
	- AWS Batch: *To be evaluated*

miazima:
## Problem

A cron job is required to send weekly unsubscribe email digests to users.

Closes #580

**Features**:

- New serverless function `unsubscribe-digest`
- Gets list of unsubscribers grouped by campaigns and user
- Send emails to the users
- Updates `sent_at` in `unsubscribers` table for recipients in each campaign

## Todo

- Update guide link in email body
- Create cloudwatch event to trigger lambda weekly
- Create production lambda and travis environment variables
- Create new db role `serverless_unsubscribe`


lshaowei18:
I'm wondering how we can make this a better experience for the campaign admin. 

For example, if I as a campaign admin, receives the email with 100 unsubscribe requests, what do I have to do? 

I remember @pettypychen mentioning something about vlookup.

lamkeewei:
> I'm wondering how we can make this a better experience for the campaign admin.
> 
> For example, if I as a campaign admin, receives the email with 100 unsubscribe requests, what do I have to do?
> 
> I remember @pettypychen mentioning something about vlookup.

I think the current workflow for the agency user when they receive this email will be as follow:
1. Copy the list of unsub emails into a spreadsheet
2. Use VLOOKUP against this new column of unsub emails to clean up their current email spreadsheet

However, this workflow is with the assumption that the agency user diligently clean up the unsubs each time he/she receives the digest email. Otherwise, there will be an extra step where they have to first collate all the unsubs across the last few emails before doing the aforementioned steps. 

I was thinking a couple of things that we can do to improve the experience for the users:
1. Allow for the download of all unsubs for a campaign in the campaign detail page (potentially implemented as part of the customisable export issue). This will allow the user to avoid the need to collate and keep track of which digest email he/she has already processed. 
2. We can include the unsubscribers in CSV files (one for each campaign) instead of presenting it as a long list in the email. This might improve the readability of the digest email if there are many unsubs.

serverless/unsubscribe-digest/src/mail-client.class.ts
```diff
@@ -0,0 +1,69 @@
...
+        subject: input.subject,
+        replyTo: input.replyTo,
+        html: input.body,
+        headers: {},
+      }
+      if (input.referenceId !== undefined) {
+        // Signature expected by Sendgrid
+        // https://sendgrid.com/docs/for-developers/tracking-events/event/#unique-arguments
+        const headerValue = JSON.stringify({
+          unique_args: { message_id: input.referenceId },
```
jeantanzj, line 56: Note to add in environment unique_arg

serverless/unsubscribe-digest/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+const handler = async (): Promise<{ statusCode: number }> => {
+  try {
+    await init()
+
+    // Retreive unsubscribed recipients grouped by campaigns and users
+    const unsubscribeDigests = await getUnsubscribeList()
+
+    // generate email digest and send email to each user
+    for (const userDigest of unsubscribeDigests) {
+      await sendEmailAndUpdateUnsubscribers(userDigest)
```
jeantanzj, line 16: why is this in a for loop

miazima: `unsubscribeDigests` returns `{ email: [ { campaign, unsubscribers }, ...] }`, hence this iterating through to send the email digest and subsequently update the `unsubscribers` table for each user. I considered updating the `usubscribers` separately from this, but thought any error occurred in between could possibly cause the `unsubscribers` table to be not updated for emails already sent out.

serverless/unsubscribe-digest/src/unsubscribe.ts
```diff
@@ -0,0 +1,90 @@
...
+  const emailBody = createEmailBody(unsubscribe_list)
+
+  await mailClient.sendMail({
+    recipients: [email],
+    subject: 'Postman: Weekly Digest [Unsubscribe]',
+    body: emailBody,
+  })
+
+  for (const { id } of unsubscribe_list) {
+    await updateUnsubscribers(id)
```
lamkeewei, line 71: Can we do this in one query? Query can be something like: 
```sql
UPDATE unsubscribers SET sent_at = now() WHERE sent_at IS NULL AND campaign_id in (:campaign_ids);
```

.travis.yml
```diff
@@ -197,6 +200,48 @@ jobs:
...
+            - DB_URI=$PRODUCTION_UNSUBSCRIBE_DIGEST_DB_URI
+            - DB_READ_REPLICA_URI=$PRODUCTION_UNSUBSCRIBE_DIGEST_DB_READ_REPLICA_URI
+            - SES_FROM=$PRODUCTION_UNSUBSCRIBE_DIGEST_SES_FROM
+            - SES_HOST=$PRODUCTION_UNSUBSCRIBE_DIGEST_SES_HOST
+            - SES_PASS=$PRODUCTION_UNSUBSCRIBE_DIGEST_SES_PASS
+            - SES_USER=$PRODUCTION_UNSUBSCRIBE_DIGEST_SES_USER
+            - SES_PORT=$PRODUCTION_UNSUBSCRIBE_DIGEST_SES_PORT
+        - provider: lambda
+          edge: true
+          function_name: telegram-test
```
lamkeewei, line 226:
```suggestion
          function_name: unsubscribe-digest-staging
```

PR #600, opened 2020-07-23
fix: wording for non-protected email template

pregnantboy:
Remove `message A` from template instructions


frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -78,10 +78,7 @@ const EmailTemplate = ({
         onChange={setSubject}
       />
       <h4>{protect ? 'Message A' : 'Message'}</h4>
-      <p>
-        You can use the following keywords in Message A to personalise your
-        message.
-      </p>
+      <p>You can use the following keywords to personalise your message.</p>
```
jeantanzj, line 81: But in non protected email,  they can use any keyword.

PR #603, opened 2020-07-23
feat: remove new lines from tables, fix css

pregnantboy:
## Problem

New lines were being replaced with `<br>`'s and hoisted to the top of the table.

## Solution

Remove newlines in between table tags and fix css

Screenshot 2020-07-23 at 6 56 31 PM

### DIFF
--- a/frontend/src/services/validate-csv.service.ts
+++ b/frontend/src/services/validate-csv.service.ts
@@ -74,8 +74,17 @@ export async function validateCsv(
   })
 }
 
+function removeNewLinesFromTables(template: string) {
+  // Get all text within <table> tags
+  return template.replace(/<table\s*>(.*?)<\/table\s*>/gs, (match) =>
+    // Remove all new lines
+    match.replace(/(\r\n|\r|\n)/g, '')
+  )
+}
+
 export function hydrateTemplate(template: string, row: Record<string, any>) {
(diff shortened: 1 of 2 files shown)


frontend/src/services/validate-csv.service.ts
```diff
@@ -74,8 +74,17 @@ export async function validateCsv(
   })
 }
 
+function removeNewLinesFromTables(template: string) {
+  // Get all text within <table> tags
+  return template.replace(/<table\s*>(.*?)<\/table\s*>/gs, (match) =>
```
lshaowei18, line 79: I tried to insert a table like this in template B: 

<table style="width:100%">
  <tr>
    <th>Firstname</th>
    <th>Lastname</th>
    <th>Age</th>
  </tr>
  <tr>
    <td>Jill</td>
    <td>Smith</td>
    <td>50</td>
  </tr>
  <tr>
    <td>Eve</td>
    <td>Jackson</td>
    <td>94</td>
  </tr>
</table>

This is the result: 
Screenshot 2020-07-24 at 11 20 47 AM

I think when the table has attributes, eg: `table style="width:100%"` the regex doesn't match and the white spaces won't be trimmed

pregnantboy: ure right

pregnantboy:
Merged with #609

PR #604, opened 2020-07-24
refactor: Remove GET unsubscriber status endpoint

lamkeewei:
## Problem

GET endpoint wasn't required. Should be removed to minimise API surface. 

## Solution

**Improvements**:
- Removed `GET /unsubscribe` endpoint
- Changed `POST /unsubscribe` to `PUT /unsubscribe/:campaignId/:recipient`
   - Used `PUT` instead as `POST` endpoints are typically not idempotent
   - Moved `campaignId` and `recipient` to URL params to be in line with [RFC2616](https://tools.ietf.org/html/rfc2616#section-9.6) where the _"PUT method requests that the enclosed entity be stored under the supplied Request-URI"_
   - Also conforming with [RFC2616](https://tools.ietf.org/html/rfc2616#section-9.6), if a new unsubscriber is added, the  endpoint should return `201 Created`. Otherwise a `200 OK` is returned if the unsubscriber already exists. 
   - `h` and `v` are still passed in the request body JSON

## Tests
**Successfully add new subscriber**
- Create a campaign and note down the campaignId
- Call `PUT /unsubscribe/:campaignId/:recipient` with the following payload
```json
{
   "h":"hash",
   "v":"v1"
}
```
- Endpoint returns unsubscriber object in the request body with a status of `201 Created`

**Test idempotence**
- Call `PUT /unsubscribe/:campaignId/:recipient` with the following payload. The `campaignId` and and `recipient` should already been previously added before. 
```json
{
   "h":"hash",
   "v":"v1"
}
```
- Endpoint returns unsubscriber object in the request body with a status of `200 OK`

**Test invalid unsubscribe request**
- Create a campaign and note down the campaignId
- Call `PUT /unsubscribe/:campaignId/:recipient` with an invalid hash 
- Endpoint returns an error message with a `400 Bad Request` status
- Repeat this with invalid values for `campaignId` and `v`

### DIFF
--- a/backend/src/core/middlewares/unsubscriber.middleware.ts
+++ b/backend/src/core/middlewares/unsubscriber.middleware.ts
@@ -64,9 +64,9 @@ const findOrCreateUnsubscriber = async (
   const statusCode = created ? 201 : 200
   const { createdAt } = unsubscriber
   return res.status(statusCode).json({
-    campaignId,
+    campaign_id: campaignId,
     recipient,
-    createdAt,
+    created_at: createdAt,
   })
 }
 
--- a/backend/src/core/middlewares/unsubscriber.middleware.ts
+++ b/backend/src/core/middlewares/unsubscriber.middleware.ts
@@ -59,10 +59,7 @@ const findOrCreateUnsubscriber = async (
   })
 
   const statusCode = created ? 201 : 200
-  return res.status(statusCode).json({
-    campaign_id: campaignId,
-    recipient,
-  })
+  return res.sendStatus(statusCode)
 }
 
 export const UnsubscriberMiddleware = {
--- a/backend/src/core/middlewares/unsubscriber.middleware.ts
+++ b/backend/src/core/middlewares/unsubscriber.middleware.ts
@@ -1,5 +1,4 @@
 import { Request, Response, NextFunction } from 'express'
-import { UniqueConstraintError } from 'sequelize'
 
 import { CampaignService, UnsubscriberService } from '@core/services'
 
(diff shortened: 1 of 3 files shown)


backend/src/core/middlewares/unsubscriber.middleware.ts
```diff
@@ -38,67 +42,35 @@ const isUnsubscribeRequestValid = async (
...
+  })
 
-    next(err)
-  }
+  const statusCode = created ? 201 : 200
+  const { createdAt } = unsubscriber
+  return res.status(statusCode).json({
+    campaign_id: campaignId,
+    recipient,
+    created_at: createdAt,
```
lshaowei18, line 69: Hmmm how is created_at going to be used on the frontend?

lamkeewei: Don't think it will be used in the frontend. Let me drop that from the response. Thanks for pointing this out!

backend/src/core/middlewares/unsubscriber.middleware.ts
```diff
@@ -38,67 +42,30 @@ const isUnsubscribeRequestValid = async (
...
+  const { campaignId, recipient } = res.locals.unsubscriber
+  const { 1: created } = await UnsubscriberService.findOrCreateUnsubscriber({
+    campaignId,
+    recipient,
+  })
 
-    next(err)
-  }
+  const statusCode = created ? 201 : 200
+  return res.status(statusCode).json({
```
lshaowei18, line 62: Actually according to Sarah's design we don't need campaignId or recipient at all, right? Sorry should have mentioned it along with `created_at`

lamkeewei: No worries, you're right that it's redundant. I've remove the response body before merging. Thanks!

Issue #501, opened 2020-07-08
Investigate slow sms rates


jeantanzj:
**Describe the bug**
A campaign that was supposed to take 140 seconds took 15 minutes. Campaigns to investigate:  3330, 3329

**Additional context**
Not sure if it's because of the ` await Promise.all(messages.map((m) => sendMessage(m)))` in message-worker, which would block on network calls, or because of database contention

jeantanzj:
Update: I think we were bottlenecking on CPU utilization. Just increased the CPU available. Will observe

jeantanzj:
After increasing CPU available, the time taken went down on average (time taken / expected time decreased from ~6 to ~1-2). But there are some campaigns that seem to take a long time. See campaigns 3848 and 3782

jeantanzj:
I wanted to replace `await Promise.all(messages.map((m) => sendMessage(m)))` with Bluebird's concurrency option, eg.
```
manyMessages = // get more messages than the current rate
sendMessageV2 = async (message) => {
const start = Date.now()
await sendMessage(message)
await waitForMs(Date.now() -start)
}
Promise.map(manyMessages.map(sendMessageV2), {concurrency: rate})
```
but you have to determine how many messages to get at once. We risk OOM, and if the worker dies midsend, we may have to resend more duplicates.

pettypychen:
Can I close this one? I thought we fixed it.

PR #608, opened 2020-07-24
Speed up worker

jeantanzj:
## Problem
Addresses #501
The following query computes time taken for email messages to be sent. 
```sql
-- Compute ratio of time taken to expected time. The closer it is to 1, the better.
select *,  time_taken/expected_time as ratio from (
        -- Expected time taken (based on number of messages / send rate)
	select a.campaign_id, num_messages, extract(epoch from time_taken) as time_taken, (num_messages::float/job_queue.send_rate) as expected_time, job_queue.updated_at from (
		-- Actual Time taken for num_messages to be sent
		select campaign_id, count(*) as num_messages, max(delivered_at)-min(sent_at) as time_taken  from email_messages, campaigns where email_messages.campaign_id = campaigns.id  and sent_at is not null group by campaign_id
	) a, 
job_queue where job_queue.campaign_id = a.campaign_id and num_messages>10
) b order by updated_at desc;
```
It shows that most of the email campaigns take longer to send than necessary. 
## Implementation 2

With Ian's advice, use plimit to limit the number of messages being sent concurrently. This resolves the risks that are listed in the following description of implementation 1.
Campaign 338 was sent using the current implementation which awaits on the network calls to send messages. 339 and 340 were sent using this plimit implementation. 
Screenshot 2020-08-25 at 6 45 49 AM

## Previous implementation 1
Do not await on sending messages. Instead, just fire off the message at the specified rate. 
Risks:
- Too many unresolved connections (since we do not await, we're not sure if a connection is closed). I tested keeping some connections open, reaching around 1500 connections before ECONNRESET is thrown. 
- log_next_job will finalize a campaign when 1) all the messages that have sent_at set, also have delivered_at set, or, 2) all the messages have sent_at set, and twenty seconds have passed (to account for orphan campaigns where the sending worker died). With this fire-and-forget modification, in scenario (2), some messages may not have delivered_at set before getting logged. This situation is mitigated by the callback status being set on the messages table instead of the ground truth table. 

On staging - campaigns 245 and 246 were sent using this implementation. The ratio is much closer to 1. 
Screenshot 2020-07-25 at 1 29 06 AM

### DIFF
--- a/worker/package-lock.json
+++ b/worker/package-lock.json
@@ -2830,10 +2830,9 @@
       }
     },
     "p-limit": {
-      "version": "2.2.2",
-      "resolved": "https://registry.npmjs.org/p-limit/-/p-limit-2.2.2.tgz",
-      "integrity": "sha512-WGR+xHecKTr7EbUEhyLSh5Dube9JtdiG78ufaeLxTgpudf/20KqyMioIUZJAezlTIi6evxuoUs9YXc11cU+yzQ==",
-      "dev": true,
+      "version": "3.0.2",
+      "resolved": "https://registry.npmjs.org/p-limit/-/p-limit-3.0.2.tgz",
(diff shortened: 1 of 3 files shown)


miazima:
Tested sending messages on staging, lgtm

miazima (approved):

PR #614, opened 2020-07-27
Add warning message on sms template exceeding character limit threshold

### LINKED ISSUE #528: Warn if SMS message seems likely to exceed 1600 character limit
A user is getting the error_code "The concatenated message body exceeds the 1600 character limit." for sms. 
We could set a character count.

Setting a character count is actually hard because each message is of variable length due to templating. 
We could check the character count of the preview message. If it is 1500 characters (? magic number ? ), then warn the user "Hey, your message seems pretty long. If any of your messages exceeds 1600 characters, they won't be sent. Consider shortening your template". 

Or we could set a character limit on the template?

miazima:
## Problem

We want to warn users if their SMS messages seem likely to exceed 1600 character limit.

Closes #528 

**Features**:

- Display a warning message if a campaign template exceeds 1000 characters (threshold defined by us)

## Before & After Screenshots

image

### DIFF
--- a/frontend/src/components/dashboard/create/Create.module.scss
+++ b/frontend/src/components/dashboard/create/Create.module.scss
@@ -113,3 +113,7 @@
   color: $dark-blue;
   text-decoration: underline;
 }
+
+.characterCount {
+  margin-top: -1rem;
+}
--- a/frontend/src/components/dashboard/create/Create.module.scss
+++ b/frontend/src/components/dashboard/create/Create.module.scss
@@ -115,7 +115,8 @@
 }
 
 .characterCount {
-  margin-top: -1.5rem;
+  margin-top: -1rem;
   text-align: right;
-  font-size: 0.8rem;
+  font-size: 0.85rem;
+  color: $light-grey;
 }
--- a/frontend/src/components/dashboard/create/Create.module.scss
+++ b/frontend/src/components/dashboard/create/Create.module.scss
@@ -115,5 +115,7 @@
 }
 
 .characterCount {
-  margin-top: -1rem;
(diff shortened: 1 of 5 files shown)


jeantanzj:
Would this make more sense on the template page with a displayed character count? I feel like I wouldnt know what triggered the warning

miazima:
> Would this make more sense on the template page with a displayed character count? I feel like I wouldnt know what triggered the warning

@jeantanzj  Yeah seems fair in a user point of view, I considered it but thought we achieve the same result by checking after the template was finalised rather than parsing the template (to determine the char count without variables) on template page in a debounced manner. Is this being too stringent, or actually we could display this warning without parsing the template to remove variables as it probably makes a difference of ~100 characters.

jeantanzj:
@sarahsalim Thoughts?

chinying:
I hate the whole "Javascript .length()" is broken meme, but do we have to account for it here? If people start throwing non english characters that may potentially mess up the character count.

miazima:
Discussed this with @sarahsalim and @pettypychen, the consensus was 
1. to have a character count and warn when user exceeds character threshold in template page
2. no significant benefit in extracting out the template without variables for the character count, as this is a warning and not a hard requirement

jeantanzj:
> I hate the whole "Javascript .length()" is broken meme, but do we have to account for it here? If people start throwing non english characters that may potentially mess up the character count.

You are right, it will mess up the count. But this is pretty complicated due to the encoding and may not be worth the effort. 
https://www.twilio.com/docs/glossary/what-is-gsm-7-character-encoding
https://www.twilio.com/blog/adventures-unicode-sms

PR #618, opened 2020-07-28
chore: Restrict Telegram bot token access

### LINKED ISSUE #594: [Telegram] Restrict access to bot token through environment tag on secrets
Bot tokens should be used exclusively in either production or staging and not both. This is to prevent unintended switching of the webhook URL for the bot.

- Tag bot token secrets on SecretsManager with either `production` or `staging` when the secret is first created
- Set IAM policies to restrict reading of tokens tagged with either `production` or `staging`
- Subsequent `putSecrets` should fail if it is executed in a different environment from the one it was first created in
- We only want to apply this policy to bot tokens secrets and not Twilio creds since there's no issue with using them in both `production` and `staging` environment at the same time

**Some helpful links**
- [Limiting Access to Secrets with Specific Staging Labels or Tags](https://docs.aws.amazon.com/secretsmanager/latest/userguide/auth-and-access_identity-based-policies.html#permissions_grant-limited-actions)
- [...IfExists Condition Operators](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html#Conditions_IfExists) can be used to only apply the policy when it has the specific environment tag 
- [IAM Policy simulator](https://policysim.aws.amazon.com/) to test/debug policies

dvrylc:
## Problem

Storing a bot token repeatedly in staging/production causes the webhook URL to keep changing - difficult to debug. 

Closes #594. 

## Solution

This PR:

- Tags new bot token secrets with a `environment` tag set to `staging` or `production`
- Relies on IAM policies to restrict access 
- Displays generic error message to users

## Tests

Deploy this to staging to test: 

1. Create a new Telegram bot, store on staging
2. Check that `environment` is set to `staging` in the stored secret on Secrets Manager
3. Change the `environment` value to something else
4. Delete the credential in Postman and try storing again
    - You should get "Error storing credentials" on the frontend
    - In CloudWatch, there should be an entry like: 
    ```
    error - User: arn:<beanstalk instance role> is not authorized to perform: secretsmanager:PutSecretValue
    ```

## Deploy Notes

- Manually tag all existing bot tokens in Secrets Manager with their respective environments
- Update Secrets Manager policy for production EB instance - refer to staging's as an example
    - Add `secretsmanager:TagResource` action
    - Add condition to only allow access to production secrets
    ```
    "Condition": {
        "StringEqualsIfExists": {
            "secretsmanager:ResourceTag/environment": "production"
        }
    }
    ```

### DIFF
--- a/backend/src/core/services/credential.service.ts
+++ b/backend/src/core/services/credential.service.ts
@@ -16,13 +16,13 @@ const secretsManager = new AWS.SecretsManager(configureEndpoint(config))
  * Upserts credential into AWS SecretsManager
  * @param name
  * @param secret
- * @param isBotToken
+ * @param restrictEnvironment
  * @throws error if update fails
  */
 const upsertCredential = async (
   name: string,
   secret: string,
-  isBotToken: boolean
+  restrictEnvironment: boolean
 ): Promise<void> => {
   // If credential doesn't exist, upload credential to secret manager, unless in development
   if (!config.get('IS_PROD') && !config.get('aws.awsEndpoint')) {
@@ -50,7 +50,7 @@ const upsertCredential = async (
         .createSecret({
           Name: name,
           SecretString: secret,
-          ...(isBotToken
+          ...(restrictEnvironment
             ? { Tags: [{ Key: 'environment', Value: config.get('env') }] }
             : {}),
         })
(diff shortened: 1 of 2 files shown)


backend/src/core/services/credential.service.ts
```diff
@@ -62,12 +68,17 @@ const upsertCredential = async (
  * Save a credential into secrets manager and the credential table
  * @param name
  * @param secret
+ * @param isBotToken
  */
-const storeCredential = async (name: string, secret: string): Promise<void> => {
+const storeCredential = async (
+  name: string,
+  secret: string,
+  isBotToken = false
```
lamkeewei, line 76: Minor nitpick. Was thinking if we can rename this argument to something that is more generic and expresses the intent better. Something along the lines of `restrictEnvironment` (?). Otherwise someone without familiarity with this PR might be wondering why bot tokens are treated differently.

dvrylc: Good idea, changed it!

PR #625, opened 2020-07-29
Poll export status to prevent being stuck in loading status

### LINKED ISSUE #623: Export Error Spinner keeps spinning unless you you refresh the page
**Describe the bug**
On the campaign stats page, the `export error spinner` will keep spinning unless you refresh the page again. 

**To Reproduce**
1. Send a normal campaign with an invalid recipient
2. Keep the campaign stats page open for report generation export button
3. No export button. The spinner will keep spinning. 

**Expected behavior**
The report export button should show up after the spinner is done spinning to let the user know that the report is ready for export.

miazima:
## Problem

Export button is stuck in loading status after campaign has been sent out, until page is refreshed.

Closes #623 

## Solution

After campaign status became `Status.Sent`, `ExportRecipients` component stopped checking for the export status. To ensure that the export status is updated after campaign has been sent, poll the export status until it returns a finalised status.

### DIFF
--- a/backend/src/core/interfaces/campaign.interface.ts
+++ b/backend/src/core/interfaces/campaign.interface.ts
@@ -50,8 +50,8 @@ export interface CampaignDetails {
 
 export interface CampaignStats extends CampaignStatsCount {
   status: string
-  updated_at: Date
   halted?: boolean
+  status_updated_at?: Date
 }
 
 export interface CampaignStatsCount {
@@ -59,7 +59,7 @@ export interface CampaignStatsCount {
   unsent: number
   sent: number
   invalid: number
-  updatedAt?: Date
+  updated_at: Date
 }
 
 export interface CampaignInvalidRecipient {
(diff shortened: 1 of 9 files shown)


jeantanzj:
Isnt the spinner on the dashboard also going to spin forever too? but im not sure if polling would be the right way to go about doing this~

miazima:
For the campaign list, since we do not poll the campaign status, there is no way we can know the updated export status. If a campaign export status is `Loading` when received on the campaign list page, I considered manually transitioning to `Ready` state after 1 min, but realised we cannot determine between `Ready` state and `No Error` state without updating campaign list data.

miazima:
One possible solution is, if any campaigns in the list have `Loading` export status, refresh the campaign data every min to get the updated campaign status.

miazima:
Realised what I mentioned before was wrong, the campaigns list returns `has_failed_recipients` for sent campaigns that can be used to determine if the campaign is of `Ready` or `No Error` state. Refactoring the export component and fixing the issue.

jeantanzj:
This doesn't fix the spinner issue when refresh stats is clicked. One preliminary way I can imagine doing it is to prevent the export button from changing state after it has been enabled once. So it overrides the condition for waiting for 1 min to elapse since the updatedAt timestamp of stats. 

Haven't thought about it thoroughly yet.

PR #627, opened 2020-07-29
Save credentials from campaign sending page

### LINKED ISSUE #219: Feature: Save credentials from the campaign sending view
**Baseline**: Credentials can only be saved at the settings/account page. 

**Ideal flow**: Save credentials inside of the campaign sending view.

miazima:
## Problem

When users inputs new credentials within the campaign sending steps, the credentials are not saved to the user.

Closes #219

**Features**:

- Add credential label input component in `SMSCredentials` and `TelegramCredentials` page
- Extend the `campaign/{campaignId}/{channel}/new-credentials` endpoint for telegram and sms to accept `label` in the request body
- Add `checkUserCredentialLabel` and `storeUserCredential` middleware to the endpoint
- Modify `storeUserCredential` middleware to advance to the next middleware if `campaignId` exists, to associate the campaign with the stored credential


backend/src/sms/routes/sms-campaign.routes.ts
```diff
@@ -35,6 +36,11 @@ const uploadCompleteValidator = {
 
 const storeCredentialsValidator = {
   [Segments.BODY]: Joi.object({
+    label: Joi.string()
+      .min(1)
+      .max(50)
+      .pattern(/^[a-z0-9-]+$/)
+      .required(),
```
lamkeewei, line 43: Slight concern here to make this required. This would be a breaking change for any application that is integrating with Postman using the API.

lamkeewei:
Currently, repeated label name error is only raised when the user selects "Validate credentials". Not sure if we want to standardise the UX with the add new credentials modal where the error message gets surfaced immediately when the user enters the label name (reference image below). Quicker feedback could be less frustrating for the user. 

image

backend/src/sms/routes/sms-campaign.routes.ts
miazima: True, I'll assess the backward compatibility again

miazima:
> Currently, repeated label name error is only raised when the user selects "Validate credentials". Not sure if we want to standardise the UX with the add new credentials modal where the error message gets surfaced immediately when the user enters the label name (reference image below). Quicker feedback could be less frustrating for the user.

I think the check and error block for existing label name should be included in within the `CredLabelInput` component. We could pass in the existing credential labels as a prop. Will make a separate issue for this.

frontend/src/services/sms.service.ts
```diff
@@ -30,29 +30,34 @@ export async function saveTemplate(
...
+  label: string
   recipient: string
   accountSid: string
   apiKey: string
   apiSecret: string
   messagingServiceSid: string
 }): Promise<void> {
   try {
-    await axios.post(`/campaign/${campaignId}/sms/new-credentials`, {
+    await storeCredentials({
```
lamkeewei, line 51: @miazima can you explain again why we've chose to split up storing and validation of the credentials into separate API calls? I think you've explained this before but it has slipped my mind (sorry!). 

My concern is that we are storing credentials into SecretsManager without first validating them. In the current flow, if I pass in an invalid credential, it will be stored into SecretsManager even though an error is thrown eventually. Over time we could have a bunch of invalid creds in SecretsManager that will be hard to clean up.

miazima: I see your concern, my first approach was to modify the `/new-credentials` endpoint to require label as one of the parameters - this would have been a breaking change for API users and making it an optional param would mean that `checkUserCredentialLabel` and `storeUserCredential` middleware would have to handle undefined `label` values which defeats the purpose of those middleware.

Hence we chose to split up the storing and validation calls. I understand setting `validate` at the settings route may not be a good idea. If we move this optional validate params to `/campaign/{campaignId}/sms/credentials` I believe there would not be any unintended side effects. `validate === false` feels hackish, but I couldn't think of an alternative solution to this.

lamkeewei: @miazima I've tried to briefly refactor to see if we can conditionally apply the middlewares for saving user credentials. It's on the `feature/save-credentials-alt` branch. You can check out the diffs at [here](https://github.com/opengovsg/postmangovsg/compare/feature/save-credentials...feature/save-credentials-alt). Not sure if it's cleaner though.

lamkeewei:
One other minor issue. If I reselect "Enter new credentials" right after inserting a new credential, the "Credential Label" field is not cleared.

frontend/src/services/sms.service.ts
miazima: I think it is a more straightforward way of implementing this conditional instead of modifying the middlewares, esp since `validateAndStoreCredentials` is shared by multiple endpoints, which would be confusing. Including comments explaining the need for the new middleware would make this clearer too. Feel free to rebase your changes on this branch.
In `storeUserCredential`, we could return `next()` if `campaignId` params exists, else return `return res.json({ message: 'OK' })`

lamkeewei:
@miazima I've merged in the changes and updated docs and `storeUserCredential`. Let me know if it's ok. Thanks!

lamkeewei (approved):

PR #632, opened 2020-07-29
chore: Remove IS_PROD

### LINKED ISSUE #521: Get rid of config.get('IS_PROD')
For our team's developers, our IAM permissions enable us to retrieve secrets that have been tagged as 'staging'. So we can call secrets manager directly.

For open source contributors, they can use localstack to create and retrieve secrets.

There is no longer a need for a prod check. For sequelize-loader,  we can simply use the node_env to decide whether to use ssl.

dvrylc:
## Problem

We don't need `IS_PROD` anymore, development vs. staging/production can be determined through `env`.

Closes #521.

## Solution
**Features** 
- Replace `IS_PROD` checks with appropriate `env` checks.
- Use localstack Secrets Manager for local development

**Bug Fix**
- Fix protected email CSV upload to localstack S3
- Add `wait-on` for backend to be ready before starting worker to prevent situation where database tables are not ready yet

**Improvements**
- Mount init scripts within localstack container to remove the need to manually run scripts to initialise S3 bucket
- Add alternative for Docker based services (postgres and redis) 

## Tests
**Localstack Secrets Manager**
- Start localstack by running `npm run dev:localstack`
- Configure environment variable to use `AWS_ENDPOINT` (refer to `.env-example`) and start application
- Create a set of new user credentials (either Twilio/Telegram). 
- Verify that the credentials are now stored in localstack by running `aws --endpoint-url=http://localhost:4566 secretsmanager list-secrets`.
   - _Make sure the configure AWS region and the localstack region matches if you are receiving an empty response_ 
- Create a new campaign with the newly added credentials and ensure that both backend and worker are using the right credentials

## TODO 

- [x] Change Secrets Manager access policy for Postman team's credentials to use `development` tag instead of `staging` to prevent mix up with staging bot tokens that are also tagged with `staging` (implemented in #618).


package.json
```diff
@@ -13,7 +13,9 @@
     "dev": "concurrently -k -p \"[{name}]\" -n \"Backend,Worker,Frontend\" -c \"yellow.bold,cyan.bold,white.bold\" \"npm run dev:backend\" \"npm run dev:worker\" \"npm run dev:frontend\"",
     "dev:frontend": "cd frontend && npm run dev",
     "dev:backend": "cd backend && npm run dev",
-    "dev:worker": "cd worker && npm run dev",
+    "dev:worker": "npm run wait:backend && cd worker && npm run dev",
+    "dev:localstack": "docker-compose -f docker-compose.yml up localstack",
+    "dev:services": "docker-compose -f docker-compose.yml up",
```
jeantanzj, line 18: when is this used?

backend/src/core/config.ts
```diff
@@ -287,44 +284,12 @@ const config = convict({
     default: '65',
     env: 'DEFAULT_COUNTRY_CODE',
   },
-  smsOptions: {
```
jeantanzj, line 290: Great!

package.json
jeantanzj: Oh I see. When you dont want to spin up redis and pg using docker. Document it?

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      - "4566-4599:4566-4599"
+    environment:
+      - DEBUG=1
+      - DATA_DIR=/tmp/localstack/data
+      - FILE_STORAGE_BUCKET_NAME=${FILE_STORAGE_BUCKET_NAME:-localstack-upload}
+      - AWS_LOG_GROUP_NAME=${AWS_LOG_GROUP_NAME:-postmangovsg-beanstalk-localstack}
+    volumes:
+      - ./localstack:/docker-entrypoint-initaws.d
+      - localstack-data:/tmp/localstack
+      - /var/run/docker.sock:/var/run/docker.sock
```
jeantanzj, line 29: Why do you need to mount the docker socket? This is quite risky as the containerized service now can make docker commands as if they were the docker host.

jeantanzj: @lamkeewei

lamkeewei: This is required if we would like to use the lambda functionality in localstack since it uses the host Docker daemon to execute lambda code in a container. 

Will remove it. Accidentally, left it there previously while experimenting with the localstack lambda stuff.

jeantanzj: Will it help us to debug lambdas?

lamkeewei: Personally, I feel that it's pretty tedious. For reference the steps required to run a lambda within localstack is as follow: 
1. Create lambda functions within localstack on startup (can be added into init scripts):
```
aws --endpoint-url http://localhost:4566 lambda create-function \
    --function-name unsusbcribe-digest \
    --code S3Bucket="__local__",S3Key="/local/path/to/serverless/unsubscribe-digest/build" \
    --handler index.handler \
    --runtime nodejs12.x \
    --role serverless
```
3. Make sure that Typescript code is compiled locally
2. To test invoke lambda function manually using AWS cli
```
aws --endpoint-url http://localhost:4566 \
    --function-name unsubscribe-digest \
    --payload @file://payload.json \
    response.json
```
3. Inspect `response.json` for the output

I feel that the only benefit for using localstack is the ability to simulate triggering of lambdas based on another action on AWS. For e.g., triggering a lambda when a file is uploaded on an S3 bucket. 

But for our case, there's currently no support to automatically trigger a lambda to simulate a bounce/complaint events when an email is sent. I think in the pro version there's a built in SMTP server that can trigger bounce/complaint events when a email is routed through it, but cannot confirm this. 

I was also looking at potentially running a local version of our API Gateway setup through localstack. But again, it gets pretty tedious to deploy onto localstack if you're not using something like [SAM](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/what-is-sam.html). You'll either have to write a super convoluted bash script to call the AWS cli commands to setup the resources/methods ([example](https://gist.github.com/crypticmind/c75db15fd774fe8f53282c3ccbe3d7ad)) or roll your own Cloudformation template to set it up. Both of which feels like overkill and would require developers to have knowledge of how to set this whole thing up.

jeantanzj: Agreed.  SAM requires you to write a convoluted template, and it's not easy to use. We would rather have a server with endpoints that our server can hit. 

> you can't just refer to the database running locally through localhost
Don't you refer to the database by using the service name defined in the docker-compose.yml file?

lamkeewei: >  Don't you refer to the database by using the service name defined in the docker-compose.yml file?

Yeah. But that would put a hard requirement to run all your services through Docker.

backend/src/core/services/credential.service.ts
```diff
@@ -26,31 +26,29 @@ const upsertCredential = async (
 ): Promise<void> => {
   // If credential doesn't exist, upload credential to secret manager
   try {
-    logger.info(`Updating credential in AWS secrets manager for name=${name}`)
-    // Secrets stored in localstack will not be tagged as no error is thrown by localstack if the secret
-    // does not exists.
+    logger.info('Storing credential in AWS secrets manager')
     await secretsManager
-      .putSecretValue({
```
lamkeewei, line 38: Since our developer IAM policy explicitly deny all SecretsManager actions unless they're tagged with "development", `putSecretValue` will always result in an IAM explicit deny error instead of an `ResourceNotFoundException`. This is because the `putSecretValue` call [does not allow us to pass in tags](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_PutSecretValue.html#API_PutSecretValue_RequestParameters). 

To resolve this, there's two possible solution I thought of:
1. Allow `DescribeSecret` regardless of environment tag in our developer IAM policy. We will then call this function before `putSecretValue`.  If the secret does not exists, `ResourceNotFoundException` will be thrown. 
2. Switch around `putSecretValue` and `createSecret`. Since `createSecret` allows us to pass in tags, the call will not be blocked by our IAM policy. 

Decided to go with option 2. While `DescribeSecret` does not return encrypted values, I think it would still be good to restrict access.

PR #640, opened 2020-08-17
feat: Add detect-secrets precommit hook

### LINKED ISSUE #393: Add precommit check to prevent secrets from being committed
Intern accidentally committed credentials to forms and go. We could add a precommit check. 

Consider https://github.com/awslabs/git-secrets

lamkeewei:
## Problem

Add pre-commit check for secrets using `detect-secrets`. Chose this over `git-secrets` largely because there's already built-in support for detection of both Twilio and AWS credentials. 

The `.secrets.baseline` file is generated by running `npm run detect-secrets:scan`. I've chosen to disable the keyword scanner plugin as it raises many false positives due to the names of our variables (E.g. `apiSecrets`, etc). There's currently a single false positive in `frontend/.env-example` for the example Sentry DSN. 

When a secret is flagged in future commits, two options are available if it is a false positive:
1. Add an [inline pragma](https://github.com/Yelp/detect-secrets#inline-allowlisting) to explicitly allow the secret
2. Update secrets baseline by running `npm run detect-secrets:update`

Closes #393 

## Limitations
Detection of high entropy string is only done on quoted strings. For instance, the following line will not be detected as a a secret as the value is not quoted. 
```
TOP_SECRET_STRING=highentropysecrethere
```

`detect-secrets` does this to avoid high number of false positives originating from possibly valid strings (variable names, etc). Reference to [regex](https://github.com/Yelp/detect-secrets/blob/e65371f1fbbb6df4c67655ac72e74161da09e9bf/detect_secrets/plugins/high_entropy_strings.py#L38) used and [issue](https://github.com/Yelp/detect-secrets/issues/203) in `detect-secrets` repo discussing this. 

Regex based checks (Twilio and AWS) does not have this limitation.

## Solution
- Add pre-commit hook to check for secrets in staged files using `detect-secrets`

## Tests
- Add a secret in one of the files (E.g. Add AWS credentials in `.env-example` file)
- Staged file using `git add`
- Run `npm run detect-secrets` to run `detect-secrets` hook
- Error should raised

## Deploy Notes

**New dev dependencies**:

- `detect-secrets` : Tool to detect secrets in files to be committed. Can be installed by running `pip install detect-secrets`.


miazima (approved):
Tested using secrets and high entropy strings, lgtm.

PR #643, opened 2020-08-18
feat: swap out squirrelly for mustache

chinying:
## Problem

Squirrelly is in violation of our original CSP policy (`unsafe-eval`) as it uses `new Function`. We originally did not require this because our templating was done on the backend, but our password-protected mail feature is carried out client-side (frontend), which resulted in CSP violation.

## Solution

Switch it out for a templating library that is not in violation of our CSP, in this case I've gone with Mustache. 
In terms of implementation details, I've stuck to as much of the old code as possible, including the function signatures as I'm wary of breaking stuff. It can slowly be refactored away if we find that this is stable. 

The main tradeoff I can think of is that mustache is not feature rich, and likely deliberately so. I initially picked Squirrelly it allows for the possibility of having filters and partials inside the templates, if we ever think of going with these options. 

## Tests

`npm run test` in the postman-templating module. All tests should pass. 

## Deploy Notes

**New dependencies**:
- `mustache` : templating library


modules/postman-templating/src/template-client.ts
```diff
@@ -102,47 +67,43 @@ export class TemplateClient {
...
+          if (!params) continue
 
-          // if params provided == attempt to carry out templating
           if (dict[key]) {
             const templated = dict[key]
             tokens.push(templated)
-            return
+          } else {
+            // recipient key must have param
+            if (key === 'recipient') {
```
jeantanzj, line 80: The previous code pattern was cleaner.  Return early, dont nest ifs if possible. 
https://blog.codinghorror.com/flattening-arrow-code/

modules/postman-templating/src/template-client.ts
```diff
@@ -102,47 +67,43 @@ export class TemplateClient {
...
     } catch (err) {
       console.error({ message: `${err.stack}` })
-      if (err.message.includes('unclosed tag'))
+      if (err.message.includes('Unclosed tag'))
         throw new TemplateError(
           'Check that all the keywords have double curly brackets around them.\nA correct example is {{ keyword }}, and incorrect ones are {{ keyword } or {{ keyword . '
         )
-      if (err.message.includes('unclosed string'))
+      // reserved chars in mustache are '^' and '#' and '/'
+      if (
```
jeantanzj, line 100: Try `{{=<%name %>=}}`. This throws "Invalid tags"

jeantanzj: `{{&name}}` should throw an error - it doesn't

jeantanzj: {{{name}}} should... throw an error? probably. it doesnt

jeantanzj: `{{#name}}Never shown!{{/name}}` should also throw an error - it doesn't.

jeantanzj: `{{! name }}`, `{{> name}}` all pass when they shouldn't. Let's test the cases on https://github.com/janl/mustache.js

modules/postman-templating/src/template-client.ts
chinying: Ok yeah I can fix this, I just personally thought having an explicit `else` was clearer.

jeantanzj:
Eh doesnt work?? 
These still pass
{{&name}}
{{{name}}}
{{#name}}Never shown!{{/name}}

jeantanzj (approved):

jeantanzj:
To figure out why it works on localhost but not on amplify before merging. Test on mustache.postman.gov.sg

PR #662, opened 2020-08-24
refactor: split serverless builds on travis

jeantanzj:
## Problem

Travis builds for the serverless functions were taking pretty long as the pre-deploy build process was being repeated for every function.

## Solution

Split the builds into separate jobs


lamkeewei (approved):
lgtm! Tested and works.

PR #673, opened 2020-08-28
chore: force sms for numbers starting with +6580 to be delivered

### LINKED ISSUE #672: Singapore numbers starting with 802 fail Twilio validation
**Describe the bug**
SMS messages to Singaporean numbers starting with `802` will result in an error from Twilio's API. I suspect Twilio uses `libphonenumber` for validation on their server side. So while we bypassed our own validation in #636, Twilio's validation still throws a false positive. 

**To Reproduce**
1. Use test Twilio credentials to avoid sending out actual SMS messages
2. Send a validation message to a number starting with `802`
3. Following error will be thrown `Error: The 'To' number +65802XXXXX is not a valid phone number.`

**Possible solution**
Use `forceDelivery` for Singaporean numbers. [Example](https://github.com/opengovsg/FormSG/blob/525c1bb8e042e74adcca429df0f544d97a6c821a/src/app/services/sms.service.ts#L176) from FormSG.

jeantanzj:
## Problem

Closes #672 

## Solution

This parameter is undocumented in Twilio, but recommended by Twilio Support to get past the validation issue

Screenshot 2020-08-28 at 2 33 21 PM

### DIFF
--- a/backend/src/sms/services/twilio-client.class.ts
+++ b/backend/src/sms/services/twilio-client.class.ts
@@ -11,12 +11,17 @@ export default class TwilioClient {
     this.messagingServiceSid = messagingServiceSid
   }
 
-  public send(recipient: string, message: string): Promise<string | void> {
+  public send(
+    recipient: string,
+    message: string,
+    forceDelivery = false
+  ): Promise<string | void> {
     return this.client.messages
       .create({
         to: recipient,
         body: this.replaceNewLines(message),
         from: this.messagingServiceSid,
+        forceDelivery: forceDelivery,
       })
       .then((result: { [key: string]: string }) => {
         const { status, sid, error_code: errorCode, code } = result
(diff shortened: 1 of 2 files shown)


backend/src/sms/services/twilio-client.class.ts
```diff
@@ -11,17 +11,26 @@ export default class TwilioClient {
...
+  ): Promise<string | void> {
     return this.client.messages
       .create({
         to: recipient,
         body: this.replaceNewLines(message),
         from: this.messagingServiceSid,
+        forceDelivery: forceDelivery,
       })
       .then((result: { [key: string]: string }) => {
         const { status, sid, error_code: errorCode, code } = result
```
lamkeewei, line 27: Tested this through the settings page and the force delivery doesn't seem to be triggered.

Looks like when the validation fails on Twilio's side, the promise from `this.client.messages.create(...)` doesn't resolve and is [rejected](https://github.com/twilio/twilio-node/blob/abf3470d89c68f5efbb6eeb811242404ec59762d/lib/rest/api/v2010/account/message.js#L135) instead.

PR #674, opened 2020-08-28
fix: wrong column name in twilio callback

jeantanzj:
## Problem
User reported discrepancy between twilio logs and Postman's logs. Turns out we weren't updating the error_code column correctly.

## Solution

Update the correct column

## Tests
- Locally: Post an event to your lambda running locally

Before
Screenshot

After
Screenshot

- End to end : I've built it on the staging lambda, and sent a message with the id '300643'. Twilio still reports the message as 'Sent', not 'Failed' (it's expected to fail because that number doesn't exist). When it returns the failed callback, the message status should change from 'SUCCESS' to 'ERROR'.


lshaowei18 (approved):
lgtm

Issue #658, opened 2020-08-24
Improve user experience for queued campaigns


miazima:
**Is your feature request related to a problem? Please describe.**
Campaigns which have been queued, refer to campaigns which are waiting to be sent out while other campaigns are being sent out. During this wait time, the user sees a stagnant progress bar and statistics on the frontend. This leads to confusion and poor user experience for users as they do not understand why their campaign is 'stuck'. This issue is most notable for email campaigns, as in our current implementation, only one email campaign can be in progress at a time and other concurrent campaigns would be queued. 

**Describe the solution you'd like**
We can pre-empt users about the status of their campaign by displaying their position in the queue, or by providing an estimate wait time for their campaign to begin sending.

Refer [here](https://docs.google.com/document/d/1WdGOvLR_wagiuVMX4k-6r-dTzhfqzknJ_q86Xfu6Uss/edit#heading=h.85ppgox58m7b) for relevant discussion on email send rate and possible solutions on user education.

@sarahsalim It would be good to have some insights on what form of information would be useful to users.

PR #677, opened 2020-08-31
Compute queue position and wait time for queued campaigns

miazima:
## Problem

It would be informative to users who have queued campaigns to know what is the queued status of their campaign and how long they it would take for their campaign to start sending out.

Backend changes to return metrics relevant to #658 

## Solution

- Extend the existing `/stats` endpoint to also return campaign queue status if campaign's job is in `READY` status
- Get the position of the job in the queue
- Compute an approximate time taken for preceding jobs to complete

### DIFF
--- a/backend/src/core/services/stats.service.ts
+++ b/backend/src/core/services/stats.service.ts
@@ -120,7 +120,7 @@ const getCurrentStats = async (
     ],
   })
 
-  if (job == null)
+  if (job === null)
     throw new Error('Unable to find campaign in job queue table.')
 
   // Get stats from statistics table
@@ -146,8 +146,8 @@ const getCurrentStats = async (
       halted: job.campaign.halted,
     }
   }
-  // else, return archived stats
-  return {
+
+  const stats = {
     error: archivedStats.error,
     unsent: archivedStats.unsent,
     sent: archivedStats.sent,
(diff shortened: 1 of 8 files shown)


backend/src/core/services/stats.service.ts
```diff
@@ -154,6 +154,73 @@ const getCurrentStats = async (
...
+    if (job.campaign.statistic?.unsent && job.sendRate) {
+      const timeCurrJob = job.campaign.statistic.unsent / job.sendRate
+      totalTime += timeCurrJob
+    }
+  })
+
+  return {
+    ...stats,
+    queueStatus: {
+      position,
```
jeantanzj, line 220: Position can be found based on the updated_at of the existing jobs which use the same credential. But actually, knowing the position isn't important right? The user just needs to know the wait time.

Maybe something like: 
Get the updated_at and cred_name for this campaign
```sql
SELECT j.updated_at, c.cred_name FROM job_queue  j , campaigns c WHERE 
	j.campaign_id = :campaign_id 
	AND c.id = j.campaign_id
```

Calculate the wait time.
```sql
SELECT SUM(s.unsent / j.send_rate) AS wait_time FROM job_queue j, campaigns c, statistics s WHERE 
c.cred_name = :cred_name
AND s.campaign_id = c.id
AND j.campaign_id = c.id 
AND j.status = 'READY' 
AND j.updated_at < :updated_at;
```

jeantanzj (changes requested):

backend/src/core/services/stats.service.ts
```diff
@@ -158,6 +158,45 @@ const getCurrentStats = async (
...
     status_updated_at: job.updatedAt, // Timestamp when job was logged
   }
+
+  // If job is not in send queue, return archived stats
+  if (job.status !== JobStatus.Ready) {
+    return stats
+  }
+
+  const campaign = await Campaign.findByPk(campaignId)
+  if (!campaign || !campaign.credName) {
```
jeantanzj, line 168: The existence of the campaign is checked in isEmailCampaignOwnedByUser, isSmsCampaignOwnedByUser and isTelegramCampaignOwnedByUser

jeantanzj: And it must have had a non-null credName for a job to be created

jeantanzj (approved):

PR #686, opened 2020-09-01
refactor(callback): telegram

jeantanzj:
To test, set up the webhook on your telegram bot.

```bash
npm run dev # spins up server on localhost:4000
ngrok http 4000 # creates a tunnel

export BOT_TOKEN=123:ABC # your own bot token
export WEBHOOK=https://abc.ngrok.io # your local tunnel

curl -X POST --header "Content-Type: application/json" "https://api.telegram.org/bot$BOT_TOKEN/setWebhook" -d '{"url":"'"$WEBHOOK/v1/callback/telegram/$BOT_TOKEN"'"}'
```

- [x] /start will prompt you to share your phone number
- [x] "share your phone number" will store the phone number and telegram id
- [x]  /help shows help message
- [x] /updatenumber will prompt you to share your phone number


lamkeewei (approved):
Tested and works as expected. 

One minor point to note is that there's a slight change in behaviour as compared to the lambda implementation. Previously, `200 OK` is returned even if there's an error. This implementation now returns a non-`2XX` response when an error occurs. This will trigger Telegram's bot API to automatically retry the request. From my observation, Telegram API seems to keep retrying (with exponential backoff) until it succeeds [for 24 hours](https://core.telegram.org/bots/api#getting-updates). 

You can try this out by setting the webhook to one with an invalid bot token and then interact with the bot (`/updatenumber`, `/help`, etc).

I don't see any obvious issue with this, plus this has the added benefit of ensuring delivery of Telegram messages in the event of an app crash on our side.

Issue #688, opened 2020-09-01
Replace hard coded enums with enum values


miazima:
**Describe the issue**
Noticed a few places in the code where we used the hardcoded enum strings instead of the enum values from `constants.tx`. This is susceptible to accidental typo errors and would not be caught by the compiler. 
E.g. https://github.com/opengovsg/postmangovsg/blob/dd2b9c62c67bc0ca89f2c7308ea3cbfa2cbd4c9a/backend/src/core/services/stats.service.ts#L128

It would be better practice to import the relevant enums and use them instead.

PR #690, opened 2020-09-01
Replace hardcoded strings with enum values in stats.service.ts

miazima:
## Problem

Part of #688 

## Solution

Replace hardcoded strings with enum values in stats.service


lamkeewei (approved):
lgtm!

PR #692, opened 2020-09-02
fix: Disable campaign pagination arrow when there is only a single page

### LINKED ISSUE #687: Page indicator arrows not disabled for 1 page of campaigns
**Describe the bug**
When the campaigns page has 1 page of campaigns, the right navigation arrow is not disabled.

image

**To Reproduce**

1. Test with <5 campaigns
2. Go to 'Campaigns'
3. See page indicator 

**Expected behavior**
When there is only 1 page of campaigns, both the left and right page indicator arrows should be disabled.

lamkeewei:
## Problem

Closes #687 

## Solution

**Bug Fixes**:

- `pageCount` wasn't rounded up to an integer

**Improvements**
- Removed unused `useState` and `useEffect` imports that were raising a warning

## Before & After Screenshots
image

## Tests
- Have <= 10 campaigns 
- Check that both arrows for pagination component are disabled


miazima (approved):
Tested and resolves issue, lgtm!

PR #693, opened 2020-09-02
Speed up worker part 2

jeantanzj:
Fixing https://github.com/opengovsg/postmangovsg/pull/608

## Problem 

`SENT` state was prematurely set when messages were being retrieved faster than the network calls were resolving. 
This caused the logger to log messages early, resulting in a condition where the messages were actually sent, but not updated to the ops table (because the messages had already been transferred back from ops table to the messages table)

## Solution

Only set the state to `SENT` when all the promises have resolved / the failsafe is breached.

## Other considerations
The entire purpose of the initial design was to prevent the app from having to hold all the messages in memory (ie, make one getMessages call for all the messages, and send them out at the specified rate). This prevents OOM, and since it only dequeues messages as needed,, we would lose fewer messages (whose result would not be written to disk)  in the event of a crash. 

This new design consequentially loses those advantages as it is perfectly possible for all the messages to be dequeued before the promises are resolved.

### DIFF
--- a/worker/package-lock.json
+++ b/worker/package-lock.json
@@ -2830,10 +2830,9 @@
       }
     },
     "p-limit": {
-      "version": "2.2.2",
-      "resolved": "https://registry.npmjs.org/p-limit/-/p-limit-2.2.2.tgz",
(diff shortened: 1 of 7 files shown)


miazima (approved):
Tested with 100 recipients, all messages got logged successfully.

jeantanzj:
I have no idea what's wrong. Works on localhost for email and smses, works on staging for email, but gets logged way too early for smses.

PR #710, opened 2020-09-14
feat: Encrypted database backup lambda function

### LINKED ISSUE #703: Regular encrypted backup of RDS database to S3
As part of our implementation for disaster recovery, we need to regularly `pg_dump` our database to S3. These dumps will then be shipped to an alternate site outside of AWS for storage.

**Requirements**
- Create a lambda function that will run `pg_dump` on the read replica and store the output to a dedicated S3 bucket
- The output from `pg_dump` should be encrypted using symmetric encryption
- Lambda function will be periodically triggered by the [RDS backup completed event](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_Events.html#USER_Events.Messages) (`RDS-EVENT-0022`)
- Upon successful backup, notification should be sent to Slack

lamkeewei:
## Problem

Implements a lambda function that create a backup of a Postgresql database by running `pg_dump` and encrypting its output with a symmetric encryption algorithm. 

Closes #703 

## Solution

**Features**:
- Created a lambda layer that installs and package up `pg_dump` and it's dependencies so that it can be used within the lambda runtime
- Implement lambda to run, encrypt and upload `pg_dump` output to a S3 bucket

## Tests
- Run `database-backup-staging` lambda through the console using `RDSEvent`
- Download the backup files from `postmangovsg-rds-backup-staging` S3 bucket. The folder name follows the following format `YYYY-MM-DD`.
- Build and decrypt the files using `scripts/decrypt.ts`. You can choose to either use the staging backup private key in our 1Password or GCP KMS. (See below for detailed instructions) 
- Test out the dump by running `pg_restore` and inspecting the content of the decrypted `secrets.dump`

**Using local private key for decryption**
Download staging private key PEM from 1Password (Staging - Backup Encryption Key). Download the dump files from S3 and run the following to decrypt the pgdump (`postman.dump`). 

```bash
$ node build/scripts/decrypt.js \
    --privateKey=key.pem \
    --input postman.dump \
    --output postman.decrypted.dump \
    --params postman.json
```
Repeat again for `secrets.dump`. 

**Using GCP KMS for decryption**
Ensure you have [installed and setup](https://cloud.google.com/sdk/docs/install) the gcloud CLI (`gcloud init`). 

Download the GCP service account credentials from 1Password (Staging - GCP KMS Decrypter Credentials). Download the dump files from S3 and run the following to decrypt the pgdump (`postman.dump`). 

```bash
$ export GOOGLE_APPLICATION_CREDENTIALS=[path to the JSON key file]
$ export KEY_ID=[copy key ID from 1Password]
$ node build/scripts/decrypt.js \
    --kmsKey=$KEY_ID \
    --input postman.dump \
    --output postman.decrypted.dump \
    --params postman.json
```
Repeat again for `secrets.dump`. 

## Deploy Notes

- `PG_VERSION` : The current version of Postgresql that we're running. Used to determine the version of `pg_dump` that we will be packaging with the lambda.
- Staging
   - `STAGING_BACKUP_ROLE` : Backup lambda IAM role ARN
   - `STAGING_BACKUP_DB_URI` : Database URI to use for `pg_dump`. Use read replica.
   - `STAGING_BACKUP_BUCKET_NAME` : S3 bucket name to store backup in
   - `STAGING_BACKUP_PUBLIC_KEY` : Public key used to encrypt encryption key
- Production
   - `PRODUCTION_BACKUP_ROLE` : Backup lambda IAM role ARN
   - `PRODUCTION_BACKUP_DB_URI` : Database URI to use for `pg_dump`. Use read replica.
   - `PRODUCTION_BACKUP_BUCKET_NAME` : S3 bucket name to store backup in
   - `PRODUCTION_BACKUP_PUBLIC_KEY` : Public key used to encrypt encryption key

**Infrastructure updates**
- [ ] Standardise backup window for all database to between 3:00AM - 3:30AM SGT
- [ ] Set CloudWatch alarm for backup lambda failure
- [x] Generate and [import new key in GCP KMS](https://cloud.google.com/kms/docs/importing-a-key) for production backup

## TODOs
- [x] Create IAM role for backup lambda with necessary permissions (RDS, S3)
- [x] Update Travis to build and package lambda with `pg_dump`
- [x] ~~Implement the use of custom KMS CMK for generation of data key for symmetric encryption~~
   - _Ciphertext generated by KMS is [not portable](https://docs.aws.amazon.com/kms/latest/developerguide/importing-keys.html#importing-keys-considerations). Even if you have the same key material, decryption is not possible outside of KMS_ 
- [x] Asymmetric encrypt encryption key for each dump
- [x] Add dumping for SecretsManager
- [x] Setup SNS queue and connect up RDS events to trigger lambda


serverless/database-backup/src/backup.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.secretsDump = secretsDump
+  }
+
+  private getBackupFolderName(): string {
+    const today = new Date()
+    const yyyy = today.getFullYear()
+    const dd = today.getDate().toString().padStart(2, '0')
+    const mm = (today.getMonth() + 1).toString().padStart(2, '0')
+
+    return `${yyyy}-${mm}-${dd}`
```
jeantanzj, line 36: does the timezone matter...? will this always be in utc / match with the date of the rds snapshot? what's wrong with `(new Date()).toISOString().substring(0,10)`

serverless/database-backup/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "database-backup",
+  "version": "1.0.0",
+  "description": "Backup database to S3",
+  "main": "src/index.js",
+  "scripts": {
+    "build:docker": "docker run --rm  -e PG_VERSION=$PG_VERSION -v $PWD:/var/task lambci/lambda:build-nodejs12.x ./build.sh",
```
jeantanzj, line 7: why did you choose to run a build script directly, instead of building a dockerfile from the lambci/lambda:build-nodejs12.x image?  
where is PG_VERSION set?

lamkeewei: `PG_VERSION` is set in Travis. Not too sure if that's the best place to track this actually. The [docs](https://www.postgresql.org/docs/11/app-pgdump.html#PG-DUMP-NOTES) mentioned that we should use the same major version as the server you're dumping from. I've added this instead of hardcoding some value to allow us to update without code change when we upgrade to a new major version of Postgres.

I was thinking that if I use a Dockerfile, I'll have to do the following to get the built lambda zip out since you can't do volume mounts for `docker build`. 

```sh
$ docker build -t database-backup -f Dockerfile .
$ docker run --rm -v $PWD:/output database-backup cp /var/task/code.zip /output/code.zip
```

Doesn't seem to be a way to copy files out of a built image without running the container. 

Chose to directly run the script instead since it can be done in one step. Downside is that it might pollute your mounted host directory or a bug in the script might accidentally rm some stuff. But given that this script is only run in the CI server, my thoughts are that it's ok. Open to changing.

serverless/database-backup/src/backup.ts
lamkeewei: Time in lambda runtime will be UTC. The `TZ` env var in the runtime is [fixed to UTC](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime). Backup window for RDS is also defined in UTC.

Will update the generation of the folder name to the one-liner! Thanks! 👍

serverless/database-backup/build.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+PG_MAJOR_VERSION=$(echo $PG_VERSION | cut -d'.' -f1)
```
jeantanzj, line 3: how about
```
amazon-linux-extras enable postgresql11

yum clean metadata && yum install postgresql -y
```

serverless/database-backup/build.sh
```diff
@@ -0,0 +1,29 @@
...
+# Create layer directory structure
+rm -rf $LAMBDA_TASK_ROOT/bin $LAMBDA_TASK_ROOT/lib
+mkdir -p $LAMBDA_TASK_ROOT/{bin,lib}
+
+# Copy pg_dump binary
+cp $(which pg_dump) $LAMBDA_TASK_ROOT/bin
+
+# Copy the libraries that are required by pg_dump but are not available in the standard lambda runtime. To determine this list, manually run this script from within a container.
+# After which, run ldd $(which pg_dump) to determine the list of shared libraries the binary is linked to. Compare this list with what is available in the lambda runtime. The
+# list of libraries in the standard lambda runtime can be obtained by running ls /lib64 with a test lambda.
```
jeantanzj, line 19: how to run ls /lib64 in a lambda?

jeantanzj: ok figured it out
```
const cp = require('child_process')
exports.handler = async (event) => {
    const pgdump = cp.spawn('ls', ['/lib64'])
    pgdump.stdout.pipe(process.stdout)
    return 
};
```

serverless/database-backup/build.sh
lamkeewei: Was using this approach originally. But then came across [this issue](https://unix.stackexchange.com/questions/77083/pg-dump-version-mismatch-on-debian), which made me feel a little unsure on whether `pg_dump` is indeed compatible within a major version. Perhaps it's fixed after postgresql9, have not looked deeper into this. 

Other concern was also that postgresql12 is not available through this method. Amazon release cycle seems to be quite slow too. [This issue](https://stackoverflow.com/a/57973741) in 2019 was still asking for postgresql11 support. If there's some sort of security vulnerability that require us to bump the major version, this method might break.

jeantanzj (approved):
It works well. Given that you had to share state, the classes makes sense, but in general, prefer functional style over OOP, and avoid holding state if possible

PR #711, opened 2020-09-14
feat: Fetching secrets in Docker, from AWS Secrets Manager

lshaowei18:
## Problem

We want to centralize our AWS related secrets in AWS Secrets Manager (SM). 

Needed a way to fetch the secrets from SM in Elastic Beanstalk.

## Solution

Initially we wanted to fetch the secrets at the app level (#666), but we decided that it is added too much complexity. 

Instead, we will fetch the secrets & export them as environment variables in Docker, after the container has started. 

### Dockerfile
We replaced the `npm start` cmd with a bash script `docker-entrypoint.sh`.

In order to fetch the secrets from SM and export them as environment variables, we have to install `aws-cli` & `jq` on the image. 

**Features**:

- Installed aws-cli & jq on the docker image
- Added a travis before-deploy script to replace the secret id according to the travis branch
- Added a script to be used as a entry point
- Script fetches the secrets from SM & export them as environment variables

## TODO
- [ ] Add staging secrets to SM
- [ ] Add production secrets to SM
- [ ] Test on staging. Save one of the env vars from staging in 1password, remove it from EB env vars and then see if it works. Alternatively, we can create a separate env to test this.

Add travis env vars: 
- [ ] STAGING_SECRET_ID
- [ ] PRODUCTION_SECRET_ID

## Deploy Notes

**New Travis environments variables**:

- `STAGING_SECRET_ID` 
- `PRODUCTION_SECRET_ID`

**New scripts**:

- `before-deploy.sh` : For Travis to run before deploying EB
- `docker-entrypoint.sh` :  Entry point for docker image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ jobs:
         - npm run lint-no-fix
         - npm run test
       before_deploy:
-        - cp -R ../modules ./
+        - ./before-deploy.sh
       deploy:
         - provider: elasticbeanstalk
           edge: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,14 +12,14 @@ jobs:
   include:
     - name: backend
       before_install:
-        - cd backend
+        - cd $TRAVIS_BUILD_DIR/backend
       install:
         - npm ci
       script:
         - npm run lint-no-fix
         - npm run test
       before_deploy:
-        - ./before-deploy.sh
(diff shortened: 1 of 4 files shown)


jeantanzj (approved):
Works with no env vars set. Should set up a trigger to restart app servers when secret is updated.

PR #728, opened 2020-09-21
Custom domain backend & worker

lshaowei18:
## Solution
For convenience of testing, you may check out https://github.com/opengovsg/postmangovsg/tree/custom-domain-test

### Endpoints

`GET /settings/email/from`
- Returns an array of valid custom 'from' email addresses for the user. Should always return default email that is set in the backend. If the user's email is verified as a custom email address, it will be in the array.

`POST /settings/email/from`
- Accepts emails in the form of `email@email.com`, and `DISPLAY_NAME <email@email.com>`
- Verifies the user's email by:
  - Checking with AWS
  - Dig and check the DNS record of the email domain
- If it is verified, sends an email to the user to inform him/her.
- Stores the email in `email_from_address`.

Modified `PUT email/template`
- Does the verification with AWS & dig the DNS record.
- Add the `from` address to `template`.

### Mail Clients
Refactored `worker` & `backend` `MailClient` to set the `from` address to follow the `template's`. 

### Database
- New table `email_from_address` to store all the email addresses that is verified by AWS SES & the DKIM records added to the email domain's DNS. 
- email_from_address contains two columns, `email` and `name`, where name is a display name for that email address. Eg. `DISPLAY_NAME <email@email.com>`
- Add from column to EmailTemplate

### Misc
- Added user's email to `req.session.user` for OTP logins. 

## Deployment
Take note that we need to backfill all existing templates' from address with the current default `Postman.gov.sg <donotreply@mail.postman.gov.sg>`.

- [ ] add new table & column to staging 
- [ ] backfill `from` address in `template` table for staging

```sql
alter table email_templates add column "from" text;
select count(*) from email_templates where "from" is null;
update email_templates set "from" = 'Postman.gov.sg <donotreply@mail.postman.gov.sg>' where "from" is null;
```

## Tests
You can use my email account - saved in our password vault under `Custom Domain test email`
- [x] GET /settings/email/from should return an array of usable custom from addresses, including the default email
- [x] POST /settings/email/from/verify should accept emails in the form of `email@email.com`, and `DISPLAY_NAME <email@email.com>`

**Verification of From address**
- [x] If the default email is supplied, verification on AWS and verification of DKIM records should be skipped 
- [x] If the default email is supplied, a validation message should still be sent out
- [x] A validation message should be sent from the custom from address
- [x] If the email is not verified on AWS, it cannot be saved  in email_from_address
- [x] If the email's DKIM records are not set, it cannot be saved  in email_from_address
- [x] If the email is verified and its DKIM records are correct, it should be saved into email_from_address
- [x] If a from address is supplied with a display name, the display name should be saved into the name column, and the email address into the email column.
- [x] If the same from address is supplied with a different display name, the display name should be upserted into the name column. 
- [x] If the same from address is supplied with no display name, a null should be upserted into the name column. 
- [x] A user can only verify the email address that they logged in with as a from address. 

**Use of From address**
- [x] The custom from address and default email should be selectable when creating email template
- [x] A user can use the default email to create the email template 
- [x] A user can use the email they logged in with as a from address
- [x] A user cannot use any other valid email addresses in email_from_address to send a message, if they are not logged in with that particular email (limited access).
- [x] The test message should be sent from the custom from address
- [x] The email messages from that campaign should be sent from the custom from address

**Previously valid From Address which became invalid**
- [x] If the custom from address exists in email_from_address, and is selected, but has since become invalid (either it was removed from SES, or its DKIM records were modified), the template should not be saveable. 
- [x] If the template was saved with a custom from address which has since become invalid, the `*test message* step should fail.
- [x] If the template was saved with a custom from address, which has since become invalid,  and the `*test message* has already been sent, the *send campaign* step should result in errors for all the email messages in that campaign.


backend/src/email/services/email.service.ts
```diff
@@ -272,6 +283,93 @@ const uploadProtectedCompleteOnChunk = ({
...
+const verifyCnames = async (
+  tokens: Array<string>,
+  email: string
+): Promise<void> => {
+  // get email domain
+  const domain = email.slice(email.lastIndexOf('@') + 1)
+  const cnames = tokens.map((token) => `${token}._domainkey.${domain}`)
+  try {
+    for (const cname of cnames) {
+      await dns.promises.resolve(cname, 'CNAME')
```
jeantanzj, line 300: you need to check that the value is as expected

backend/src/email/services/email.service.ts
```diff
@@ -272,6 +283,93 @@ const uploadProtectedCompleteOnChunk = ({
...
+ * 1. Against the user's email address
+ * 2. With AWS to ensure that we can use the email address to send
+ * 3. Checks the domain's dns to ensure that the cnames are there
+ * @throws Error if any of the verification fails
+ */
+const verifyFromEmailAddress = async (
+  email: string,
+  userId: number
+): Promise<void> => {
+  await verifyEmail(email, userId)
```
jeantanzj, line 352: we shouldn't let them pass the email parameter? we could get it from req.session.user

lshaowei18:
Testing the different cases: 

- [ ] AWS verification fails & dig fails
- [ ] AWS verification pass & dig fails
- [ ] AWS verification fails & dig pass

backend/src/email/middlewares/email-template.middleware.ts
```diff
@@ -35,7 +36,10 @@ const storeTemplate = async (
 ): Promise<Response | void> => {
   try {
     const { campaignId } = req.params
-    const { subject, body, reply_to: replyTo } = req.body
+    const { subject, body, reply_to: replyTo, from } = req.body
+
+    const fromAddress = from === 'default' ? config.get('mailFrom') : from
```
lshaowei18, line 41: @jeantanzj Unsure about my implementation of `from` as `default`. Let me know what you think!

backend/src/email/middlewares/email.middleware.ts
```diff
@@ -92,9 +92,89 @@ const previewFirstMessage = async (
...
+  next: NextFunction
+): Promise<Response | void> => {
+  const email = req.session?.user?.email
+  try {
+    const isVerified = await CustomDomainService.isEmailVerified(email)
+    if (isVerified) return res.status(200).json({ email })
+  } catch (err) {
+    return next(err)
+  }
+  return res.status(200).json({ email: '' })
```
lshaowei18, line 144: Was considering between this or empty object.

backend/src/email/middlewares/email.middleware.ts
```diff
@@ -92,9 +92,89 @@ const previewFirstMessage = async (
...
+  res: Response,
+  next: NextFunction
+): Promise<Response | void> => {
+  const { from } = req.body
+  if (from === 'default') return next()
+
+  const email = req.session?.user?.email
+
+  // from should be the same as email if it is not default
+  if (from !== email)
```
lshaowei18, line 161: I could just use the user's email instead of doing this check.

backend/src/email/routes/email-campaign.routes.ts
```diff
@@ -27,6 +27,12 @@ const storeTemplateValidator = {
       .lowercase()
       .allow(null)
       .required(),
+    from: Joi.string()
```
jeantanzj, line 30: Remember I wanted to make this a default / optional because we need to maintain backwards compatibility?

backend/src/email/services/custom-domain.service.ts
```diff
@@ -0,0 +1,113 @@
+import dns from 'dns'
+import AWS from 'aws-sdk'
+
+import { EmailFromAddress } from '@email/models'
+
+const ses = new AWS.SES({ region: 'eu-central-1' })
```
jeantanzj, line 6: region should not be hardcoded - it can be parsed from SES_HOST

lshaowei18: Thanks for catch this; forgot about it

backend/src/email/routes/email-campaign.routes.ts
lshaowei18: Oh ya ok now I remember will make the changes

backend/src/email/services/custom-domain.service.ts
```diff
@@ -0,0 +1,115 @@
+import dns from 'dns'
+import AWS from 'aws-sdk'
+import config from '@core/config'
+
+import { EmailFromAddress } from '@email/models'
+
+const [, region] = config.get('mailOptions.host').split('.')
```
lshaowei18, line 7: Was considering between this or `config.get('mailOptions.host').split('.')[1]`

backend/src/email/routes/email-settings.routes.ts
```diff
@@ -1,5 +1,82 @@
...
+ *      tags:
+ *        - Settings
+ *      requestBody:
+ *        required: true
+ *        content:
+ *          application/json:
+ *            schema:
+ *              type: object
+ *              properties:
+ *                from:
```
jeantanzj, line 36: Update this

backend/src/email/middlewares/email.middleware.ts
```diff
@@ -79,22 +81,157 @@ const previewFirstMessage = async (
...
+ * Stores the verified email address that we can use to send out emails.
+ */
+const storeFromAddress = async (
+  req: Request,
+  res: Response,
+  next: NextFunction
+): Promise<Response | void> => {
+  const { from } = req.body
+  const defaultEmail = config.get('mailFrom')
+  if (from === defaultEmail) return res.sendStatus(200)
```
jeantanzj, line 175: Is there a better way of handling the check against defaultEmail?  Worried about users adding donotreply@mail.postman.gov.sg

backend/src/email/services/custom-domain.service.ts
```diff
@@ -0,0 +1,148 @@
...
+import AWS from 'aws-sdk'
+import escapeHTML from 'escape-html'
+import config from '@core/config'
+
+import { EmailFromAddress } from '@email/models'
+import { MailService } from '@core/services'
+import logger from '@core/logger'
+import { formatFromAddress } from '@core/utils/from-address'
+
+const [, region] = config.get('mailOptions.host').split('.')
```
lamkeewei, line 11: Thought about this while testing the frontend. We might have an issue deriving the region this way because we're using different SES credentials on the backend and the workers. The domain might be verified on the worker's SES region but not on the backend's SES region.

backend/src/core/middlewares/auth.middleware.ts
```diff
@@ -52,6 +52,7 @@ const verifyOtp = async (
         id: user.id,
         createdAt: user.createdAt,
         updatedAt: user.updatedAt,
+        email: user.email,
```
lamkeewei, line 55: Will need to include `email` also in the user object returned by from `AuthService.getUserForApiKey`. Otherwise, calling `GET /settings/email/from` will throw a 500 because `email` is undefined.

backend/src/email/middlewares/email.middleware.ts
```diff
@@ -79,22 +81,160 @@ const previewFirstMessage = async (
...
+/**
+ * Checks that the from address is either the user's email or the default app email
+ */
+const isFromAddressAccepted = async (
+  req: Request,
+  res: Response,
+  next: NextFunction
+): Promise<Response | void> => {
+  const { from } = req.body
+  const userEmail = req.session?.user?.email
```
lamkeewei, line 107: This could be `undefined` if a user had logged in previously and the session has not expired. `500 Internal Server Error` will be thrown for the From Addresses routes.

lamkeewei (approved):
lgtm. All test cases works as expected.

PR #735, opened 2020-09-23
Verify backup database dump in Google Cloud

### LINKED ISSUE #705: Verification of database backups
We should verify the backups in order to prevent silent failures. This can be fulfilled either manually (through documentation) or programatically (write a script to restore and verify against an actual postgresql instance). 

**Requirements**
- Download and decrypt the dump from GCS bucket
- Run `pg_restore` on the dump and check that it completes successfully
- Run functional checks against the restored database (API calls, simple queries, etc)

miazima:
## Problem

The backup db dumps in google cloud storage need to be verified in order to prevent silent failures.

Closes #705

## Solution

- Get gcloud backup account service key from Secret Manager 
- Downloads the latest backup from backup GCS bucket using key
- Use KMS private key to decrypt the dump from GCS bucket
- Run pg_restore on the dump and check that it completes successfully
- Run functional checks against the restored database (API calls, simple queries, etc)

**Refer to [Setup GCP services for verifying backups #758](https://github.com/opengovsg/postmangovsg/issues/758) for more details on how to run the scripts and setup services on GCP.**

## Testing
- a service account key is required to test locally
- build and run the docker image


serverless/verify-backup/Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN npm ci
 COPY src ./src
 COPY scripts ./scripts
 COPY tsconfig.json ./
+COPY postmangovsg-service-account-key.json ./
```
lamkeewei, line 14: Would avoid doing this as it would bake our credentials permanently into the Docker image and potentially expose it. Another person can just run another command using the built Docker image utilising the included credentials. Might be better if we pass it in from the environment variables instead.

miazima: Forgot to remove this, the service account key is only needed for local testing. In cloud run, the credentials are auto-detected from the environment. I have removed this now.

jeantanzj (commented):
I've verified the retrieval of files from gcp buckets, decryption of dumps, and restoration into db, on my machine. Have not tested the pubsub. Could you walk me through the set up on GCP and how to test it? 

- Appreciate some documentation on env vars to set (at least describe what's in secret manager - I know it's the service account json) / save a copy of the .env in 1Password
- Rename the service account (the spelling error is irritating)
- Where is notify-slack being used?

serverless/verify-backup/scripts/verify-backup.sh
```diff
@@ -0,0 +1,18 @@
...
+su - postgres -c "psql -lqt | cut -d \| -f 1 | grep -qw postmangovsg_dev && dropdb postmangovsg_dev"
+su - postgres -c "createdb postmangovsg_dev"
+
+# Restore dump; On error, print error message and exit script
+echo 'Restoring data dump'
+pg_restore --no-owner -h localhost -p 5432 -U postgres -d postmangovsg_dev postman.decrypted.dump || \
+{ echo 'Data dump restoration failed!' && exit 1; }
+
+echo 'Making simple query on db'
+su - postgres -c "psql -d postmangovsg_dev -c 'SELECT 1+1'"
```
jeantanzj, line 17: - if the verification query fails, script should exit with an error code. 
- can we notify slack on success? eg with number of users, campaigns, messages etc, for sanity check. Can be done in separate PR.

serverless/verify-backup/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  console.log(
+    `gs://${bucket}/${srcFilename} downloaded.`
+  );
+}
+
+async function getLatestBackup(files: Array<File>) {
+  // get last three files in list which is in lexicographic order
+  // db dump, secrets dump and params json
+  files.reverse()
+  for (let i = 0; i < 3; i++) {
```
jeantanzj, line 55: - minor nitpick:  why not use`files.slice(-3).forEach...`
- Can we be sure that the files are always sorted by date? What guarantees it?

serverless/verify-backup/src/server.ts
```diff
@@ -0,0 +1,16 @@
+import express, {Request, Response} from 'express'
+import bodyParser from 'body-parser'
+
+const app = express();
+const PORT = process.env.PORT || 8080;
+
+app.use(bodyParser.json());
+
+app.post('/', (req: Request, res:Response) => {
+  console.log('Incoming pubsub message', JSON.stringify(req.body))
```
jeantanzj, line 10: is this supposed to trigger something? didn't test because im not sure what it's supposed to do

serverless/verify-backup/src/notify-slack/index.js
```diff
@@ -0,0 +1,25 @@
+const { IncomingWebhook } = require('@slack/webhook')
```
jeantanzj, line 1: Can we use sentry? Create another project / use the serverless one.

jeantanzj: this is actually not being used?

serverless/verify-backup/src/index.ts
miazima: When listing the objects in the bucket, the objects are ordered lexicographically by name. This would mean by the naming convention `YYYY-MM-DD/file-name` the latest files would always appear at the end of the list.

serverless/verify-backup/src/server.ts
miazima: As discussed, this is to accept the incoming POST request from the cloud scheduler which pings the cloud run instance, this in turn wakes up the instance and executes the entrypoint script. I'll elaborate on this in the PR description.

serverless/verify-backup/scripts/verify-backup.sh
miazima: Noted, will create a separate issue for that

serverless/verify-backup/src/notify-slack/index.js
miazima: I didn't write a script to deploy this, this lives as a cloud function in gcp which is triggered on failure to restore dump. Okay will include sentry in it.

jeantanzj: haha I meant to use sentry instead of the slack webhook

jeantanzj: because sentry already sends results to slack

jeantanzj (approved):
TODO: 
- Alert on success
- Change alerts to use sentry

PR #738, opened 2020-09-24
feat: Check file integrity on upload and download from S3

### LINKED ISSUE #715: Add computeChecksums to uploads to s3
https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html
computeChecksums (Boolean) — whether to compute checksums for payload bodies when the service accepts it (currently supported in S3 only)

We need to verify the integrity of file when they're uploaded

lamkeewei:
## Problem

We are not checking the integrity of the CSV files during both upload from frontend and download on backend for processing. This may lead to the following issues:
1. Uploads might get corrupted in-transit to S3 during upload from the frontend (network issues, etc).
2. Small possibility that files might be modified during the time window between frontend completing the upload to S3 and the `/upload/complete` call is made. 

Closes #715 

## Solution
**Backend**
- Add `GET /upload/start-v2` endpoint to support the use of `Content-MD5` in the generation of pre-signed URLs
- Add `POST /upload/complete-v2` endpoint to support use of `If-Match` for [verification of ETag](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestSyntax) when downloading CSV file for processing in the backend

**Frontend**
- Include the `Content-MD5` header during upload on the frontend. S3 uses this header to verify the integrity of a file upload. An error will be thrown if this value differs from the one that S3 calculates after receiving the file 
- Frontend passes `ETag` on completion of upload. This will be used by the backend in the `If-Match` header when calling `GetObject` to ensure that we are retrieving the specific uploaded file

## For discussion
**Multipart Uploads** 
This PR does not include support for MD5 checksums for multipart uploads. Only ETag verification is done.

`Content-MD5` must be passed during generation of pre-signed URLs. However, pre-signed URL for each chunk is [currently generated ahead](https://github.com/opengovsg/postmangovsg/blob/43910bcbec4577660c3e865a9565f0c19e9d86a3/frontend/src/services/protect-csv.service.ts#L101) of transforming/uploading the content. In order to add multipart upload support, this requires changing the generation of pre-signed URL for each chunk to after encryption is completed. This adds the overhead of an additional API call per chunk to the backend in order to retrieve the pre-signed URL with the `Content-MD5` header. 

**Versioning of endpoint**
Eventually, MD5 checksums and ETag verification should be made compulsory for uploads to S3. However, changing the request params on the existing `/upload/start` and `/upload/complete` endpoint would break any existing API user's scripts. However, the changes to these endpoints do not warrant a bump on the overall version of the API (moving to `/v2/...`.  As a stop-gap, `-v2` is appended to the affected endpoint. 

## Tests
_Run the tests against our staging S3 bucket. There are slight variations between localstack and S3 on the verification of ETags and MD5 checksums._

**Basic checks**
- Upload CSV for email/sms/telegram campaign
- Upload CSV for protected email campaign
- Test with a large CSV file (> 10MB) to ensure that MD5/multipart upload works with more than 1 chunk

**Backward compatibility**
- Test creation of campaign using the API. Use the v1 routes for `/upload/start` and `/upload/complete`. Reference our [API usage documentation](https://github.com/opengovsg/postmangovsg/blob/develop/docs/api-usage.md) 

**Wrong MD5**
- Modify the [frontend upload service](https://github.com/opengovsg/postmangovsg/blob/742243ccbe939178384cb85a4c734b1f3e7b3726/frontend/src/services/upload.service.ts#L67) locally to include an invalid MD5 checksum (hardcode a value)
- An error should be raised during uploading the CSV file

**Wrong ETag**
- Modify the [backend S3 client](https://github.com/opengovsg/postmangovsg/blob/742243ccbe939178384cb85a4c734b1f3e7b3726/backend/src/core/services/s3-client.class.ts#L23) locally to include an invalid `IfMatch` value (hardcode an invalid etag value)
- An precondition failure error should be raised during uploading the CSV file

## Deploy Notes
**New dependencies**:
- `spark-md5` : Used to calculate MD5 checksum on the frontend. SubtleCrypto doesn't support MD5 digest.


chinying (commented):
Tested locally and on staging, works as intended. 

My only issue is that the error messages shown to the user are pretty unclear. 

Eg. When the ETag for the backend doesn't match, the user sees `At least one of the pre-conditions you specified did not hold`, and for the frontend when the MD5 checksum doesn't match, the user sees `Forbidden`, perhaps we should inform the user to contact us instead?

chinying (approved):

PR #755, opened 2020-09-28
Custom domain frontend

jeantanzj:
For convenience of testing, you may check out https://github.com/opengovsg/postmangovsg/tree/custom-domain-test
A checklist for cases to test for is available on #728

**How to review a PR**
Functionality
- [ ] Is it usable on desktop?
- [ ] Is it mobile responsive?
- [ ] Does it transit to the expected states?
- [ ] Does it show expected errors? 
- [ ] Are the errors informative enough / too informative?

Code style
I'm horrible at frontend, but there are general things to watch out for
- [ ] Does it follow the code base's style?
- [ ] Does it introduce code smells (https://refactoring.guru/refactoring/smells https://refactoring.guru/design-patterns/typescript)?
- [ ] Do the variable names make sense?
- [ ] (frontend) Does the CSS use mixins and variables, or is it a mess of hardcoded values?
- [ ] Is there repeated code that can be extracted and reused? 
- [ ] Is there a smarter way to accomplish the objective? 
- [ ] Are there things that should be documented? (eg special regexes, numbers, convoluted logic)
- [ ] Is the documentation accurate? 
- [ ] Are the checks for falsy values correct? esp checks like `if(value)`
- [ ] Are errors being logged when they should be logged?

Screenshot 2020-09-29 at 5 00 14 AM
Screenshot 2020-09-29 at 5 00 24 AM
Screenshot 2020-09-29 at 5 00 32 AM
Screenshot 2020-09-29 at 5 00 41 AM
Screenshot 2020-09-29 at 5 01 02 AM
Screenshot 2020-09-29 at 5 01 11 AM
Screenshot 2020-09-29 at 5 01 19 AM
Screenshot 2020-09-29 at 5 01 39 AM
Screenshot 2020-09-29 at 5 02 01 AM
Screenshot 2020-09-29 at 5 02 15 AM


frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -60,17 +75,30 @@ const EmailTemplate = ({
...
+    const fromAddresses = await getCustomFromAddresses()
+    const options = fromAddresses.map((v) => ({ label: v, value: v }))
+    setCustomFromAddresses(options)
+  }
+
+  // Get custom from addresses
+  useEffect(() => {
+    populateFromAddresses()
+  }, [])
+  // useEffect(() => {
```
jeantanzj, line 91: remove this

frontend/src/components/dashboard/settings/Settings.tsx
```diff
@@ -100,6 +129,9 @@ const Settings = () => {
             <Route exact path="/settings/api">
               <ApiKey hasApiKey={hasApiKey} />
             </Route>
+            <Route exact path="/settings/email">
+              <CustomDomain onSuccess={fetchCustomFromAddresses} />
```
lamkeewei, line 133: Given that we call `getCustomFromAddresses` in this top-level component, perhaps we can pass the list of From Addresses down as a prop instead of calling it again in `populateFromAddresses` in the `CustomDomain` component. This will avoid holding state in the `CustomDomain` component.

frontend/src/components/dashboard/settings/Settings.tsx
```diff
@@ -56,12 +70,23 @@ const Settings = () => {
...
   }
 
+  function onAddCustomDomainClicked() {
+    modalContext.setModalContent(
+      <AddCustomDomainModal onSuccess={fetchCustomFromAddresses} />
+    )
+  }
+
   async function fetchUserSettings() {
     const { hasApiKey, creds } = await getUserSettings()
```
lamkeewei, line 80: `GET /settings` currently returns info about the API key as well as a list of credentials. Will it make sense to include the custom From Addresses as part of the response from this call?

frontend/src/components/dashboard/settings/verify-custom-domain-modal/VerifyCustomDomainModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+            </PrimaryButton>
+          </div>
+        )
+      // Credentials failed to store
+      case VerifyEmailStep.Failure:
+        return (
+          <div className={styles.centerAlign}>
+            <img src={FailureImage} alt="" />
+            <h3>Sorry, something went wrong.</h3>
+            <ErrorBlock>{errorMessage}</ErrorBlock>
```
lamkeewei, line 86: When verification fails, the error message is not very informative. It's not clear how they can remedy the verification failure. In this case, the cause could either be that the agency's IT has not added or has removed the necessary DNS records. 

Perhaps we can add a message to tell them to try again in a while (while their IT team add the records) or to contact us. 

94486472-103c8280-0212-11eb-8007-3accced4b8dc

frontend/src/components/dashboard/settings/custom-domain/CustomDomain.tsx
```diff
@@ -0,0 +1,130 @@
...
+            to={i18n._(LINKS.customDomainRequestUrl)}
+            target="_blank"
+          >
+            <PrimaryButton className={styles.request}>
+              Fill in request form <i className="bx bx-right-arrow-alt"></i>
+            </PrimaryButton>
+          </OutboundLink>
+          <PrimaryButton
+            className={styles.status}
+            onClick={() => onVerifyFromAddressClicked(email)}
```
lamkeewei, line 110: Quick check, will users be allowed to customise the display name of their From Address?

jeantanzj: I didn't add an input box for that, but the backend supports a custom display name.

frontend/src/components/dashboard/settings/Settings.tsx
jeantanzj: Is there a particular advantage in doing so, besides saving one network call?  the endpoint GET /email/from has to exist, for the dropdown in EmailTemplate.  IMO it doesn't fit into getUserSettings because the email_from_address table does not associate the email address to a user. 

Alternatively, if you want to centralize the retrieval of credentials / from addresses, then maybe we should also get rid of GET /telegram/credentials and GET/sms/credentials

lamkeewei: Agree on no additional benefit beside saving the one API call. `GET /telegram/credentials` and `GET /sms/credentials` also can't be removed since they're used in the creation flow too. There's also no good way to link between the user and From Addresses he's allowed to access. 

I think we can keep this as it is.

frontend/src/components/dashboard/settings/custom-domain/CustomDomain.tsx
jeantanzj: Added the input box

frontend/src/components/dashboard/settings/verify-custom-domain-modal/VerifyCustomDomainModal.tsx
jeantanzj: Will change in backend

lamkeewei (approved):
Works well. Might want to indicate in the label input placeholder that the field is optional.

PR #775, opened 2020-10-06
Release v1.15.0

jeantanzj:
## Features
- Support custom domain

## Improvements
- Documentation
- Check file integrity on upload and download from s3
- Add encrypted backup

### Add encrypted backup
Check that these are set in travis
- Production
   - `PRODUCTION_BACKUP_ROLE` : Backup lambda IAM role ARN
   - `PRODUCTION_BACKUP_DB_URI` : Database URI to use for `pg_dump`. Use read replica.
   - `PRODUCTION_BACKUP_BUCKET_NAME` : S3 bucket name to store backup in
   - `PRODUCTION_BACKUP_PUBLIC_KEY` : Public key used to encrypt encryption key

### Support custom domain
 Backfill db
``` sql
alter table email_templates add column "from" text;
select count(*) from email_templates where "from" is null;
update email_templates set "from" = 'Postman.gov.sg <donotreply@mail.postman.gov.sg>' where "from" is null;
```

### Add encrypted backup 
- [ ] Trigger lambda for creating backup 
- [ ] Download backup and decrypt it using local key and gcloud 

### Check file integrity
- [ ] Upload file with correct md5 - should be rejected
- [ ] Upload file with wrong md5
- [ ] Upload protected email file (multipart)

### Support custom domain
- [ ] GET /settings/email/from should return an array of usable custom from addresses, including the default email
- [ ] POST /settings/email/from/verify should accept emails in the form of `email@email.com`, and `DISPLAY_NAME <email@email.com>`

**Verification of From address**
- [ ] If the default email is supplied, verification on AWS and verification of DKIM records should be skipped 
- [ ] If the default email is supplied, a validation message should still be sent out
- [ ] A validation message should be sent from the custom from address
- [ ] If the email is not verified on AWS, it cannot be saved  in email_from_address
- [ ] If the email's DKIM records are not set, it cannot be saved  in email_from_address
- [ ] If the email is verified and its DKIM records are correct, it should be saved into email_from_address
- [ ] If a from address is supplied with a display name, the display name should be saved into the name column, and the email address into the email column.
- [ ] If the same from address is supplied with a different display name, the display name should be upserted into the name column. 
- [ ] If the same from address is supplied with no display name, a null should be upserted into the name column. 
- [ ] A user can only verify the email address that they logged in with as a from address. 

**Use of From address**
- [ ] The custom from address and default email should be selectable when creating email template
- [ ] A user can use the default email to create the email template 
- [ ] A user can use the email they logged in with as a from address
- [ ] A user cannot use any other valid email addresses in email_from_address to send a message, if they are not logged in with that particular email (limited access).
- [ ] The test message should be sent from the custom from address
- [ ] The email messages from that campaign should be sent from the custom from address

**Previously valid From Address which became invalid**
- [ ] If the custom from address exists in email_from_address, and is selected, but has since become invalid (either it was removed from SES, or its DKIM records were modified), the template should not be saveable. 
- [ ] If the template was saved with a custom from address which has since become invalid, the `*test message* step should fail.
- [ ] If the template was saved with a custom from address, which has since become invalid,  and the `*test message* has already been sent, the *send campaign* step should result in errors for all the email messages in that campaign.


jeantanzj:
>  If the template was saved with a custom from address which has since become invalid, the `test message step should fail.

This actually doesn't fail if dkim records have been modified. It only fails if the email has been removed from SES

PR #780, opened 2020-10-07
feat: Include successfully sent messages in export

### LINKED ISSUE #489: Customisable exports for campaign statistics
Currently, export on campaign detail page only support messages that have either `INVALID_RECIPIENT` or `ERROR` statuses. Users might be interested in having the ability to export messages that have other statuses (`SUCCESS`, etc). 

For example, users can determine who in their recipient list is subscribed to their Telegram bot by generating an export for messages that have a `SUCCESS` status. While this can be technically accomplished now by generating a diff between their uploaded recipient list and the exported error list, it maybe not be immediately obvious to less technical users. 

Export functionality should allow the user to specify which statuses should be included in the generated CSV file.

lamkeewei:
## Problem

Add successfully sent messages in the exported recipient list.

Closes #489 

## Solution
**Features**:
- Include messages with "SUCCESS" status in `GET /export` endpoint
- Show export button on frontend even if there are no failed recipients

**Improvements**
- Added loading spinner for "Delivery report" Button while exporting

**Bug fixes**
- Standardised the copywriting on the campaign details page for all channels. Telegram was different previously.

### Backend generation of CSV
Decided not to implement backend generation/streaming of the responses to the frontend as the added complexity was not worth it given the following reasons:
- Typical number of recipients for our campaigns are < 1000
- Even in the worst case (200k recipients), memory usage in the browser maxes out at ~130MB
- Shifting the CSV generation to the backend is just shifting the memory load to the backed. Multiple concurrent large requests can degrade the performance of the backend. Decentralising the generation of the CSV to the frontend would be a better choice.

image

## Tests
_Note: Since this feature identify successfully sent messages using the `SUCCESS` status, it's easiest to test with Telegram campaign locally as the other channels relies on a callback to update the final status to `SUCCESS`._

- Send a new campaign with only successful recipients
- Download CSV by clicking "Delivery report"
- Verify that CSV report has SUCCESS recipients

### DIFF
--- a/backend/src/core/interfaces/campaign.interface.ts
+++ b/backend/src/core/interfaces/campaign.interface.ts
@@ -62,10 +62,11 @@ export interface CampaignStatsCount {
   updated_at: Date
 }
 
-export interface CampaignInvalidRecipient {
+export interface CampaignRecipient {
   recipient: string
   status: string
   updated_at: Date
+  error_code?: string
 }
 
 /**
--- a/backend/src/core/services/campaign.service.ts
+++ b/backend/src/core/services/campaign.service.ts
@@ -73,12 +73,6 @@ const listCampaigns = ({
           ['updated_at', 'status_updated_at'],
         ],
       },
-      {
-        model: Statistic,
-        attributes: [
-          [literal('(errored + invalid) > 0'), 'has_failed_recipients'],
-        ],
-      },
     ],
   }
   if (offset) {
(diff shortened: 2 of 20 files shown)


chinying (approved):
Tested and lgtm.

If we ever have to worry about downloads taking up too much memory we can always pipe the exports to S3 and let the browser handle the streaming on download.

Issue #729, opened 2020-09-23
Style guide for logging


lshaowei18:
**Is your feature request related to a problem? Please describe.**
The logs that we have now are messy, making it harder to troubleshoot.

**Describe the solution you'd like**
Style guide or patterns that the team should follow. It could be: 
- Establishing a format that each line of log should follow and what information it should contain. Eg: `2018-07-29 | 21:10:29.178 |  INFO  | Service started in 3434 ms.`
- Explain the different levels of logging & give relevant examples from the codebase of when & why a level is picked. (Levels referring to Debug, Info, Warning, Error)

**Describe alternatives you've considered**
Thought of formatting it in a way that can be used by dashboards and metrics services like Grafana, still unsure if it's an overkill.

**Additional context**

jeantanzj:
They should be namespaced, perhaps in a meta property (https://www.npmjs.com/package/winston#streams-objectmode--and-info-objects). 

Ideally, we should log in json (https://www.scalyr.com/blog/aws-lambda-logging-best-practices/) and the format might look like 

`{ level: 'info', datetime: '2020-09-23T01:03:00.000Z', namespace: 'serviceNameOrFunctionName',  message: 'some message'}`

winston-cloudwatch formats our json logs into text right now

PR #781, opened 2020-10-08
[Monitoring] Improve application logs

### LINKED ISSUE #743: [Monitoring] Application
- [ ] Key application transactions are logged. (As a team maybe we have to decide what are the key transactions)
- [ ] Dashboards created for situational awareness. (Decide what services to use, Grafana?)

miazima:
## Problem

Though we have implemented logging throughout our app, we want to ensure that we are capturing key transactions in our app and that our log formats allow us to efficiently analyse them. Structured log formats allow for easier searching and analysis in log management solutions.

Closes #743, also addresses #729

## Solution

**Features**:

- Use existing winston logger to introduce structured logging (previously we stored everything in the message attribute)
- Structured logging (use objects and properties to log key data such as campaignId, userId)
- Use a helper function to return logger instance with module name included in the metadata for logs in each module, include caller function names in logs

**Example:**
image

## Test
Observe cloudwatch `postmangovsg-staging-beanstalk-logs` to see log examples

## Todo
Update alarm metrics


jeantanzj:
note that metrics for alarms probably have to be changed.

backend/src/core/logger.ts
```diff
@@ -1,4 +1,13 @@
+import path from 'path'
 import winston from 'winston'
+import requestTracer from 'cls-rtracer'
+
+const getModuleLabel = (callingModule: NodeModule): string => {
+  // Remove the file extension from the filename and split with path separator.
+  const parts = callingModule.filename.replace(/\.[^/.]+$/, '').split(path.sep)
+  // Join the parts of the file path after build directory together
+  return path.join(...parts.slice(parts.lastIndexOf('build') + 1))
```
jeantanzj, line 9: if we trust that 'build' is indeed the name of the build directory, then this is equivalent
```js
const moduleName =  callingModule.filename
moduleName.substring(moduleName.lastIndexOf('build')+6, moduleName.lastIndexOf('.'))
```

backend/src/core/loaders/express.loader.ts
```diff
@@ -29,9 +32,23 @@ morgan.token('user-id', userId)
 // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
 // @ts-ignore
 const loggerMiddleware = morgan(config.get('MORGAN_LOG_FORMAT'), {
-  stream: logger.stream,
+  stream: loggerInstamce.stream,
```
jeantanzj, line 35: spelling... but actually for our sanity, could we add a getter and change this to 
`Logger.getStream()`

backend/src/core/logger.ts
```diff
@@ -29,7 +43,21 @@ class Logger implements LoggerInterface {
...
+      debug: (logMeta: any): winston.Logger =>
+        this.logger.debug({ label, ...logMeta }),
+      warn: (logMeta: any): winston.Logger =>
+        this.logger.warn({ label, ...logMeta }),
+    }
+  }
 }
 
-const logger = new Logger().logger
+const logger = new Logger()
```
jeantanzj, line 62: Does Logger really need to be a class? It's supposed to be a singleton. So technically, this module could be restructured and exported as
```js
const getModuleLabel = ()=>{}
const logger = winston.createLogger(...)
logger.stream = ...
const loggerWithLabel = () => {}
export {
getStream = () => (logger.stream),
loggerWithLabel
}
```

My aim is to prevent devs from accidentally using logger.logger instead of loggerWithLabel. 

Alternatively, don't change the class, but just export the relevant methods
```js
const logger = new Logger()
export { getStream = () => (logger.stream), loggerWithLabel: (...args) => (logger.loggerWithLabel(...args))}
```

backend/src/core/middlewares/auth.middleware.ts
```diff
@@ -127,6 +168,12 @@ const logout = async (
       if (!err) {
         resolve(res.sendStatus(200))
       }
+      logger.error({
```
jeantanzj, line 171:
```js
if (!err) {...} 
else { 
logger.error()
reject(err)
}
```

This line is currently being printed regardless of success

backend/src/core/middlewares/auth.middleware.ts
```diff
@@ -101,9 +132,19 @@ const isCookieOrApiKeyAuthenticated = async (
...
+      logger.info({
+        message: 'User authenticated by API key',
+        email: user.email,
+        id: user.id,
+        action: 'isCookieOrApiKeyAuthenticated',
+      })
       return next()
     }
 
+    logger.error({
```
jeantanzj, line 144: This doesn't provide much info... we don't know who provided the invalid api key / didn't provide an api key / was not a user / did not have a ssession. I don't think this needs to be logged actually

backend/src/core/middlewares/campaign.middleware.ts
```diff
@@ -22,6 +25,11 @@ const canEditCampaign = async (
     if (!hasJob && !csvStatus?.isCsvProcessing) {
       return next()
     }
+    logger.error({
```
jeantanzj, line 28: I think the main point of logging is for troubleshooting errors. In this case, and also for other transactions, perhaps you could consider whether the data that you've logged would be helpful for debugging, or adds extra information to the http status of the request that would have been logged with morgan.

backend/src/core/services/parse-csv.service.ts
```diff
@@ -42,6 +45,7 @@ const parseAndProcessCsv = async (
       },
 
       chunk: async (rows: ParseResult<CSVParams>, parser: Papa.Parser) => {
+        const logMeta = { readStream, action: 'parseAndProcessCsv.chunk' }
```
jeantanzj, line 48: Don't log the readstream

backend/src/email/utils/callback/parsers/sendgrid.ts
```diff
@@ -43,7 +45,11 @@ const isEvent = (req: Request): boolean => {
 
 const parseRecord = async (record: SendgridRecord): Promise<void> => {
   if (record.message_id === undefined) {
-    logger.info(`No reference message id found for ${record['smtp-id']}`)
+    logger.info({
+      message: 'No reference message id found',
+      smptpId: record['smtp-id'],
```
jeantanzj, line 50: smtpId

jeantanzj:
The implementation is sound,  but I feel that some logs are superfluous and don't aid in debugging. I can pick this up and change/remove some of the logs if you're happy for me to do so.

miazima:
@jeantanzj I have reviewed the log statements and attempted to remove logs that seemed not as useful for debugging purposes, feel free to change or remove logs as you perceive!

jeantanzj (approved):
<3

PR #788, opened 2020-10-19
feat: lambda that updates environment when secrets manager is updated

jeantanzj:
## Problem
When environment variables for the elastic beanstalk environment are updated in Secrets' Manager, the developer may forget to redeploy the environment.

## Solution
**EDIT: Reverted to option 1 because keewei saw that the configuration using rolling deployments with rolling update type would enable the containers to be restarted in a rolling fashion.**

Option 1. Restart app server
Originally I wanted to go with this as it was super fast. But restarting the app server appeared to cause downtime as all of the containers would be restarted at once, instead of in a rolling fashion.

Option 2. Pretend to update an env var
This implementation updates the elastic beanstalk environment with the request id of the lambda invocation, to trigger a redeploy. There is now no time savings from using secrets manager as the immutable deployment will be triggered.

Steps taken:
1. Create a cloudtrail for AWS Secrets manager events (by default, we already have this because we log all the events that affect our AWS account)
2. Create a rule in Cloudwatch that watches for events from AWS Secrets Manager, with a specific operation (PutSecretValue) and specific secret id, and invokes the lambda when the event occurs. 
Example of event pattern
```json
{
   "source": [
      "aws.secretsmanager"
   ],
   "detail-type": [
      "AWS API Call via CloudTrail"
   ],
   "detail": {
      "eventSource": [
         "secretsmanager.amazonaws.com"
      ],
      "eventName": [
         "PutSecretValue"
      ],
      "requestParameters": {
         "secretId": [
            "env/eb/<environment_name>"
         ]
      }
   }
}
```

### DIFF
--- a/serverless/eb-env-update/src/config.ts
+++ b/serverless/eb-env-update/src/config.ts
@@ -43,13 +43,6 @@ const config = convict({
     env: 'SECRET_ID',
     format: 'required-string',
   },
-  aws: {
-    awsRegion: {
-      doc: 'Region for where the lambda will be deployed in.',
-      default: 'ap-southeast-1',
-      env: 'AWS_REGION',
-    },
-  },
   sentryDsn: {
     doc: 'Sentry DSN for serverless',
     default: '',
--- a/serverless/eb-env-update/src/config.ts
+++ b/serverless/eb-env-update/src/config.ts
@@ -26,6 +26,12 @@ convict.addFormats({
 })
 
 const config = convict({
+  env: {
+    doc: 'The application environment',
+    format: ['production', 'staging', 'development'],
+    default: 'production',
+    env: 'NODE_ENV',
+  },
   prefix: {
     doc: 'prefix of secret id',
     default: 'env/eb/',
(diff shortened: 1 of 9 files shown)


jeantanzj:
> When you specify a batch size for rolling updates, Elastic Beanstalk also uses that value for rolling application restarts. Use rolling restarts when you need to restart the proxy and application servers running on your environment's instances without downtime.
Source: https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features.rolling-version-deploy.html

Let me change staging to rolling instead of all-at-once and see if it helps

serverless/eb-env-update/src/config.ts
```diff
@@ -0,0 +1,51 @@
...
+    env: 'PREFIX',
+  },
+  secretId: {
+    doc: 'Id of secret that got updated',
+    default: '',
+    env: 'SECRET_ID',
+    format: 'required-string',
+  },
+  aws: {
+    awsRegion: {
```
lamkeewei, line 41: Seems like this config variable is unused?

lamkeewei (approved):
Tested and working on staging. Might be useful to add Sentry monitoring to flag possible misconfigurations (e.g. secretId in config does not match the one in `requestParameters`, etc) and errors from the API calls. 

On another note, `updateEnvironment` is still causing Pingdom to randomly detect downtime for staging even with rolling updates with additional batch.

PR #790, opened 2020-10-20
Trial mode backend

jeantanzj:
## Problem
Users find it difficult to onboard Postman (SMS and Telegram) because they have to procure credentials.

## Solution
In this PR, I've introduced a demo mode, where any public officer would be able to use Postman's SMS credentials and Telegram bot toke to send out a limited number of messages.
This enables the public officer to gain familiarity with Postman's workflow, and test the system before going through a cumbersome procurement process.

My main considerations for this design are:
1) a user should have separate trials for SMS and Telegram (instead of combining the number of demos) as they might mistakenly use all the demos for one channel
2) the developer should be able to easily add a number of demos to a user, as our agency users are likely to ask for them
3) the developer should at their discretion be able to increase the number of messages

With the abovementioned points in mind, I made the following changes: 
1) added a `user_demos` table holding the number of demos the user has for each channel, and a column indicating the whether to show demo mode to the user
2) when a campaign is created in demo mode for that channel, the number of demos remaining is decremented by one
3) added an additional column to the `campaigns`table, called `demo_message_limit`. It defaults to null for non-trial campaigns, but accepts a value between 1-20 messages (via API, validated by Joi).
4) when a csv is uploaded and parsed, if the number of records read exceeds the `demo_message_limit`, an error will be thrown and the upload will be rejected. 
5) only campaigns in demo mode can use the `SMS_DEFAULT` credential (upserted into the `credentials` table when the app is started). This credential has been added to Secrets Manager
6) When the trial mode bar is opened or closed on the frontend, `user_demos.is_displayed` will be updated

As a sidenote, I initially planned to introduce a `user_features` table, holding the `num_demos_sms` or `num_demos_telegram` columns, so that we could manage features at a user level. I abandoned this idea for now, because a feature management implementation might look more like 

```sql
user_features (user_id int fk(users), demos_enabled boolean, phonebook_enabled boolean)
user_demos (user_id int fk(users), num_demos_sms int, num_demos_telegram int)
-- phonebook stuff
```

## Deployment notes

- [ ] Add column to campaigns
`alter table campaigns add column demo_message_limit integer default null;`

- [ ] Be sure to backfill user_demos
`
INSERT INTO user_demos (user_id, num_demos_sms, num_demos_telegram, is_displayed, created_at, updated_at) (SELECT id as user_id, 3,  3,  true, NOW(), NOW() FROM users);
`
- [x] Add SMS credential to secrets manager
- [x] Add Telegram credential to secrets manager


backend/src/core/constants.ts
```diff
@@ -20,6 +20,12 @@ export enum MessageStatus {
   InvalidRecipient = 'INVALID_RECIPIENT',
 }
 
+export enum DefaultCredentialName {
+  Email = 'EMAIL_DEFAULT',
+  SMS = 'SMS_DEFAULT',
```
lamkeewei, line 25: Perhaps we want to read the default credential names for SMS/Telegram from config so that we can have the option of using different creds for each deployment environments (staging, production).

jeantanzj: yep setting it up now demo/staging/SMS_DEFAULT

jeantanzj:
```  if (label === DefaultCredentialName.SMS) {
      const campaign = await SmsService.findCampaign(+campaignId, userId) // TODO: refactor this into res.locals
      if (campaign.demoMessageLimit) {
        credentialName = `demo/${config.get('env')}/${label}`
      } else {```

lamkeewei (approved):
Works as expected. Tested the following on local and staging:
1. Send demo campaigns for Telegram and SMS using default credentials
2. Max number of demo campaigns is enforced
3. Non-demo campaign do not have access to default credentials
4. Max demo recipients is checked and error thrown is not within limits
5. Hide/show demo banner preference is persisted

Might want to add docs to get developers to seed either localstack or Secrets Manager with the default SMS/Telegram credentials if they want to test out trial mode.

PR #791, opened 2020-10-20
feat: Restrict allowed image sources for protected mail message B

### LINKED ISSUE #575: Images not displaying in registered mail (CORS)
**Describe the bug**
Images from other domains like moe.gov.sg is not being shown on staging.
**To Reproduce**

1) Include a img src that is from other domains in Template B, registered mail
2) Unlock the mail

**Expected behavior**
Image should show.

**Screenshots**
Screenshot 2020-07-17 at 6 32 33 AM

**Additional context**
We could let the users know when they are filling Template B that images from other domains will not show, and direct them to use file.go.gov.sg to host their images.

lamkeewei:
## Problem

Validate the host of images. This is currently only enabled for protected email Message B but can be extended to other channels.

Closes #575 

## Solution

**Features**:
- Extended XSS options to filter for image src for protected mail message B

**Bug fixes**
- Update Telegram XSS options to allow the use of variables in `href` value

## Before & After Screenshots
**Error when preview cannot be generated**

**Error when templating fails for non-preview rows**

## Tests
**Template module**
- Run unit tests to ensure that errors are thrown when an image src is not in the allowed list

**Allow file.go.gov.sg hosted image**
- Create a protected email campaign template with an image hosted on `file.go.gov.sg` as the src
```
<img src='https://file.go.gov.sg/...' />
```
- Upload recipient list and an error should not be shown

**Invalid image source in template**
- Create a protected email campaign template with an invalid image source
```
<img src='https://invalid.com/image.jpg' />
```
- Proceed to upload recipient list
- An error should be shown flagging out the line that has the invalid image source

**Invalid image source in recipient list variable**
- Create a protected email campaign template with an invalid image source
```
<img src='{{profileImage}}' />
```
- Proceed to upload the following recipient list
```
recipient,password,profileImage
test@test.com,password,https://invalid.com/profile.jpg
```
- An error should be shown flagging out the line that has the invalid image source


modules/postman-templating/src/template-client.ts
```diff
@@ -2,16 +2,37 @@ import { mapKeys } from 'lodash'
...
+    allowedImageSources,
+  }: {
+    xssOptions?: xss.IFilterXSSOptions
+    lineBreak?: string
+    allowedImageSources?: Array<string>
+  }) {
+    this.xssOptions = xssOptions || {}
+    this.lineBreak = lineBreak || '<br />'
+
+    if (allowedImageSources) {
```
jeantanzj, line 25: what if it's an empty array

jeantanzj:
did you intend to validate only the first record?

lamkeewei:
> did you intend to validate only the first record?

Nope. It's a bug. Let me look into it.

modules/postman-templating/src/template-client.ts
```diff
@@ -23,10 +23,27 @@ export class TemplateClient {
...
+      this.validateImageSources(allowedImageSources)
       this.xssOptions = filterImageSources(this.xssOptions, allowedImageSources)
     }
   }
 
+  /**
+   * Validate image sources
+   * @param sources
+   */
+  validateImageSources(sources: Array<string>): void {
```
lamkeewei, line 35: I only did non-empty strings checks for the list of sources passed in. Not sure if that's sufficient or the validation should be more thorough (check valid hostname string, etc). 

If only non-empty string check is required, alternatively, the type of `allowedImageSources` can be further restricted to something like `allowedImageSource?: [string, ...string[]]` to use compile time checks. Let me know which is preferred!

frontend/src/services/validate-csv.service.ts
```diff
@@ -62,8 +62,18 @@ export async function validateCsv({
             preview = hydrateTemplate(template, row, removeEmptyLines)
           }
         } catch (e) {
+          // Exit early if it is a templating error. This can be caused by either the template or first row having invalid values.
```
lamkeewei, line 65: If we want to validate every single row for an invalid image source at this stage, we will have to run `templateClient.template` for every row. This is will be an additional duplicate call for each row before the processing done later on in `protectAndUploadCsv`.

lamkeewei:
> > did you intend to validate only the first record?
> 
> Nope. It's a bug. Let me look into it.

Turns out validation is done for beyond the first record. It's just that it happens during the `protectAndUploadCsv` step.

modules/postman-templating/src/template-client.ts
```diff
@@ -23,10 +23,27 @@ export class TemplateClient {
     this.lineBreak = lineBreak || '<br />'
 
     if (allowedImageSources) {
+      this.validateImageSources(allowedImageSources)
```
jeantanzj, line 26: how about setting the allowedImageSources here?
this.allowedImageSources = allowedImageSources.filter(Boolean)

modules/postman-templating/src/template-client.ts
jeantanzj: didnt you check for valid hostname in safeAttrValue

jeantanzj:
Meh. 
this.allowedImageSources isn't used anywhere. Didnt need to be stored. I'll make a change ok

jeantanzj (approved):

PR #794, opened 2020-10-21
fix: Preview component not rendering HTML for message body

lamkeewei:
## Problem

Preview block was escaping the message body. This results in the raw HTML code being shown instead of rendering the HTML. Users will not be able to see how their rendered message body will look like until they send a test message.

## Solution

**Bug Fixes**:
- Remove escaping of message body for PreviewBlock. Escaping is still preserved for other fields (From, Subject, Reply To).

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

## Tests
- Create a new email campaign with the following template and choose `Postman.gov.sg <donotreply@mail.postman.gov.sg>` as the From email
```
<b>bold</b>, bold
<i>italic</i>, italic
<u>underline</u>, underline
strikethrough, strikethrough, strikethrough
<b>bold <i>italic bold italic bold strikethrough <u>underline italic bold</u></i> bold</b>
<a href="https://www.google.com/">Google</a>
```
- Check that the preview block is rendering the HTML and the From is not being truncated

### DIFF
--- a/frontend/src/components/common/preview-block/PreviewBlock.tsx
+++ b/frontend/src/components/common/preview-block/PreviewBlock.tsx
@@ -33,13 +33,17 @@ const PreviewBlock: React.FunctionComponent<PreviewBlockProps> = ({
   }
 
   function constructHtml() {
-    function h(name: string, value?: string | null) {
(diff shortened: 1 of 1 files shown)


jeantanzj (approved):
lgtm

Issue #795, opened 2020-10-22
Refactor Postman Frontend


miazima:
Postman frontend has grown larger in the past few months with the addition of new features and channels. Our frontend works smoothly, but our code could be cleaner and less repetitive. This would help us understand the code better with less clutter, onboarding new engineers to a better structured code base and possibly improve our app’s performance. 

**Possible areas of refactoring**
- Introduce React context for campaign state management
   -  Removes the need for passing `onNext` down to channel-specific components
   - Removes the need for `finishLaterCallbackRef`
   - Removes the need to pass campaign props from Create component to <Channel>Create component, to all channel specific components
- Abstract out duplicated logic in channel components
  - Explore Higher Order Components to achieve some sort of component factory
- Prevent setting state after component is dismounted
- Explore using custom hook/react-query for observing status of asynchronous queries

Refer to the discussion document for more detailed evaluations
https://docs.google.com/document/d/1epUV4A0tjYn8RHKt7cJIDrcx6WIGhiNy3SY8QNpHKJQ

PR #799, opened 2020-10-26
Introduce React Context for campaign state management

miazima:
Create a context to maintain the ‘global’ state of the current campaign during the create-campaign flow. 

### Improvements
- Removes the need for passing `onNext` down to channel-specific components
   - Can update the campaign state from any component subscribed to context
   - Components will update state of campaign state in component e.g. `setCampaignContext({ ...campaign, subject, body })`
- Removes the need for` finishLaterCallbackRef`
   - finishLaterCallbackRef was used to save current state of unsaved draft messages when finish later button was clicked
   - Instead use a finish later callback state, e.g. `const [finishLaterCallback, setFinishLaterCallback] = useState(null)`
- Removes the need to pass campaign props from Create component to <Channel>Create component, to all channel specific components - passing bulky props down two levels of component tree
   - Campaign state is shared across all channel components 

Part of efforts for #795

### Features
- Campaign context for campaign state management
- FinishLater context for finish later callback state management

### Considerations

Components subscribed to a context will re-render for every update to context values, hence resulting in unnecessary re-renders after context value changes.

- Created a separate context for finish later callback instead of including in campaign context
  - only re-renders specified subscribers
- From manually observing the performance of the app, there seemed to be no difference in performance. 
- To assess the changes on the render level, I used react dev tools profiler to profile the campaign creation flow events. Here is a comparison of before and after of the number of renders during an SMS campaign creation flow. From the profiler results, there is no significant increase in renders after introducing context.
https://docs.google.com/document/d/1TOVil0bsWdsftlPoGqTnsRwdpTRS7W_V1mU712exAOA/edit?usp=sharing


lamkeewei (changes requested):
Accessing of campaign state looks much cleaner now! 👍  

Was thinking if we can combine the functionality of `FinishLaterModalContext` into `CampaignContext`. 

Something along the of lines of just calling `campaignContext.saveDraft()` instead of having to set the `finishLaterContent` in each component and then clearing it on unmount might be cleaner. But I think that will require us to move `handleSaveTemplate` to the `CampaignContext` since the the logic is currently defined in each of the Template component. Perhaps something to think about when we further refactor. 

Other than that just need to resolve the merge conflicts. I think there's a new `onPrevious` method added.

frontend/src/components/dashboard/create/telegram/TelegramCredentials.tsx
```diff
@@ -21,14 +28,15 @@ import TelegramCredentialsInput from './TelegramCredentialsInput'
...
-  hasCredential: initialHasCredential,
-  onNext,
+  setActiveStep,
 }: {
-  hasCredential: boolean
-  onNext: (changes: any, next?: boolean) => void
+  setActiveStep: Dispatch<SetStateAction<TelegramProgress>>
 }) => {
+  const { campaign, setCampaign } = useContext(CampaignContext)
+  const { hasCredential: initialHasCredential, progress } = campaign
```
lamkeewei, line 39:
```suggestion
  const { hasCredential: initialHasCredential } = campaign
```

Minor issue, I think `progress` is unused.

frontend/src/components/dashboard/create/telegram/TelegramCredentials.tsx
```diff
@@ -371,10 +370,10 @@ const TelegramCredentials = ({
                 <ErrorBlock>{errorMessage}</ErrorBlock>
               </StepSection>
 
-              <ButtonGroup>
-                <NextButton disabled={!hasCredential} onClick={onNext} />
-                <TextButton onClick={onPrevious}>Previous</TextButton>
-              </ButtonGroup>
+              <NextButton
```
lamkeewei, line 373: Previous button is missing here.

frontend/src/components/dashboard/create/Create.tsx
```diff
@@ -23,59 +18,39 @@ const Create = () => {
...
+      setCampaign(campaign)
+      setLoading(false)
+    }
     loadProject(id)
-  }, [id])
+  }, [id, setCampaign])
 
   async function handleFinishLater() {
-    if (campaign.status === Status.Draft) {
+    if (campaign.status === Status.Draft && finishLaterContent) {
```
lamkeewei, line 39: Minor point. Behaviour for this is slightly changed. The GA event is now only sent for pages that trigger the save draft modal. It doesn't fire for the other pages (credentials, send).

jeantanzj:
82e994d -- this is a very strange commit? dont you have to fix the merge conflicts when you merge develop in

frontend/src/components/dashboard/create/Create.tsx
jeantanzj: I'll remove this and merge it in

jeantanzj (approved):

PR #809, opened 2020-10-29
Fix: Check if sanitised template is empty before upserting template

### LINKED ISSUE #792: Missing check for empty message template after XSS sanitisation
**Describe the bug**
Edge case. Message templates that only contain invalid HTML tags will result in an empty template being saved. The campaign will still be advanced to the recipient upload step. The CSV upload will fail to complete as the template is empty.

**To Reproduce**
Steps to reproduce the behaviour:

1. Create a new SMS campaign with the following template:
```
<b></b>
```
2. Press "Next". The campaign will be advanced to the recipient upload step
3. Try to upload a valid recipient CSV file
4. "Error completing file upload" will be shown

**Expected behavior**
An error should be thrown when trying to save a template that only contain invalid HTML tags. Template should not be upserted.

**Screenshots**

image

image

miazima:
## Problem

Saving a message template with only invalid HTML tags results in an empty template being saved to the database.

Closes #792

## Solution

Check that for empty message templates and that message template is not empty after it has been sanitised.

**Improvements**:
- Frontend: Trim whitespaces to check for empty message template to disable next button
- postman-templating: Add new function in template client to strip all html tags from input

**Bug Fixes**:

- Strip all tags and leading whitespace from raw template and check that message is not empty
- Do same check on frontend for protected email template before processing csv
- If message template is empty, throw a `TemplateError` with the error message `Message template is empty!`
- Check that sanitised message template is not empty before upserting into template table in database
- If sanitised message template is empty, throw a `TemplateError` with the error message `Message template is invalid as it only contains invalid HTML tags!`

## Tests

- [x] Create a sms campaign, and save the template with an invalid tag e.g. `<foo></foo>`
- [x] Save the template with only whitespaces and/or `<br>` tags
- [x] Save the template with an valid tags but with empty content e.g. `<p>    </p>`
- [x] An error message should be displayed
- [x] Campaign will not be advanced to the next step
- [x] Repeat for telegram and email (check subject and body)

### DIFF
--- a/backend/src/email/services/email-template.service.ts
+++ b/backend/src/email/services/email-template.service.ts
@@ -169,7 +169,9 @@ const storeTemplate = async ({
   const sanitizedSubject = client.replaceNewLinesAndSanitize(subject)
   const sanitizedBody = client.replaceNewLinesAndSanitize(body)
   if (!sanitizedSubject || !sanitizedBody) {
-    throw new TemplateError('Message template is empty!')
+    throw new TemplateError(
+      'Message template is invalid as it only contains invalid HTML tags!'
+    )
   }
   const updatedTemplate = await upsertEmailTemplate({
     subject: sanitizedSubject,
--- a/backend/src/email/services/email-template.service.ts
+++ b/backend/src/email/services/email-template.service.ts
@@ -3,7 +3,11 @@ import { difference, keys } from 'lodash'
 import { isSuperSet } from '@core/utils'
 import { HydrationError } from '@core/errors'
 import { Campaign, Statistic } from '@core/models'
-import { TemplateClient, XSS_EMAIL_OPTION } from 'postman-templating'
+import {
+  TemplateClient,
+  XSS_EMAIL_OPTION,
+  TemplateError,
+} from 'postman-templating'
 
 import { EmailTemplate, EmailMessage } from '@email/models'
 import { StoreTemplateInput, StoreTemplateOutput } from '@email/interfaces'
@@ -162,9 +166,14 @@ const storeTemplate = async ({
   from,
 }: StoreTemplateInput): Promise<StoreTemplateOutput> => {
   // extract params from template, save to db (this will be done with hook)
+  const sanitizedSubject = client.replaceNewLinesAndSanitize(subject)
+  const sanitizedBody = client.replaceNewLinesAndSanitize(body)
+  if (!sanitizedSubject || !sanitizedBody) {
+    throw new TemplateError('Message template is empty!')
+  }
   const updatedTemplate = await upsertEmailTemplate({
-    subject: client.replaceNewLinesAndSanitize(subject),
-    body: client.replaceNewLinesAndSanitize(body),
+    subject: sanitizedSubject,
+    body: sanitizedBody,
     replyTo,
     campaignId,
     from,
--- a/backend/src/sms/services/sms-template.service.ts
+++ b/backend/src/sms/services/sms-template.service.ts
@@ -3,7 +3,11 @@ import { difference, keys } from 'lodash'
 import { isSuperSet } from '@core/utils'
 import { HydrationError } from '@core/errors'
 import { Campaign, Statistic } from '@core/models'
-import { TemplateClient, XSS_SMS_OPTION } from 'postman-templating'
+import {
+  TemplateClient,
+  XSS_SMS_OPTION,
+  TemplateError,
+} from 'postman-templating'
 
 import { SmsTemplate, SmsMessage } from '@sms/models'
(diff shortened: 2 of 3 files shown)


backend/src/sms/services/sms-template.service.ts
```diff
@@ -149,8 +153,17 @@ const storeTemplate = async ({
   campaignId,
   body,
 }: StoreTemplateInput): Promise<StoreTemplateOutput> => {
+  if (!client.stripTags(body).trim()) {
+    throw new TemplateError('Message template cannot be empty!')
```
lamkeewei, line 157: If you enter `<foo></foo>`. The error thrown will still be `Message template cannot be empty`. Might be confusing for user since it's not technically empty.

backend/src/email/services/email-template.service.ts
```diff
@@ -161,10 +165,20 @@ const storeTemplate = async ({
   replyTo,
   from,
 }: StoreTemplateInput): Promise<StoreTemplateOutput> => {
+  if (!client.stripTags(subject).trim() || !client.stripTags(body).trim()) {
+    throw new TemplateError('Message subject or template cannot be empty!')
+  }
   // extract params from template, save to db (this will be done with hook)
+  const sanitizedSubject = client.replaceNewLinesAndSanitize(subject)
+  const sanitizedBody = client.replaceNewLinesAndSanitize(body)
+  if (!sanitizedSubject || !sanitizedBody) {
```
lamkeewei, line 174: Hmm...what is the edge case that this second check is checking for? Given that the first step has already stripped all tags and checked if it is empty, wouldn't the second check be a subset of the first step?

miazima: I believe you are right, this was the first check I implemented, and then I refined it to add a check to reject templates which may have valid tags but are essentially empty. The second implementation would encapsulate the logic of the former one. Will remove!

backend/src/sms/services/sms-template.service.ts
miazima: Yep was wondering how to better the error to be more user-friendly. How about `Message template cannot be empty or only contain HTML tags with empty content`.

lamkeewei: I think that works!

lamkeewei:
Perhaps we could also add some unit tests for `stripTags` to cover the different cases (invalid tags, only whitespaces, etc) and ensure that they indeed return an empty string after escaping.

lamkeewei (approved):
lgtm!

frontend/src/components/dashboard/create/sms/SMSTemplate.tsx
```diff
@@ -128,7 +128,7 @@ const SMSTemplate = ({
         <p className={styles.characterCount}>{body.length} characters</p>
       </StepSection>
 
-      <NextButton disabled={!body} onClick={handleSaveTemplate} />
```
jeantanzj, line 131: What if the user actually wanted new lines? :/

jeantanzj:
Currently, we sanitize after the body has been replaced with the parameters because parameters might contain html.
Consequently, do you also want to account for messages that appear empty because the keyword is replaced with a html tag?

To reproduce:
1. Create a template with the body `{{variable}}`
2. Upload a file containing the column `variable`, and a row with the value `<br>` under this column
3. The message will appear to be empty

miazima:
While solving the initial issue of not handling sanitised templates which are empty, my idea was to expand the solution to disable saving empty templates of any kind to prevent such issues and also from the viewpoint that empty templates were not productive. I had considered a template as empty when it was empty after being sanitised, or if it is entirely made up of whitespaces. However, there were several issues with this implementation that were highlighted, 

1. It was not sufficient to check if the template itself was empty, the hydrated template maybe empty too (as described by jean in the previous comment)
2. If the template was to be made up of an image tag which has no content e.g. `</img>`, it was still considered empty - this would be an incorrect conclusion
3. There may be other edge cases similar to above which we may fail to consider

Considering the above issues, it would be difficult to come up with a list of exhaustive edge cases and hence we decided that stringent checks for empty templates was not a feasible solution. We shall leave it up to user discretion to handle empty templates.

Going back to the original [issue](https://github.com/opengovsg/postmangovsg/issues/792) which stemmed from a template consisting of only invalid tags e.g. `<foo></foo>`, the most straightforward solution would be to handle the error thrown by the upload process and show a user friendly error message to the user. However if the user refreshes the page, their sanitised template is now empty and hence they are brought back to Step 1 of creating their template again - this would be confusing for users.

For a more intuitive user experience, I believe the best solution going forward - empty template errors should be raised when users save their template. This would make it clear that their template was invalid and hence they are not allowed to proceed to the next step.

PR #814, opened 2020-11-02
build(ci): use lockfile-lint to inspect lockfiles

LoneRifle:
## Problem

Lockfiles do not get properly inspected at PRs, so it might be trivial to sneak something
past the PR that we may not want

## Solution

 use lockfile-lint to inspect lockfiles

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,7 @@ jobs:
         - npm ci
       script:
         - npm run lint-no-fix
-        - npx lockfile-lint --type npm --path frontend/package-lock.json --validate-https --allowed-hosts npm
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -17,6 +17,8 @@ jobs:
         - npm ci
       script:
         - npm run lint-no-fix
+        - npx lockfile-lint --type npm --path frontend/package-lock.json --validate-https --allowed-hosts npm
```
jeantanzj, line 20: Has to be run in each job or in the root job.

jeantanzj (changes requested):

PR #815, opened 2020-11-03
chore: Add Google Analytics tracking for delivery report download

lamkeewei:
## Problem

Currently, we don't have visibility into whether and when users are downloading the delivery reports. Having this information will help us make a more informed decision on when we can carry out campaign redaction.

## Solution

**Features**:

- Add Google Analytics event to track when user click the download delivery report button
- Google Analytics event captures the following:
   - Action: `Download delivery report`
   - Label: Campaign channel type
   - Value: Days since campaign was sent out
- From GA events dashboard, we can determine the number of downloads and the average campaign age for when the download was triggered 

**Improvements**
- Added event labels prefix to differentiate outbound links to `guide.postman.gov.sg` from navbar and empty dashboard

## Tests
- Trigger a download of the delivery report. Verify that an event is sent to GA through the Chrome developer console


frontend/src/components/landing/nav-bar/Navbar.tsx
```diff
@@ -38,7 +38,7 @@ const NavBar = () => {
           </OutboundLink>
           <OutboundLink
             className={styles.link}
-            eventLabel={i18n._(LINKS.guideUrl)}
+            eventLabel={`Navbar / ${i18n._(LINKS.guideUrl)}`}
```
miazima, line 41: There is another NavBar component at `components/common/nav-bar`, need to update the guide link in that too

miazima (approved):
Demo mode removed the guide button from the empty dashboard screen which resulted in a conflict, other than that lgtm!

PR #828, opened 2020-11-08
Fix to restore saving of new telegram credentials 

jeantanzj:
## Problem

New telegram credentials could not be saved due to incorrect refactoring

## Solution

Restore the assignment of credentialName


lamkeewei (approved):
Tested the following flows and works as expected:
1. Save new credentials from settings
2. Save new credentials during campaign creation flow
3. Use existing credentials to send campaign

PR #833, opened 2020-11-10
fix: await on creation of campaign

jeantanzj:
## Problem

User managed to create a demo campaign ( demo_messages_limit is not null) but their numDemosSms count was not decremented. 

## Solution

Await on the transaction to create a campaign.

### DIFF
--- a/backend/src/core/services/campaign.service.ts
+++ b/backend/src/core/services/campaign.service.ts
@@ -69,7 +69,7 @@ const createDemoCampaign = async ({
 /**
  * Helper method to create a campaign
  */
-const createCampaign = ({
+const createCampaign = async ({
   name,
   type,
   userId,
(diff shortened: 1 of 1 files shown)


lamkeewei:
Can't seem to repro the error. Was thinking if the issue is indeed caused by not awaiting on the promises, then `campaign` [here](https://github.com/opengovsg/postmangovsg/blob/44171fee53aa0480924c0b97f03f3e6e04fd715b/backend/src/core/middlewares/campaign.middleware.ts#L63) should be `undefined` or some sort of error will be thrown when we try to access an attribute on `campaign` further down. 

```js
const campaign = await CampaignService.createCampaign({
    name,
    type,
    userId,
    protect,
    demoMessageLimit,
})

if (!campaign) {
    return res.status(400).json({
        message: `Unable to create campaign with these parameters`,
    })
}

// ...

return res.status(201).json({
      id: campaign.id,
      name: campaign.name,
      created_at: campaign.createdAt,
      type: campaign.type,
      protect: campaign.protect,
      demo_message_limit: campaign.demoMessageLimit,
})

```

But the fact that we returned `201 Created` would indicate that the promises were being awaited/resolved successfully (?). More broadly even if we don't await, promises should still run though. :/ 

For example:
```js
new Promise((resolve) => {
  setTimeout(() => {
    console.log('Here 2')
    resolve()
  }, 5000)
})

console.log('Here')
```

Will only exit when `Here 2` is printed.

jeantanzj:
Yep, tried again to reproduce, but couldn't. The awaits in the transaction are needed, though, so will still merge this.

PR #840, opened 2020-11-12
Fix: campaign progress not updated when campaign properties were updated

### LINKED ISSUE #838: Previous steps in progress pane are disabled
**Describe the bug**
Completed steps after 'Create message' are disabled when progressing to next steps.

**To Reproduce**

1. Create a new sms campaign
2. Save a template, upload a file and click next
3. Now you should be at the insert credentials step
4. **Unexpected behaviour**: Previous step 'Upload recipients' is now disabled
5. Insert credentials and proceed to 'Preview and send' step, then navigate to 'Create message' step via the progress pane
6. **Unexpected behaviour**: All the previously completed steps are disabled

image
image

**Expected behavior**
All previously completed steps in the progress pane should be accessible and not disabled.

miazima:
## Problem

Previously completed steps in progress pane were disabled. 

Closes #838

## Solution

Campaign progress was not being updated when campaign state was updated. The progress pane steps render enabled/disabled state based on the campaign progress property.

**Bug Fixes**:

- Create a function to handle updates to the campaign state in the campaign context provider
- Call the campaign `setProgress` function to update the progress value for the campaign state
- Implement `useCallback` for `updateCampaign` function to prevent re-render due to function being re-initialised for every render (**Looking into if this is the best way to do this**)

## Tests

1. Create a new sms campaign
2. Save a template, upload a file and click next
3. Now you should be at the insert credentials step
4. Previous step 'Upload recipients' is enabled
5. Insert credentials and proceed to 'Preview and send' step, then navigate to 'Create message' step via the progress pane
6. All the previously completed steps are enabled

### DIFF
--- a/frontend/src/contexts/campaign.context.tsx
+++ b/frontend/src/contexts/campaign.context.tsx
@@ -10,7 +10,9 @@ import { SMSCampaign, EmailCampaign, TelegramCampaign, Campaign } from 'classes'
 
 interface ContextProps {
   campaign: SMSCampaign | EmailCampaign | TelegramCampaign
-  updateCampaign: any
+  updateCampaign: (
+    changes: Partial<SMSCampaign | EmailCampaign | TelegramCampaign>
+  ) => void
   setCampaign: Dispatch<
     SetStateAction<SMSCampaign | EmailCampaign | TelegramCampaign>
   >
--- a/frontend/src/contexts/campaign.context.tsx
+++ b/frontend/src/contexts/campaign.context.tsx
@@ -46,6 +46,9 @@ const CampaignContextProvider = ({
     <CampaignContext.Provider
       value={{
         campaign,
+        // updateCampaign is re-created for every re-render of CampaignContextProvider
+        // Memoise updateCampaign to prevent unnecessary renders in subscribers when
+        // defined as dependencies in useEffect
         updateCampaign: useCallback(updateCampaign, []),
         setCampaign,
       }}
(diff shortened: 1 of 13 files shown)


miazima:
For reference, this was how campaign state was updated previously before introduction of campaign context
https://github.com/opengovsg/postmangovsg/blob/44171fee53aa0480924c0b97f03f3e6e04fd715b/frontend/src/components/dashboard/create/sms/SMSCreate.tsx#L39-L51

frontend/src/contexts/campaign.context.tsx
```diff
@@ -19,10 +29,27 @@ const CampaignContextProvider = ({
...
+  }
+
   return (
     <CampaignContext.Provider
       value={{
         campaign,
+        // updateCampaign is re-created for every re-render of CampaignContextProvider
+        // Memoise updateCampaign to prevent unnecessary renders in subscribers when
+        // defined as dependencies in useEffect
+        updateCampaign: useCallback(updateCampaign, []),
```
miazima, line 52: When `CampaignContextProvider` re-renders due to changes in campaign state, the `updateCampaign` function is re-created. This affects hooks which have `updateCampaign` as a dependency.
```
useEffect(() => {
    updateCampaign({ isCsvProcessing, csvFilename, numRecipients })
}, [isCsvProcessing, csvFilename, numRecipients, updateCampaign])
```
For example, the useEffect hook will execute again as `updatedCampaign`'s value changed. This executes `updateCampaign` within the hook and the context campaign state is updated, `CampaignContextProvider` re-renders .... This would result in infinite re-renders. Hence we are using `useCallback` here to memoize the `updateCampaign` function to prevent unnecessary side effects.

This article helps to elaborate on this concept - https://reacttraining.com/blog/when-to-use-functions-in-hooks-dependency-array/, particularly the section on `useCallback`

jeantanzj (approved):

PR #841, opened 2020-11-12
chore: Add Cronitor monitoring for scheduled jobs

lamkeewei:
## Problem

Existing CloudWatch alerts do not account for the event where the lambdas do not run (e.g. deleted accidentally). 

## Solution

**Features**:

- Setup and added support for Cronitor monitoring to `database-backup` and `unsubscribe-digest`.
   - Cronitor is only enabled if a Cronitor code is passed. For testing, I've setup Cronitor monitors for staging. Will remove them once we've deployed to production to prevent using up an extra slot on the shared Cronitor. 

## Tests

**Successful run**
- Invoke both lambdas locally with the staging Cronitor codes set in `.env`. The codes can be retrieved from Cronitor dashboard.
- Login to Cronitor dashboard and check that runs were recorded

**Failed run**
- Manually inject error into both lambdas
- Alert should be sent to `postman-alerts` Slack channel
- Remove injected error and rerun lambdas
- Alert resolved message should be sent to `postman-alerts` Slack channel

## Deploy Notes

Added Cronitor codes for production and staging in Travis:
- `STAGING_UNSUBSCRIBE_CRONITOR_CODE`
- `PRODUCTION_UNSUBSCRIBE_CRONITOR_CODE`
- `STAGING_BACKUP_CRONITOR_CODE`
- `PRODUCTION_BACKUP_CRONITOR_CODE`

^ I've already set the appropriate values in Travis.

**New dependencies**:

- `axios` : Added to `database-backup` lambda to ping Cronitor


serverless/unsubscribe-digest/src/utils/cronitor.ts
```diff
@@ -0,0 +1,54 @@
...
+import { Logger } from './logger'
+
+const logger = new Logger('db-backup')
+
+const CRONITOR_PING_URL = `https://cronitor.link`
+const CRONITOR_CODE = config.get('cronitor.code')
+
+const buildUrl = (action: string, message?: string): string => {
+  const url = `${CRONITOR_PING_URL}/${CRONITOR_CODE}/${action}`
+  return message ? `${url}?msg=${message}` : url
```
jeantanzj, line 13: if you're really going to do this, best use the url package? otherwise you have to sanitize/url encode `message`.

serverless/unsubscribe-digest/src/utils/cronitor.ts
```diff
@@ -0,0 +1,54 @@
+import axios from 'axios'
+import { Cronitor } from '../interface'
+import config from '../config'
+import { Logger } from './logger'
+
+const logger = new Logger('db-backup')
+
+const CRONITOR_PING_URL = `https://cronitor.link`
```
jeantanzj, line 8: why did you choose to set two variables CRONITOR_PING_URL and CRONITOR_CODE instead of just one CRONITOR_URL? the two variables are always concatenated anyway

jeantanzj:
What happens if the job doesnt run? Does cronitor report it?

serverless/unsubscribe-digest/src/utils/cronitor.ts
lamkeewei: Makes sense. Will use one variable.

lamkeewei:
> What happens if the job doesnt run? Does cronitor report it?

Yep. The cron expression is configured on Cronitor's dashboard. So if Cronitor doesn't receive a `complete` ping within the expected timeframe (daily for backup, weekly for unsub), it will raise an alert.

jeantanzj (approved):

Issue #805, opened 2020-10-28
chore: replace /new-credentials with /new-credentials/v2


jeantanzj:
Following up on https://github.com/opengovsg/postmangovsg/pull/772
for SMS and Telegram

PR #847, opened 2020-11-18
Deprecate new-credentials endpoint

miazima:
## Problem

Part 1 of #805

## Solution

- Update `/new-credentials` endpoint to accept optional `label` body parameter
- Redirect `/new-credentials/v2` to `/new-credentials` for SMS and Telegram
- Update frontend to use the updated `/new-credentials` endpoint

## Tests

- [ ] Add new credentials with label in create campaign flow for SMS and Telegram and check that credential has been saved with corresponding label


jeantanzj (changes requested):
See #849

jeantanzj (approved):

PR #852, opened 2020-11-24
Add channels description section in landing page

### LINKED ISSUE #347: [Telegram] [Frontend] Landing
Update landing copy to tout the benefits of telegram

miazima:
## Problem

Add a section about Postman's channels and their descriptions

Closes #347

## Before & After Screenshots

image

image


frontend/src/components/landing/Landing.tsx
```diff
@@ -207,6 +208,44 @@ const Landing = () => {
...
+
+            <h4>SMS</h4>
+            <p>
+              Move away from calling. Send transactional SMS easily through our
+              Twilio integration. Remind citizens to come for their
+              appointments. The possibilities are endless.
+            </p>
+
+            <PrimaryButton className={styles.button}>
+              Learn More <i className="bx bx-right-arrow-alt"></i>
```
lamkeewei, line 243: @NatMaeTan @sarahsalim what is this the correct hover state for the "Learn more" button for the channels section of the landing page? Looks a little hard to read.

image

sarahsalim:
Oh sorry, I think something weird happened at Figma and the arrow became blue :D lol.

The button should be outlined white with white text and icon.
Screenshot 2020-12-02 at 9 24 16 AM

These are the different states of a button on dark background. We've updated the Design System to include this set.
Screenshot 2020-12-02 at 10 00 00 AM

frontend/src/styles/base/_colours.scss
```diff
@@ -1,5 +1,6 @@
 // Colours
 $white: #ffffff;
+$transparent-white: rgba(255, 255, 255, 0.2);
```
lamkeewei, line 3: Was thinking for colors with transparency, we can standardise and use `adjust-color($white, $alpha: -0.8)` instead of creating a new variable. A couple of things this will help with are: 
1. This keeps the variables aligned with the design system
2. For other colors, it's easier to remain consistent since we can reuse the variables instead of having to change to rgba values. E.g. `adjust-color($primary-light, $alpha: -0.7)`

One thing I observed while refactoring the colors out was that there could be multiple variants of "transparent-blue", each with a different alpha value. What do you think?

miazima: I agree, I was not aware of this sass function! I think this definitely prevents the spawn of multiple `$transparent` valued variables and we can stick to a pre-defined set of colour variables. 

However so far our usage of `adjust-color(color, change in alpha)` has been limited to the alpha param, which is equivalent to using `rgba(color, alpha)`. Personally I find using `rgba()` easier to read and understand as it uses the desired alpha value vs using `adjust-color` which we have back-calculate the alpha value and define the difference. Let me know if you feel otherwise?

lamkeewei: Thats true. And given that it's quite easy to get the rgb values from Figma anyway. This is fine!

lamkeewei (approved):
lgtm!

PR #853, opened 2020-11-24
feat: Verify SES notification settings

### LINKED ISSUE #777: [Custom Domain] Verify that sns topic is hooked correctly for a new from address
When adding a new from address,
1) add the from address to verified address on ses
2) get IT team to add DNS records
3) make sure the verified address's notifications go to an sns topic that calls the callback endpoints, with original headers enabled, and email forwarding disabled. 

We've checked for 1) and 2), but I'd missed out on the 3rd check.

lamkeewei:
## Problem

Closes #777 

## Solution

**Features**:

- Retrieve and verify that the notification settings are configured correctly:
   - All notifications should be directed to configured SNS topic
   - Inclusion of original headers should be enabled for all notification types
   - Disable forwarding of notification to email

## Tests
_I've setup Jean's domain for test in eu-central-1_

**Verify custom domain**
- Given the following: 
   - Email is verified
   - DKIM records are set
   - Notification settings are wrong
- Custom domain verification should fail

**Previously valid settings**
- If previously valid settings are now invalid, the following scenarios should result in an error:
   - Template cannot not be saved
   - If template is saved, send test email should raise an error
   - If template is saved and test email was already sent, triggering campaign send should raise an error

**Send with default mail from**
- Create campaign with default mail from instead of custom mail from
- Verify that campaign can still be sent out

## Deploy Notes

- `SES_NOTIFICATION_TOPIC` : SNS topic to direct all notifications to

**IAM Permissions**
- Ensure that `getIdentityNotificationAttributes` is granted for beanstalk instance role

## TODOs
- [x] Test with a custom domain

### DIFF
--- a/backend/src/email/middlewares/email.middleware.ts
+++ b/backend/src/email/middlewares/email.middleware.ts
@@ -1,5 +1,9 @@
 import { Request, Response, NextFunction } from 'express'
-import { EmailService, CustomDomainService } from '@email/services'
+import {
+  EmailService,
+  EmailTemplateService,
+  CustomDomainService,
+} from '@email/services'
 import { parseFromAddress } from '@core/utils/from-address'
 import { AuthService } from '@core/services'
 import config from '@core/config'
@@ -258,6 +262,40 @@ const getCustomFromAddress = async (
   return res.status(200).json({ from: result })
 }
 
+/*
+ * Get email template's From address for specific campaign
+ */
+const getCampaignFromAddress = async (
+  req: Request,
+  res: Response,
+  next: NextFunction
+): Promise<Response | void> => {
+  try {
+    const { campaignId } = req.params
+    const template = await EmailTemplateService.getFilledTemplate(+campaignId)
+
+    if (!template) {
+      const message = `No template associated with campaign ${campaignId}`
+      logger.error({
+        message,
+        campaignId,
+        action: 'getCampaignFromAddress',
+      })
+
+      return res.status(400).json({ message })
+    }
+
+    /* eslint-disable @typescript-eslint/no-non-null-assertion*/
+    const { name, fromAddress } = parseFromAddress(template?.from!)
+    res.locals.fromName = name
+    res.locals.from = fromAddress
+
+    next()
+  } catch (err) {
+    next(err)
+  }
+}
+
 /**
  * Sends a test email from the specified from address
  */
@@ -332,6 +370,7 @@ export const EmailMiddleware = {
   verifyFromAddress,
   storeFromAddress,
   getCustomFromAddress,
+  getCampaignFromAddress,
   existsFromAddress,
   isFromAddressAccepted,
   sendValidationMessage,
--- a/backend/src/email/middlewares/email.middleware.ts
+++ b/backend/src/email/middlewares/email.middleware.ts
@@ -161,8 +161,10 @@ const existsFromAddress = async (
 ): Promise<Response | void> => {
   const { from } = req.body
   const defaultEmail = config.get('mailFrom')
-  if (from === defaultEmail) return next()
-  const { fromName, from: fromAddress } = res.locals
+  const { fromName, from: fromAddress, templateFrom } = res.locals
+  // Skip check if either the provided from address or the template's from address is the same
+  // as the configured default mail from.
+  if (from === defaultEmail || templateFrom === defaultEmail) return next()
 
   try {
     const exists = await CustomDomainService.existsFromAddress(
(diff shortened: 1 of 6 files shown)


backend/src/email/services/custom-domain.service.ts
```diff
@@ -74,6 +74,76 @@ const verifyEmailWithAWS = async (email: string): Promise<Array<string>> => {
...
+      throw new Error(
+        'Original email headers are not included in notifications. Please include them by ' +
+          'enabling it in SES.'
+      )
+    }
+  } catch (err) {
+    logger.error({
+      message: 'Invalid SES notification settings',
+      email,
+      error: err,
```
jeantanzj, line 138: `err.message` ? info about the error is not being logged
```
{"level":"error","message":"Invalid SES notification settings","metadata":{"ip":"xxxxxxxx","userId":"xx","label":"email/services/custom-domain.service","email":"xxxxxx@xxxxx","error":{},"action":"verifyNotificationSettings","timestamp":"2020-12-15T09:41:22.609Z"}}
```

frontend/src/services/campaign.service.ts
```diff
@@ -158,8 +158,15 @@ export async function sendCampaign(
...
 ): Promise<void> {
-  const body = sendRate ? { rate: sendRate } : null
-  await axios.post(`/campaign/${campaignId}/send`, body)
+  try {
+    const body = sendRate ? { rate: sendRate } : null
+    await axios.post(`/campaign/${campaignId}/send`, body)
+  } catch (err) {
+    errorHandler(
+      err,
+      'An error occured while sending the campaign. Please try again.'
```
jeantanzj, line 167: occurred

frontend/src/services/campaign.service.ts
```diff
@@ -191,3 +198,11 @@ export async function exportCampaignStats(
     return campaignRecipients
   })
 }
+
+function errorHandler(e: AxiosError, defaultMsg: string): never {
+  console.error(e)
+  if (e.response && e.response.data && e.response.data.message) {
```
jeantanzj, line 204: I think we had this discussion before on not displaying too much information about errors to the user.

jeantanzj: Screenshot 2020-12-15 at 5 54 04 PM
was this expected?

backend/src/email/routes/email-campaign.routes.ts
```diff
@@ -608,6 +611,9 @@ router.post(
   '/send',
   celebrate(sendCampaignValidator),
   CampaignMiddleware.canEditCampaign,
+  EmailMiddleware.getCampaignFromAddress,
```
jeantanzj, line 614: if we're checking it here right before send, do we still need to check the address's validity at each stage?

jeantanzj (approved):
The checks worked - but I had to add IAM permissions for getIdentityNotificationAttributes. Left some thoughts. If it's behaving as you intended,  then I'm fine with it.

frontend/src/services/campaign.service.ts
lamkeewei: Yep. I added this because the default error message `An error occured while sending the campaign. Please try again.` is not useful if the custom domain is not configured. 

Instead, I thought that `This From Address cannot be used to send emails. Select another email address to send from, or contact us to investigate.` would be more helpful if it is an error with the custom domain. 

I think this is the same message shown in the custom domain modal. After catching and logging the specific error, this generic error is re-thrown in `verifyNotificationSettings`.

backend/src/email/routes/email-campaign.routes.ts
lamkeewei: Technically we don't need to. But I added it so that it checks at the same places as before. Main flows I tested were same as #728

frontend/src/services/campaign.service.ts
lamkeewei: ^ shouldn't be just showing status code 400. `This From Address cannot be used...` should be shown. Let me double check this.

lamkeewei: Just verified on staging again. This is the error message shown. I think staging frontend wasn't running this branch yesterday.

image

backend/src/email/middlewares/email.middleware.ts
```diff
@@ -161,8 +161,10 @@ const existsFromAddress = async (
 ): Promise<Response | void> => {
   const { from } = req.body
   const defaultEmail = config.get('mailFrom')
-  if (from === defaultEmail) return next()
-  const { fromName, from: fromAddress } = res.locals
+  const { fromName, from: fromAddress, templateFrom } = res.locals
+  // Skip check if either the provided from address or the template's from address is the same
+  // as the configured default mail from.
+  if (from === defaultEmail || templateFrom === defaultEmail) return next()
```
lamkeewei, line 167: Discovered a bug while re-testing. Campaigns using the default from address were not able to be sent. The checks were not skipped as `from` will be `null` for `POST /credentials` and `POST /send`. 

`templateFrom` has to be added because `defaultFrom` takes the form of `Postman.gov.sg <donotreply@mail.postman.gov.sg>`. 

@jeantanzj do you think this check can be better handled?

jeantanzj:
To be closed when we move to using config sets instead of setting SNS notification topics

PR #856, opened 2020-11-24
feature: copy campaign

jeantanzj:
## Problem

Users want to be able to reuse the template from a campaign that they've previously created.

## Solution

Enable the user to copy a campaign and its template. 

Screenshot 2020-11-24 at 7 15 03 PM
Screenshot 2020-11-24 at 7 15 25 PM


backend/src/email/services/email.service.ts
```diff
@@ -296,6 +296,44 @@ const uploadProtectedCompleteOnChunk = ({
...
+const duplicateCampaign = async ({
+  campaignId,
+  name,
+}: {
+  campaignId: number
+  name: string
+}): Promise<Campaign | void> => {
+  const campaign = await Campaign.findByPk(campaignId)
+  if (campaign) {
+    const duplicate = await CampaignService.createCampaign({
```
lamkeewei, line 308: Perhaps should wrap `createCampaign` and `storeTemplate` in a transaction? Prevent the situation where a new campaign is created by the template did not get copied over.

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -141,7 +142,34 @@ const Campaigns = () => {
           />
         )
       },
-      width: 'md center',
+      width: 'md',
```
lamkeewei, line 145: Looking at the design master, I think should be `md center` so that the buttons are more evenly spaced out.

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -141,7 +142,34 @@ const Campaigns = () => {
...
+                <DuplicateCampaignModal campaign={campaign} />
+              )
+            }}
+          >
+            <i className={cx('bx bx-duplicate', styles.icon)}></i>{' '}
+            <span>Duplicate</span>
+          </div>
+        )
+      },
+      width: 'md',
```
lamkeewei, line 172: Same here, `md center`

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -141,7 +142,34 @@ const Campaigns = () => {
...
+    },
+    {
+      name: '',
+      render: (campaign: Campaign) => {
+        if (
+          campaign.demoMessageLimit &&
+          ((numDemosSms === 0 && campaign.type === ChannelType.SMS) ||
+            (numDemosTelegram === 0 && campaign.type === ChannelType.Telegram))
+        ) {
+          return
```
lamkeewei, line 155: @NatMaeTan what is the behaviour/flow for "Duplicate" if the user tries to duplicate a demo campaign but does not have anymore demos left? 

I can imagine a case where a user tried out a few demo campaigns and finally decide to send it out for real. But then they find themselves unable to reuse/duplicate the demo campaign they've tested with because they no longer have anymore demos left.

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -126,7 +127,7 @@ const Campaigns = () => {
       width: 'xs center',
     },
     {
-      name: 'Export',
+      name: '',
```
lamkeewei, line 130: `name` is used as the key for `td` and `th` in `renderCampaignList` and `renderRow`. Since they're both empty strings now, React will complain about duplicated keys. Perhaps use the index value as the key instead.

image

backend/src/email/services/email.service.ts
jeantanzj: Haha i considered it but the transaction will now have to be unmanaged

jeantanzj: Let me think about it further

frontend/src/components/dashboard/campaigns/Campaigns.tsx
jeantanzj: I tried that but the spacing looked funny. This is md center for both
Screenshot 2020-11-25 at 1 25 07 PM

jeantanzj: Note the space on the right between delivery report and the duplicate icon

frontend/src/components/dashboard/campaigns/Campaigns.tsx
jeantanzj: The key is already the index.  Thanks for raising, will look into it
```
<tbody>{campaignsDisplayed.map(renderRow)}</tbody>
```
```
  function renderRow(campaign: Campaign, key: number) {
    return (
      <tr key={key} onClick={() => history.push(`/campaigns/${campaign.id}`)}>
...
```

jeantanzj: oh nvm i was debugging the wrong part, youre right

frontend/src/components/dashboard/campaigns/Campaigns.tsx
jeantanzj: Screenshot 2020-11-25 at 1 33 54 PM
how about md center and md?

NatMaeTan: The "Delivery report" button can be shorted to "Report" to make space, as it's pretty clear what the report is about. 

This can also give more space to column spacing, could we try making the 'name' column wider and pushing the other columns over?

For the text button icons, we created a new issue [here](https://github.com/opengovsg/postmangovsg/issues/859)

NatMaeTan:
Screenshot 2020-11-24 at 7 15 25 PM

For this modal, the disabled tall buttons of the other channels should be in the outline disabled style, not solid disabled.
Screenshot 2020-11-25 at 4 01 23 PM

Also, the "Password protected" text should be disabled with the checkbox. It's a bit weird if only the checkbox is disabled and not the text with it, since user's can't uncheck it either way.

frontend/src/components/dashboard/campaigns/Campaigns.tsx
NatMaeTan: Agree that demo campaigns should have the button omitted. We shouldn't encourage duplicate of demo campaigns anyways!

lamkeewei (approved):
Test following scenarios and works as expected:
- Duplicate normal campaigns (SMS, Telegram, Email and Protected Email)
- Duplicate demo campaigns
- Attempt to duplicate demo campaigns without trials left through API (fails as expected)
- Duplicate campaign without template
- Duplicate campaign with custom from

PR #862, opened 2020-11-26
Prevent setting state after component is dismounted

### LINKED ISSUE #846: Resolve warnings for components rendered when saving draft
Screenshot 2020-11-17 at 12 29 10 AM

To reproduce:
1. Be on the create template page
2. Click finish later
3. Click save draft

### LINKED ISSUE #858: SMS preview skipped
**Describe the bug**
For SMS campaigns, after uploading of a CSV is completed, the frontend automatically progresses to the credentials step. The expected behaviour is to display the message preview to the user before advancing to the next step after clicking "Next". 

This only occurs for SMS campaigns.

Discovered this while testing #809. Filing as separate issue as I think it's unrelated. 

**To Reproduce**
1. Create a new SMS campaign
2. Enter a valid SMS template
3. Upload CSV file
4. Page automatically advances to enter credential step

miazima:
## Problem

**Prevent setting state after component is dismounted**
There are many instances where `useEffect` sets components state after running asynchronous queries. When the component is dismounted before these execute finish, they result in react trying to update the state of an unmounted component. This throws a warning that this may indicate a potential memory leak in the app.

Closes #846 
Closes #858 

## Solution

Ran through different user flows to check for warnings, let me know if there are some which have not been addressed!

**Bug Fixes**:

- When handling finish later - save draft callbacks, remove use of `handleSaveTemplate()` which has redundant setState calls and instead use `saveTemplate` to directly call endpoint
- clear timers set in `useEffect` hooks for `ApiKey` and `LoginInput` component
- In `SMSValidationInput` and `EmailValidationInput`, switch order of setState calls to set internal state before calling handler from parent component
- In `Landing` component, check `isAuthenticated` is false before retrieving stats, as page gets redirected when it is true
- Create custom hook tp encapsulate `isMounted` state - used to check if component is mounted/dismounted
- Disable previous button when csv upload is in progress for all channels
- Remove `activeStep`'s incorrect dependency on progress in `SMSCreate`
- Disable progress pane for Telegram when csv upload is in progress

## Tests

- [x] Check for react warnings regarding setting state after component has been dismounted
- [x] In SMS channel, after file upload, preview should be displayed and page should not progress to next step immediately
- [x] Check that progress pane and previous, next buttons are disabled when csv upload is in progress for all channels

### DIFF
--- a/frontend/src/components/dashboard/create/email/EmailRecipients.tsx
+++ b/frontend/src/components/dashboard/create/email/EmailRecipients.tsx
@@ -1,7 +1,6 @@
 import React, {
   useState,
   useEffect,
-  useRef,
   useContext,
   Dispatch,
   SetStateAction,
@@ -29,6 +28,7 @@ import {
 } from 'components/common'
 import { EmailPreview, EmailProgress } from 'classes'
 import { sendTiming } from 'services/ga.service'
+import useIsMounted from 'components/custom-hooks/use-is-mounted'
 
 import styles from '../Create.module.scss'
 
@@ -63,13 +63,7 @@ const EmailRecipients = ({
   const [preview, setPreview] = useState({} as EmailPreview)
   const { id: campaignId } = useParams()
   const { csvFilename, numRecipients = 0 } = csvInfo
-  const isMounted = useRef(true)
-
-  useEffect(() => {
-    return () => {
-      isMounted.current = false
-    }
-  }, [])
+  const isMounted = useIsMounted()
 
   // Poll csv status
   useEffect(() => {
@@ -93,6 +87,7 @@ const EmailRecipients = ({
         if (preview) {
           setPreview(preview as EmailPreview)
         }
+
         if (isCsvProcessing) {
           timeoutId = setTimeout(pollStatus, 2000)
         }
@@ -106,7 +101,7 @@ const EmailRecipients = ({
     pollStatus()
 
     return () => clearTimeout(timeoutId)
-  }, [campaignId, csvFilename, forceReset, isCsvProcessing])
+  }, [campaignId, csvFilename, forceReset, isCsvProcessing, isMounted])
 
   // If campaign properties change, bubble up to root campaign object
   useEffect(() => {
@@ -225,7 +220,10 @@ const EmailRecipients = ({
             disabled={!numRecipients || isCsvProcessing}
             onClick={() => setActiveStep((s) => s + 1)}
           />
-          <TextButton onClick={() => setActiveStep((s) => s - 1)}>
+          <TextButton
+            disabled={isCsvProcessing}
+            onClick={() => setActiveStep((s) => s - 1)}
+          >
             Previous
           </TextButton>
         </ButtonGroup>
--- a/frontend/src/components/dashboard/create/email/EmailTemplate.tsx
+++ b/frontend/src/components/dashboard/create/email/EmailTemplate.tsx
@@ -105,7 +105,7 @@ const EmailTemplate = ({
       <SaveDraftModal
         saveable
         onSave={async () => {
-          if (subject && body && from) {
+          if (campaignId && subject && body && from) {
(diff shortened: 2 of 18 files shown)


frontend/src/components/dashboard/create/sms/SMSValidationInput.tsx
```diff
@@ -16,8 +16,8 @@ const EmailValidationInput = ({
   }
 
   async function onClickHandler() {
-    await onClick(recipient)
     setRecipient('')
```
lamkeewei, line 19: Instead of switching the order, I think it's possible to just get rid of the `setRecipient('')` call instead. Since if validation succeeds, it will be unmounted when the next step is shown. Even if the user chooses to enter new credentials again, the component will be reinitialised with an empty value, negating a need to manually reset it's value.

This way, we can still preserve the value in the input when sending. Otherwise if there's an error, the user might not be able to see what he/she had input in the first place. 

image

frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -105,16 +105,24 @@ const EmailTemplate = ({
       <SaveDraftModal
         saveable
         onSave={async () => {
-          if (subject && body && from) {
-            await handleSaveTemplate(true)
+          if (campaignId && subject && body && from) {
+            await saveTemplate(+campaignId, subject, body, replyTo, from)
```
lamkeewei, line 109: One other side effect of `handleSaveTemplate` is that it calls `setErrorMsg`. This is such that when users click "Edit draft" (this is not working at the moment) when they faced an error during the finish later flow, they can see the error message on the template page. Not sure if this behaviour should be preserved. 

image

frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -105,16 +105,24 @@ const EmailTemplate = ({
...
     )
     return () => {
       setFinishLaterContent(null)
     }
-  }, [body, subject, from, handleSaveTemplate, setFinishLaterContent])
+  }, [
+    body,
+    subject,
+    from,
+    handleSaveTemplate,
```
lamkeewei, line 121: Remove if we are not using `handleSaveTemplate` anymore.

frontend/src/components/dashboard/create/sms/SMSValidationInput.tsx
miazima: I agree, tried it out and it works as intended

frontend/src/components/dashboard/create/email/EmailTemplate.tsx
miazima: I believe we can retain this behaviour as when `saveTemplate()` fails, the page should not be navigated to the dashboard hence the component should not be dismounted - we can set state on the component without warnings. I have amended it as following, this would throw a warning as of now, but should be resolved with #871 .
https://github.com/opengovsg/postmangovsg/blob/a0a9971cb8e17fb346a13bf7e252b10d841ea9d7/frontend/src/components/dashboard/create/email/EmailTemplate.tsx#L101-L110

frontend/src/components/dashboard/create/email/EmailRecipients.tsx
```diff
@@ -63,13 +63,7 @@ const EmailRecipients = ({
...
   const { id: campaignId } = useParams()
   const { csvFilename, numRecipients = 0 } = csvInfo
-  const isMounted = useRef(true)
-
-  useEffect(() => {
-    return () => {
-      isMounted.current = false
-    }
-  }, [])
+  const isMounted = useIsMounted()
```
lamkeewei, line 66: As discussed, I think we will need to add this for `SMSRecipients` and `TelegramRecipients` also.

lamkeewei (approved):
lgtm! `setState` warnings are resolved.

PR #863, opened 2020-11-26
feat: Make reply-to address compulsory

### LINKED ISSUE #854: Make reply-to a compulsory field
Currently `reply-to` is an optional field. If it's not specified, users cannot reach the agency sender if they have any queries (change recipient, questions about the email content, etc)

Make `reply-to` a compulsory field

lamkeewei:
## Problem

Closes #854 

## Solution

**Features**:
- Set logged in user email address as reply to if not provided

## Tests
- Create a new email campaign
- Leave reply to as empty and save template
- Check that the reply to field is set as the logged in user's email in the database

### DIFF
--- a/frontend/src/components/dashboard/create/email/EmailTemplate.tsx
+++ b/frontend/src/components/dashboard/create/email/EmailTemplate.tsx
@@ -18,6 +18,7 @@ import {
 } from 'components/common'
 import SaveDraftModal from 'components/dashboard/create/save-draft-modal'
 import { FinishLaterModalContext } from 'contexts/finish-later.modal.context'
+import { AuthContext } from 'contexts/auth.context'
 import { CampaignContext } from 'contexts/campaign.context'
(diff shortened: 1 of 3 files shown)


frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -190,9 +192,7 @@ const EmailTemplate = ({
         </div>
 
         <div>
-          <h4 className={styles.replyToHeader}>
-            Replies <em>optional</em>
-          </h4>
+          <h4 className={styles.replyToHeader}>Replies</h4>
           <p>
             All replies will be directed to the email address indicated below
```
jeantanzj, line 197: Then we can change this to: `If left blank, replies will be directed to ${userEmail}`

jeantanzj (approved):
Remember to backfill the db

PR #879, opened 2020-12-07
Fix: edit draft returns user to main dashboard instead of template page

### LINKED ISSUE #871: Edit draft returns user to main dashboard instead of template page
**Describe the bug**
When user clicks "Edit draft", they should be taken back to the template page instead of returning them to the main dashboard.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create an email campaign
2. Enter an invalid value for reply to (non-email)
3. Click "Finish this later"
4. Click "Save draft"
5. Click "Edit draft"

**Expected behaviour**
User should be returned to the template page so that he/she can correct the template and try to save draft again.

**Screenshots**

miazima:
## Problem

After a user enters an invalid template and clicks 'Finish Later' button on Create template screen, the 'Finish Later' modal will open. When clicking 'Save draft', this will then display an error together with the 'Edit Draft' button. When clicking on the 'Edit Draft' button, the page routes to the dashboard view. This is an incorrect behaviour.

Closes #871 

## Solution

**Bug Fixes**:

https://github.com/opengovsg/postmangovsg/blob/347ebfa58894a162b79dfbb380d92c57f7138546/frontend/src/components/dashboard/create/Create.tsx#L38-L44

Previously, we did not `return finishLaterContextHandler() `, hence the page was routed to the dashboard regardless of the outcome of the SaveDraft modal. By returning the handler, when `finishLaterContent` is defined, i.e. it will open SaveDraft modal when clicked, any page routing will be handled within that component.

## Tests

- [x] Edit draft functionality
- Enter an Invalid template `{{test`
- Click on `Finish Later` button followed by `Save draft`
- Modal displays error, click on `Edit draft`, modal should close and current screen should be still at create template 

- [x] Click on 'Skip saving draft' in previous scenario, check that page is routed to the dashboard view
- [x] In password protected campaign, click on 'Finish later' for Message B followed by 'Back to campaigns', check that page is routed to the dashboard view

### DIFF
--- a/frontend/src/components/dashboard/create/email/EmailTemplate.tsx
+++ b/frontend/src/components/dashboard/create/email/EmailTemplate.tsx
@@ -105,13 +105,14 @@ const EmailTemplate = ({
       <SaveDraftModal
         saveable
         onSave={async () => {
-          if (campaignId && subject && body && from) {
-            try {
-              await saveTemplate(+campaignId, subject, body, replyTo, from)
-            } catch (err) {
-              setErrorMsg(err.message)
-              throw err
-            }
+          if (!campaignId) return
+          try {
+            if (!subject || !body)
+              throw new Error('Message subject or template cannot be empty!')
+            await saveTemplate(+campaignId, subject, body, replyTo, from)
+          } catch (err) {
+            setErrorMsg(err.message)
+            throw err
           }
         }}
       />
--- a/frontend/src/components/dashboard/create/Create.tsx
+++ b/frontend/src/components/dashboard/create/Create.tsx
@@ -38,7 +38,7 @@ const Create = () => {
   async function handleFinishLater() {
     if (campaign.status === Status.Draft) {
       sendUserEvent(GA_USER_EVENTS.FINISH_CAMPAIGN_LATER, campaign.type)
-      if (finishLaterContent) finishLaterContextHandler()
+      if (finishLaterContent) return finishLaterContextHandler()
     }
     history.push('/campaigns')
   }
--- a/frontend/src/components/dashboard/create/sms/SMSTemplate.tsx
+++ b/frontend/src/components/dashboard/create/sms/SMSTemplate.tsx
@@ -80,13 +80,13 @@ const SMSTemplate = ({
       <SaveDraftModal
         saveable
         onSave={async () => {
-          if (campaignId && body) {
-            try {
-              await saveTemplate(+campaignId, body)
-            } catch (err) {
-              setErrorMsg(err.message)
-              throw err
-            }
+          if (!campaignId) return
+          try {
+            if (!body) throw new Error('Message template cannot be empty!')
+            await saveTemplate(+campaignId, body)
+          } catch (err) {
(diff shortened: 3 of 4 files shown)


lamkeewei:
Ran through the tests scenarios and they work as expected. However, there's one additional scenario where I'm not sure if the current behaviour is the correct one. 

1. Create a new campaign
2. Enter and save a valid template
3. Return to Step 1 and clear the template (set it to completely empty)
4. Click finish later and save draft

In this scenario, the old non-empty template will remain unchanged. 

Perhaps if the `body` and/or `subject` is empty, we should disable the "Save Draft" button and only allow the user to "Skip saving draft". Similar to the "Next" button behaviour.

miazima:
I would like to avoid disabling the `Save Draft` button, as 
1. It requires adding logic to `SaveDraftModal` and `ConfirmDraftModal` that is unlikely to be reused. 
2. If we were to disable the button, we would still have to display an error/warning to inform user on _why_ they are not allowed to proceed with that action. 

I believe it may be more intuitive to throw an error like `Message template cannot be empty!` on clicking `Save Draft` instead. So if `body/subject` is empty we do not call `saveTemplate` and instead throw an error, which would display the error on the `SaveDraft` modal and display the `Edit Draft` button. What do you think?

lamkeewei:
> I would like to avoid disabling the `Save Draft` button, as
> 
> 1. It requires adding disabling logic to `SaveDraftModal` and `ConfirmDraftModal` that is unlikely to be reused.
> 2. If we were to disable the button, we would still have to display an error/warning to inform user on _why_ they are not allowed to proceed with that action.
> 
> I believe it may be more intuitive to throw an error like `Message template cannot be empty!` on clicking `Save Draft` instead. So if `body/subject` is empty we do not call `saveTemplate` and instead throw an error, which would display the error on the `SaveDraft` modal and display the `Edit Draft` button. What do you think?

Sounds good. Given how our `ConfirmModal` is structured, agree that it's best that we don't just bolt on a new functionality just for this edge case.

frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -105,13 +105,14 @@ const EmailTemplate = ({
...
-            try {
-              await saveTemplate(+campaignId, subject, body, replyTo, from)
-            } catch (err) {
-              setErrorMsg(err.message)
-              throw err
-            }
+          if (!campaignId) return
+          try {
+            if (!subject || !body)
+              throw new Error('Message subject or template cannot be empty!')
```
lamkeewei, line 111: Technically, we don't need to manually throw an error because the backend endpoint would throw an error if either body or subject is empty. 

But with that said, I think this error message is more user friendly compared to the one from the backend endpoint. 

image

lamkeewei (approved):
lgtm! Tested following and works as expected:
1. Attempt to save draft for an entirely new campaign with an empty template (throws error)
2. Attempt to edit and then save draft for an existing campaign with an empty template (throws error)

Issue #798, opened 2020-10-22
Announcement pop-up


jeantanzj:
During the discussion about trial mode, the designers requested for an announcement pop-up / modal, which is logically a feature by itself. 

- This announcement pop-up should be displayed when a user logs into Postman
- When the user clicks away  / closes the pop-up, the modal should be dismissed forever, until the next announcement. 

**Proposed implementation**
- Create a modal with four sections - title, subtitle, image, subtext
- Using lingui, create four messages: `announcement.title`, `announcement.subtitle`, `announcement.imageUrl`, `announcement.subtext`, which will be displayed in the sections in the modal
- Using lingui, create a fifth message `announcement.version` which is a version number, incremented whenever a new announcement is made.
- Add a user_features table (if not already introduced by trial mode): 
`user_features (user_id int fk(users), announcement_version int default null, created_at timestamp, updated_at timestamp)`
- When a user logs in, GET /auth/userinfo returns `user_features.announcement_version` for that user.
- If `user_features.announcement_version` is null, display the announcement.
- If `user_features.announcement_version` is equal to the current version on the frontend (the fifth message `announcement.version` as specified above), do not display the announcement.
-  If `user_features.announcement_version`  is less than the current version on the frontend, display the announcement. 
- Add another endpoint to update  `user_features.announcement_version` to the current version, when the user clicks away/closes the pop up.

PR #888, opened 2020-12-10
[WIP] feat: announcement modal backend

theMogget:
- added user_features table model
- updated auth/userinfo endpoint
- added announcementVersion into the auth context

## Problem

Building a modal to display new feature launch announcements to users when they revisit the app.

re: #798

## Solution

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

## Deploy Notes

_Notes regarding deployment of the contained body of work. These should note any
new dependencies, new scripts, etc._

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

- `dependency` : dependency details

**New dev dependencies**:

- `dependency` : dependency details


frontend/src/contexts/auth.context.tsx
```diff
@@ -15,6 +15,8 @@ interface ContextProps {
   setAuthenticated: Dispatch<SetStateAction<boolean>>
   email: string
   setEmail: Dispatch<SetStateAction<string>>
+  announcementVersion: number | null | undefined
```
theMogget, line 18: Should `announcementVersion` should be added to the `AuthContextProvider`? On one hand `announcementVersion` does need to be visible to all components, since the modal should pop up regardless of which page you decided to navigate to. On the other hand, I don't know if it's considered relevant to 'auth'.

PR #893, opened 2020-12-11
feat(announcement-modal): Backend add announcementVersion user metadata

theMogget:
## Problem

We need to track the latest announcement a user has seen, and display any later announcements.

re: #798 

## Solution
**Features**:
* Add user_features table model
* Add UserFeatures to User table model
* Edit `/settings` to also return announcement_version
* Create new `/settings/announcement-version` endpoint to update user's announcement_version


backend/src/core/models/user/user-features.ts
```diff
@@ -0,0 +1,27 @@
...
+  @Column({
+    type: DataType.INTEGER,
+    allowNull: false,
+    primaryKey: true,
+  })
+  userId!: number
+
+  @Default(null)
+  @Column({
+    type: DataType.INTEGER,
```
jeantanzj, line 23: string?  the announcementVersion should probably follow the package version (https://semver.org/)

backend/src/core/models/user/user.ts
```diff
@@ -35,6 +36,9 @@ export class User extends Model<User> {
   @HasOne(() => UserDemo)
   demo!: UserDemo
 
+  @HasOne(() => UserFeatures)
+  userFeatures!: UserFeatures
```
jeantanzj, line 40: rename this too

backend/src/core/routes/settings.routes.ts
```diff
@@ -27,6 +27,12 @@ const demoDisplayedValidator = {
   }),
 }
 
+const updateAnnouncementVersionValidator = {
+  [Segments.BODY]: Joi.object({
+    announcement_version: Joi.number().required(),
```
jeantanzj, line 32: type also has to change here

backend/src/core/services/credential.service.ts
```diff
@@ -250,6 +261,7 @@ const getUserSettings = async (
       hasApiKey: !!user.apiKey,
       creds: user.creds,
       demo: user.demo,
+      userFeatures: user.userFeatures,
```
jeantanzj, line 264: basically rename things across all the files where they should be renamed

jeantanzj (changes requested):

backend/src/core/models/user/user-features.ts
```diff
@@ -0,0 +1,27 @@
...
+  DataType,
+  Model,
+  Table,
+  ForeignKey,
+  Default,
+} from 'sequelize-typescript'
+import { User } from './user'
+
+@Table({ tableName: 'user_features', underscored: true, timestamps: true })
+export class UserFeatures extends Model<UserFeatures> {
```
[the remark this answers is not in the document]
theMogget: Done!

backend/src/core/models/user/user-features.ts
theMogget: Done~

PR #910, opened 2020-12-16
Migrate worker to get environment variables from Secrets Manager 

### LINKED ISSUE #813: Secrets Management: Worker should use env vars from secret manager
Options
1. Copy the implementation from backend to inject env vars
2. Look into this https://docs.aws.amazon.com/AmazonECS/latest/developerguide/specifying-sensitive-data-secrets.html 

Caveats
Some env vars have to be renamed because they're named the same in backend and worker but actually have different values

miazima:
## Problem

Worker uses environment variables set in the ECS task definition, many of of them are similar to backend en vars. The worker can share secrets with backend and all the secrets can be consolidated in a secret in Secret Manager. Hence, we need to migrate worker to pull secrets from Secrets Manager too. There are some conflicting env vars which have similar names as backend en vars, but differ in value - these have to be renamed.

Closes #813 

## Solution

**Features**:

- Pull secrets from Secret Manager 
- Load secrets in .env file before starting up worker

## Tests

- Deploy to staging
- Create a new task definition by removing e.g. `DB_URI` env var from the previous ECS task definition
- Create a campaign and ensure campaign is sent out successfully

## Deploy Notes

Renamed `worker` en vars to prepend `Worker`
- `SES_PASS` -> `WORKER_SES_HOST`
- `SES_PASS` -> `WORKER_SES_PASS`
- `SES_USER` -> `WORKER_SES_USER`
- `SES_PORT` -> `WORKER_SES_PORT`

**New scripts**:

- `before-deploy.sh` : Script that executes before travis deployment to inject secret id according to staging/prod branch
- `docker-entrypoint.sh`: Retrieve secrets from aws secrets manager and store them `.env` file

**Environment variables to be moved to Secret Manager**
- `AWS_REGION`
- `BACKEND_URL`
- `DB_URI`
- `DB_USE_IAM`
- `DEFAULT_COUNTRY`
- `NODE_ENV`
- `SECRET_MANAGER_SALT`
- `WORKER_SES_FROM`
- `WORKER_SES_HOST`
- `WORKER_SES_PASS`
- `WORKER_SES_PORT`
- `WORKER_SES_USER`
- `TWILIO_CALLBACK_SECRET`
- `UNSUBSCRIBE_HMAC_ALGO_V1`
- `UNSUBSCRIBE_HMAC_KEY_V1`

**Environment variables that will remain in ECS**
- `ECS_SERVICE_NAME`
- `MESSAGE_WORKER_LOGGER`
- `MESSAGE_WORKER_SENDER`


lamkeewei:
@miazima could I trouble you to document the following in the deploy notes: 
- Env vars that we will still need to keep in the task definitions (sending and logging)
- Env vars we can remove since they are now loaded from the `.env` file created from SecretsManager

Would be good to keep track. If we don't remove from the task definition, the will end up overriding whatever that is loaded from Secrets Manager.

miazima:
> @miazima could I trouble you to document the following in the deploy notes:
> 
> * Env vars that we will still need to keep in the task definitions (sending and logging)
> * Env vars we can remove since they are now loaded from the `.env` file created from SecretsManager
> 
> Would be good to keep track. If we don't remove from the task definition, the will end up overriding whatever that is loaded from Secrets Manager.

Yep, good to have that stated clearly, have updated the deployment notes!

lamkeewei:
@miazima seems be a failure in the worker build. Any idea why?

miazima:
> @miazima seems be a failure in the worker build. Any idea why?

Not sure either, retried and build succeeded.

lamkeewei:
> > @miazima seems be a failure in the worker build. Any idea why?
> 
> Not sure either, retried and build succeeded.

I think the builds are failing from time to time because of the timeout for `ecs-deploy`. Relevant [code](https://github.com/silinternational/ecs-deploy/blob/f2f0bfad8eceed924989b1404ca329b01fd50ff3/ecs-deploy#L515). Perhaps we can increase it, currently it's 90s and see if the build succeeds?

zwliew:
Good catch, I'll try it out.

zwliew:
Is it normal for the deployment of the sender ECS worker to take over 10 min? Maybe it's stalling?

lamkeewei:
> I've bumped the timeout to 10min but it's still not enough. I don't think this is normal. Maybe something is stalling the deployment?
> 
> 
> 
> I'd check the logs but I'm not sure which one to check for matters regarding the ECS deployment.

Perhaps we can add verbose flag to `ecs-deploy` to see if there's anything useful.

lamkeewei:
We can't have both `MESSAGE_WORKER_LOGGER` and `MESSAGE_WORKER_SENDER` in the config secret. This will prevent the container from starting up.

zwliew:
Turns out there was a typo in some of the SES environment variables (e.g. `WORKER_ SES_HOST` vs `WORKER_SES_HOST`). My bad -- I didn't catch this earlier when I checked them!

The builds run fine now.

lamkeewei:
> Turns out there was a typo in some of the SES environment variables (e.g. `WORKER_ SES_HOST` vs `WORKER_SES_HOST`). My bad -- I didn't catch this earlier when I checked them!
> 
> The builds run fine now.

Do you think we should still add the timeout flag to increase the timeout. If the deploy succeeds before the timeout, it will just move on.

zwliew:
> > Turns out there was a typo in some of the SES environment variables (e.g. `WORKER_ SES_HOST` vs `WORKER_SES_HOST`). My bad -- I didn't catch this earlier when I checked them!
> > The builds run fine now.
> 
> Do you think we should still add the timeout flag to increase the timeout. If the deploy succeeds before the timeout, it will just move on.

Yup, good point. I guess 5min should be fine?

zwliew:
Alright, sounds good!

lamkeewei (approved):
lgtm! tested on staging.

PR #920, opened 2020-12-22
feat: Announcement modal

### LINKED ISSUE #798: Announcement pop-up

theMogget:
## Problem

We want users to know when we release new features.

Closes #798 

## Solution

Create an announcement modal that pops up on login. Announcements are tied to our own package version numbers.

[Design docs](https://www.figma.com/file/29FkTqWnA4yLKzstnvyY6X/%F0%9F%91%91Design-Master?node-id=2847%3A45015)

**Features**:

* Checks what announcement version the user has last seen, and displays the latest announcement if the user's announcement version is earlier than the current version
* Can be toggled on or off, since we may not want to create announcements for every semver bump

## Screenshots

**Desktop**
image

**iPhone5/SE**
image

## Tests

(Using `1.17.2` as the current package version.)

**New users, or users who have never seen an announcement:**
* In the `user_features` table, delete the row associated with your user's `id`
* The modal should appear
* After closing, `user_features` should include a new row with your `id` and the updated `announcement_version`

**Returning users who have not seen the latest announcement:**
* In the `user_features` table, modify the row associated with your user's `id` to `0.1.1`
* The modal should appear
* After closing, `user_features` should include a new row with your `id` and the updated `announcement_version`

**Returning users who have already seen the latest announcement:**
* In the `user_features` table, modify the row associated with your user's `id` to ` 1.17.2`
* The modal should not appear

**Toggle announcement model active or not:**
To display the modal:
* In Amplify, `REACT_APP_ANNOUNCEMENT_ACTIVE` 

* In the `user_features` table, delete the row associated with your user's `id`
* If the flag is `true`, the modal should appear. If it is `false`, there should be no announcement modal.

**Update announcement version on modal close:**
* Closing the modal by clicking on the background, the `X` close button, or following the prompted link should all trigger the `updateAnnouncementVersion` call

**Announcement version validation:**
* Only calls to the DB with appropriately formatted versions (i.e `x.x.x` semantic versioning, with no letters) should succeed
* Only calls with the announcement versions less than or equal to the current announcement version are allowed


backend/src/core/routes/settings.routes.ts
```diff
@@ -27,6 +27,12 @@ const demoDisplayedValidator = {
   }),
 }
 
+const updateAnnouncementVersionValidator = {
+  [Segments.BODY]: Joi.object({
+    announcement_version: Joi.string().default(null),
```
lamkeewei, line 32: This needs to be validated. You may want to make it required. In addition, you should also check that the value 
- Has the right format (X.Y.Z)
- Is less than or equals to the current version

Otherwise a user might be able to set this to an invalid value through the API.

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -63,18 +66,45 @@ const Campaigns = () => {
     setLoading(false)
   }
 
+  // Returns true if A < B, else false
+  function compareSemver(A: string, B: string) {
```
lamkeewei, line 70: Minor nitpick. Generally we use camelCase for variable names. Perhaps more informative variable names `lastSeenVersion` and `currentVersion` also?

frontend/src/components/dashboard/campaigns/announcement-modal/AnnouncementModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { ANNOUNCEMENT } from 'config'
+import { OutboundLink } from 'react-ga'
+
+const AnnouncementModal = () => {
+  const { close, setBeforeClose } = useContext(ModalContext)
+  const [errorMessage, setErrorMessage] = useState('')
+
+  useEffect(() => {
+    setBeforeClose(() => () => {
+      updateAnnouncementVersion(i18n._(ANNOUNCEMENT.version))
```
lamkeewei, line 19: You will need to handle the error thrown if `updateAnnouncementVersion` fails.

frontend/src/locales/en/messages.po
```diff
@@ -61,6 +61,30 @@ msgstr "Verifying OTP..."
...
+
+#: src/config.ts:62
+msgid "announcement.readGuideUrl"
+msgstr "https://guide.postman.gov.sg/release-notes#2020-11-05-or-demo-campaigns-for-sms-and-telegram"
+
+#: src/config.ts:60
+msgid "announcement.subtext"
+msgstr "We've heard your requests to cater to more channels. Good news, Postman is now supporting agencies to utilise their Telegram bots! 
We have chosen to work with Telegram as it is a messaging app with 
a focus on speed and security. It's super-fast, simple and free."
```
lamkeewei, line 74: Might need to check with either Petty or the designers what should be the exact content for announcement.

backend/src/core/routes/settings.routes.ts
```diff
@@ -184,4 +190,44 @@ router.put(
...
+ *       summary: Update announcement version for this user
+ *       requestBody:
+ *         required: true
+ *         content:
+ *           application/json:
+ *             schema:
+ *               type: object
+ *               properties:
+ *                 announcement_version:
+ *                   type: number
```
lamkeewei, line 209: Should update the type here

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -63,18 +66,45 @@ const Campaigns = () => {
...
+    if (A == null) {
```
lamkeewei, line 71:
```suggestion
    if (A === null) {
```

theMogget: I did `==` instead of `===` here to account for the case where the user wasn't found in the `user_features` table yet - hence A would be `undefined`. Is there a better way to handle that case?

frontend/src/components/dashboard/campaigns/announcement-modal/AnnouncementModal.module.scss
```diff
@@ -0,0 +1,57 @@
...
+
+.close {
+  position: absolute;
+  top: 2rem;
+  right: 2rem;
+  &.modalTitleClose {
+    color: $green-900;
+    top: calc(#{$title-height} / 2 - 1.5rem); // close button has height of 3rem
+  }
+}
```
theMogget, line 57: This styling is... janky. I mainly tried to follow what was in the figma - but I realise we may have changed since then because of the new design system. Please let me know if there's something better I should model after!

frontend/src/components/dashboard/campaigns/Campaigns.tsx
lamkeewei:
```suggestion
    if (!A) {
```

Maybe check for a [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) value?

frontend/src/components/dashboard/campaigns/announcement-modal/AnnouncementModal.module.scss
lamkeewei: I think desktop looks ok. Might want to fix mobile though.

frontend/src/components/dashboard/campaigns/announcement-modal/AnnouncementModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+      setErrorMessage(err.message)
+    }
+  }
+
+  return (
+    <div className={styles.modal}>
+      <div className={styles.modalImg}>
+        <img src={i18n._(ANNOUNCEMENT.imageUrl)} alt="Modal graphic"></img>
+      </div>
+      <h2 className={styles.title}>{i18n._(ANNOUNCEMENT.title)}</h2>
```
lamkeewei, line 38:
```suggestion
      <h2>{i18n._(ANNOUNCEMENT.title)}</h2>
```

Perhaps remove the classes if they are not used?

frontend/src/components/dashboard/campaigns/announcement-modal/AnnouncementModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+      close()
+    } catch (err) {
+      setErrorMessage(err.message)
+    }
+  }
+
+  return (
+    <div className={styles.modal}>
+      <div className={styles.modalImg}>
+        <img src={i18n._(ANNOUNCEMENT.imageUrl)} alt="Modal graphic"></img>
```
lamkeewei, line 36: We might want to either:
1. Conditionally render `img` based on whether there's an image url provided
2. Fallback to a default image

We might not have a unique announcement image for each announcement.

frontend/src/components/dashboard/campaigns/announcement-modal/AnnouncementModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+      </div>
+      <h2 className={styles.title}>{i18n._(ANNOUNCEMENT.title)}</h2>
+      <h4 className={styles.subtitle}>{i18n._(ANNOUNCEMENT.subtitle)}</h4>
+      <div className={styles.content}>{i18n._(ANNOUNCEMENT.subtext)}</div>
+      <div className={styles.options}>
+        <OutboundLink
+          eventLabel={i18n._(ANNOUNCEMENT.readGuideUrl)}
+          to={i18n._(ANNOUNCEMENT.readGuideUrl)}
```
lamkeewei, line 44:
```suggestion
          to={i18n._(ANNOUNCEMENT.readGuideUrl)}
          target="_blank"
```

Might be better to open the link in a new table rather than navigating the user away from Postman entirely.

frontend/src/components/dashboard/campaigns/Campaigns.tsx
theMogget: Done!

frontend/src/components/dashboard/campaigns/announcement-modal/AnnouncementModal.tsx
theMogget: done!

frontend/src/components/dashboard/campaigns/announcement-modal/AnnouncementModal.tsx
theMogget: I clarified this with Nat; she said she'll get back to me. :)

frontend/src/components/dashboard/campaigns/announcement-modal/AnnouncementModal.tsx
theMogget: done!

backend/src/core/routes/settings.routes.ts
theMogget: Did some simple regex for the first, but for the second check - Is less than or equals to the current version - there would need to either be a DB call to check the current version (which I believe Joi can't do), or I can change the endpoint to receive two versions instead (current, lastSeen). But I'm not sure if that would necessarily provide the additional protection we want?

lamkeewei: I think should be able to retrieve the current package version using `process.env.npm_package_version`. https://stackoverflow.com/a/48619640. Mainly want to avoid being able to set the user's version to something greater than the current package version. Otherwise, they will miss all future announcements.

frontend/src/locales/en/messages.po
```diff
@@ -61,6 +61,30 @@ msgstr "Verifying OTP..."
...
+msgid "announcement.subtext"
+msgstr "Thanks for using Postman and joining us every step of the way while we mature as a product. We'll be releasing exciting new features this year. For instance, this is an announcement modal that will keep you up-to-date about our new features! Watch this space for things that are brewing. Learn more about Postman's releases by reading our guide."
+
+#: src/config.ts:59
+msgid "announcement.title"
+msgstr "Happy New Year! Congrats on making it through 2020."
+
+#: src/config.ts:58
+msgid "announcement.version"
+msgstr "1.1.1"
```
lamkeewei, line 86: Should be set this to be in line with the package version?

frontend/src/config.ts
```diff
@@ -54,3 +54,15 @@ export const SENTRY_RELEASE = process.env.REACT_APP_SENTRY_RELEASE as string
 export const SENTRY_ENVIRONMENT =
   (process.env.REACT_APP_SENTRY_ENVIRONMENT as string) || 'development'
 export const INFO_BANNER = process.env.REACT_APP_INFO_BANNER as string
+
+// Feature Launch Announcements
+// If `isActive` is false, the modal will not proc for ANY user
+export const ANNOUNCEMENT = {
+  isActive: true,
```
lamkeewei, line 61:
```suggestion
  isActive: process.env.REACT_APP_ANNOUNCEMENT_ACTIVE || false,
```

Maybe make this an env var?

backend/src/core/services/credential.service.ts
```diff
@@ -292,6 +304,35 @@ const updateDemoDisplayed = async (
...
+    { returning: true }
+  )
+
+  if (rowUpserted.userId !== userId) {
+    logger.error({
+      message: 'Updated incorrect user',
+      action: 'updateAnnouncementVersion',
+    })
+    throw new Error('Could not update announcement version')
+  }
```
miazima, line 329: I believe it is very unlikely for the wrong userId to have been updated? Maybe instead you can check if `!rowUpserted` (is null) to ensure the upsert was successful.

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -63,18 +66,50 @@ const Campaigns = () => {
...
+    async function getNumDemosAndAnnouncementVersion() {
+      const { demo, announcementVersion } = await getUserSettings()
       setIsDemoDisplayed(demo?.isDisplayed)
       setNumDemosSms(demo?.numDemosSms)
       setNumDemosTelegram(demo?.numDemosTelegram)
+      displayNewAnnouncement(announcementVersion)
     }
-    getNumDemos()
+    getNumDemosAndAnnouncementVersion()
+    // eslint-disable-next-line
```
miazima, line 112: Is there a dependency we are trying to avoid? Not sure if the infinite re-render issue was resolved from when we last discussed it

frontend/src/components/dashboard/campaigns/announcement-modal/AnnouncementModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+  useEffect(() => {
+    setBeforeClose(() => async () => {
+      try {
+        await updateAnnouncementVersion(i18n._(ANNOUNCEMENT.version))
+      } catch (err) {
+        setErrorMessage(err.message)
+      }
+    })
+    // eslint-disable-next-line
+  }, [])
```
miazima, line 26: Any dependency we are trying to avoid here too? It's best practice to include dependencies suggested by `exhaustive-deps`, usually its a logic flaw on our side that may result in errors later on.

lamkeewei:
Styling still looks off for mobile. Maybe can consider the following? 

In `Modal.module.scss`

```scss
.modalImg {
    max-height: 250px;
    
    @include mobile() {
        max-height: 20vh;
    }
}
```

In `Modal.module.scss`: 
```scss
.content {
    ...
    @include mobile() {
        padding: 6rem 3rem;
    }
}
```

backend/src/core/routes/settings.routes.ts
```diff
@@ -27,10 +27,30 @@ const demoDisplayedValidator = {
   }),
 }
 
+// Only allow updating with versions less than the current package version
+const compareSemverMethod = (value: string, helpers: any) => {
```
lamkeewei, line 31: Should this be less than or equals to?

frontend/src/components/dashboard/campaigns/announcement-modal/AnnouncementModal.tsx
theMogget: Ah yes thanks for the catch, this is a change I should've reverted when trying to make the context work. Done!

frontend/src/config.ts
theMogget: Good point, thanks~

backend/src/core/routes/settings.routes.ts
theMogget: Both work! I made it strictly less than since technically speaking there shouldn't be a case where the `updateAnnouncementVersion` function is called with equal versions, it would mean something had gone wrong. What do you think?

frontend/src/contexts/modal.context.tsx
```diff
@@ -13,14 +14,33 @@ export const ModalContext = createContext(defaultValue)
 const ModalContextProvider = ({ children }: { children: React.ReactNode }) => {
   const [modalContent, setModalContent] = useState(null)
   const [modalTitle, setModalTitle] = useState('')
-  function handleClose() {
+
+  // Important: to pass a function into setBeforeClose, you must anonymize it twice
+  // i.e `() => () => { some function }
+  // see: https://medium.com/swlh/how-to-store-a-function-with-the-usestate-hook-in-react-8a88dd4eede1
+  const [beforeClose, setBeforeClose] = useState<(() => void) | null>(null)
```
lamkeewei, line 21:
```suggestion
  const [beforeClose, setBeforeClose] = useState<(() => Promise<void>) | null>(null)
```

frontend/src/components/dashboard/campaigns/Campaigns.tsx
theMogget: Discussed this with @lamkeewei -- we decided to implement a check to see if `modalContent` is currently `null` or not. If it is not null, then we'll trigger `setModalContent`. This helps with breaking the loop since the context isn't being constantly updated.

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -63,19 +66,56 @@ const Campaigns = () => {
     setLoading(false)
   }
 
+  // Returns true if lastSeenVersion < currentVersion, else false
+  function compareSemver(currentVersion: string, lastSeenVersion: string) {
```
lamkeewei, line 70: Minor nitpick. Shouldn't currentVersion and lastSeenVersion be flipped? The function is called below with `compareSemver(announcementVersion, i18n._(ANNOUNCEMENT.version))`

lamkeewei (approved):
- Announcement modal shown to new users
- Announcement modal shown to existing users that have not seen an announcement before
- Users that have seen a previous version of the announcement is shown the latest version when they navigate to the campaigns page
- Toggling `REACT_APP_ANNOUNCEMENT_ACTIVE` on Amplify will enable/disable modal for all users regardless of the versions they have seen
- Modal is dismissed and announcement version is updated by one of the following actions
   - Click on background
   - Close button
   - Learn more button
- `PUT /settings/announcement-version` checks that version is <= current version

PR #928, opened 2020-12-24
Truncate log messages to prevent large logs causing memory issues

miazima:
## Problem

There are a few instances in the code where we write the whole error object in the `catch` block to logs. An error object that contains an abnormally large amount of data, e.g. very long stack traces or sequelize error trace, could cause the server to go out of memory when writing the logs to cloudwatch. This would cause the server to restart, causing a temporary loss in service. 

## Solution

After combing through the logs to identify such instances, my initial solution was to limit all functions with sequelize queries to output error logs with only `error.message`. However I realised this may not be a solid solution, as described below.

**Potential solutions**:

- Limit all functions with sequelize queries to output error logs with only `error.message`. This is not ideal as:
  1. The functions with sequelize queries may be also performing calls to other services for which a more verbose error message would be ideal i.e. would loose stack traces
  2. This issue may not be limited to sequelize error traces
  2. This is not foolproof as it may not be feasible to ensure that new code being written has to follow this convention
- Stringify and truncate all `error` object root properties to an arbitrary number of bytes
- Truncate the whole error object if larger than set threshold - destroys json structure

**Immediate fixes to resolve error source**:

- Only print out `error.message` in `parse-csv.service` to prevent regurgitating long bulk create queries on errors during file processing 

## Tests

- Upload a large recipient list with recipients that exceed 255 chars
- Observe logs

### DIFF
--- a/backend/src/core/logger.ts
+++ b/backend/src/core/logger.ts
@@ -44,12 +44,25 @@ const addTransport = (transport: any) => {
   logger.add(transport)
 }
 
+const truncate = (message: string, length: number, suffix = true): string => {
+  if (!message) {
+    return ''
+  }
+  return message.length > length
+    ? `${message.substring(0, length)}${suffix ? '...<TRUNCATED>' : ''}`
+    : message
+}
 const loggerWithLabel = (module: NodeModule): any => {
   const label = getModuleLabel(module)
   return {
     info: (logMeta: any): winston.Logger => logger.info({ label, ...logMeta }),
-    error: (logMeta: any): winston.Logger =>
-      logger.error({ label, ...logMeta }),
+    error: (logMeta: any): winston.Logger => {
+      if (logMeta.error && logMeta.error instanceof Error) {
+        logMeta.error = truncate(logMeta.error.stack, 400)
+      }
+      return logger.error({ label, ...logMeta })
+    },
+
     debug: (logMeta: any): winston.Logger =>
       logger.debug({ label, ...logMeta }),
     warn: (logMeta: any): winston.Logger => logger.warn({ label, ...logMeta }),
--- a/backend/src/core/services/parse-csv.service.ts
+++ b/backend/src/core/services/parse-csv.service.ts
@@ -133,10 +133,10 @@ const parseAndProcessCsv = async (
           } catch (err) {
             logger.error({
               message: 'Parsing failed',
-              error: err.message,
+              error: err,
               ...logMeta,
             })
-            reject(err.message)
+            reject(err)
           }
         } else {
           logger.info({ message: 'Parsing aborted', ...logMeta })
--- a/backend/src/core/services/parse-csv.service.ts
+++ b/backend/src/core/services/parse-csv.service.ts
@@ -131,8 +131,12 @@ const parseAndProcessCsv = async (
             await onComplete(numMessages)
             logger.info({ message: 'Parsing complete', ...logMeta })
           } catch (err) {
-            logger.error({ message: 'Parsing failed', error: err, ...logMeta })
-            reject(err)
+            logger.error({
+              message: 'Parsing failed',
+              error: err.message,
+              ...logMeta,
+            })
+            reject(err.message)
           }
         } else {
           logger.info({ message: 'Parsing aborted', ...logMeta })


miazima:
We could truncate the logs when it exceeds some threshold size

```javascript
const loggerWithLabel = (module: NodeModule): any => {
  const label = getModuleLabel(module)
  return {
    info: (logMeta: any): winston.Logger => logger.info({ label, ...logMeta }),
    error: (logMeta: any): winston.Logger => {
      const { error } = logMeta
      // Check if error message is larger than threshold size
      // convert error object into buffer array and truncate
      // replace the error object with truncated error object string
      if (
        Buffer.from(JSON.stringify(error)).length > ERROR_LOG_MAX_SIZE_IN_BYTES
      ) {
        const errorBuffer = Buffer.from(JSON.stringify(error))
        const truncatedErrorBuffer = errorBuffer.slice(
          0,
          ERROR_LOG_MAX_SIZE_IN_BYTES
        )
        return logger.error({
          label,
          ...logMeta,
          error: truncatedErrorBuffer.toString('utf8'),
        })
      }
      return logger.error({ label, ...logMeta })
    },
    debug: (logMeta: any): winston.Logger =>
      logger.debug({ label, ...logMeta }),
    warn: (logMeta: any): winston.Logger => logger.warn({ label, ...logMeta }),
  }
}
```

jeantanzj:
> We could truncate the logs when it exceeds some threshold size
> 
> ```js
> const loggerWithLabel = (module: NodeModule): any => {
>   const label = getModuleLabel(module)
>   return {
>     info: (logMeta: any): winston.Logger => logger.info({ label, ...logMeta }),
>     error: (logMeta: any): winston.Logger => {
>       const { error } = logMeta
>       // Check if error message is larger than threshold size
>       // convert error object into buffer array and truncate
>       // replace the error object with truncated error object string
>       if (
>         Buffer.from(JSON.stringify(error)).length > ERROR_LOG_MAX_SIZE_IN_BYTES
>       ) {
>         const errorBuffer = Buffer.from(JSON.stringify(error))
>         const truncatedErrorBuffer = errorBuffer.slice(
>           0,
>           ERROR_LOG_MAX_SIZE_IN_BYTES
>         )
>         return logger.error({
>           label,
>           ...logMeta,
>           error: truncatedErrorBuffer.toString('utf8'),
>         })
>       }
>       return logger.error({ label, ...logMeta })
>     },
>     debug: (logMeta: any): winston.Logger =>
>       logger.debug({ label, ...logMeta }),
>     warn: (logMeta: any): winston.Logger => logger.warn({ label, ...logMeta }),
>   }
> }
> ```

you might OOM when you JSON.stringify() the error

miazima:
> you might OOM when you JSON.stringify() the error
I see.. Not sure how to workaround that, maybe some sort of chunking to not load the whole object into memory?

jeantanzj:
I stand corrected. Apparently JSON.stringifying error objects give you an "empty object"

```
> e = new Error('test')
Error: test
    at repl:1:5
    at Script.runInThisContext (vm.js:120:18)
    at REPLServer.defaultEval (repl.js:435:29)
    at bound (domain.js:427:14)
    at REPLServer.runBound [as eval] (domain.js:440:12)
    at REPLServer.onLine (repl.js:762:10)
    at REPLServer.emit (events.js:326:22)
    at REPLServer.EventEmitter.emit (domain.js:483:12)
    at REPLServer.Interface._onLine (readline.js:329:10)
    at REPLServer.Interface._line (readline.js:658:8)
> JSON.stringify(e)
'{}'
```

jeantanzj:
I think winston going OOM is a known issue (https://github.com/winstonjs/winston/issues/1624), and this current measure is not excellent but only a stopgap measure to prevent the logger from crashing our server. 

I considered multiple methods of handling this without throwing away too much information that's needed for troubleshooting. 
1) We could print just the err.message - not very useful because we dont know what led to it

2) Print very specific details of the error object - something like `err.sql ? truncate(err.sql) : ''`, then rewrap the error object to get rid of the large statements when we rethrow the error further downstream. This solution would only solve the problem for this scenario. Rewrapping the error object might cause unintended side effects - for example, when the error is rethrown, the message is actually saved in our db to be displayed to the user, so a different message might be displayed if we were not careful to test this flow.

3) Finally, I decided to print the stack trace for errors, because we'll get both the message and where it occurs, which should still be helpful in debugging. In this particular case, we lose the information about the sql statement, but I believe we can still debug the issue by retrieving the relevant s3 file.

jeantanzj:
@miazima can take a look?

lamkeewei (approved):
Tested with file that created a large error message. Error message is truncated.

Issue #880, opened 2020-12-07
API Backend Test 


pettypychen:
Write backend test for API routes.

syan-syan:
duplicated and in progress

PR #940, opened 2021-01-07
chore: setup scaffolding for backend tests

miazima:
Setup base scaffolding for backend tests.

The main testing framework used is jest. I considered using mocha and chai as alternative, but concluded on jest so that we use a common testing framework across frontend and backend.

Testing related frameworks/libraries used:
- Jest
- SuperTest (facilitates testing HTTP endpoints)
- redis-mock

Addresses #880 

~~**Mocking database queries**
`sequelize-mock` does not actually carry out db queries. Instead we mock query results by queuing the expected results e.g. `UserMock.$queueResult()`~~

**Database queries**
In the previous implementation, I had used `sequelize-mock` to mock sequelize query return values. After discussion with the team, we decided that database queries were a essential part and it would be better to test them with real queries instead. 

**Mocking user session**
A valid user session is introduced by intercepting the request and modifying `req.session` object. This is used for all routes that require authentication. 

- This may not be the ideal solution is it doesn't allow us to write tests with the presence and absence of a valid user session as needed.
- The `auth` routes does not have a user session mocked which makes testing `auth/userinfo` difficult as it is not possible to test with and without an user session

**Improvements**:

- include tests folder in `tsconfig.json` and exclude from build files 
- move all tests to `tests` folder
- add tests for authentication and campaign routes endpoints

## Tests

Run all tests `npm test`

## Deploy Notes

**New dev dependencies**:

- supertest
- redis-mock
- @types/redis-mock
-  @types/supertest


backend/tests/setup.ts
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable no-console */
+// global.console = {
+//   ...global.console,
+//   log: jest.fn(), // console.log are ignored in tests
```
lamkeewei, line 4: Perhaps we should keep this to suppress `console.log` in tests to keep the output more tidy.

backend/tests/routes/server.ts
```diff
@@ -0,0 +1,30 @@
...
+      id: 1,
+      email: 'user@agency.gov.sg',
+    }
+  }
+  next()
+})
+app.use(routes)
+app.use(celebrateErrorMiddleware())
+
+export default app
```
lamkeewei, line 30: Rather than instantiating and exporting `app` directly, perhaps we can expose a utility method that takes in options and return an express app? This way, each tests can specify what session object should be returned based on the test scenario.

lamkeewei (commented):
When running the tests, the following warning is raised:

> A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --runInBand --detectOpenHandles to find leaks.

We might want to look into this.

Another thing to note is that [jest runs tests in different files](https://github.com/facebook/jest/issues/6957#issue-358999337) in parallel unless `--runInBand` is set. This might be an issue if we have tests in each file interfering with each other. E.g. Test in `auth` deletes a user with email `test@test.com` while another test in `campaign` tries to create a campaign belonging to `test@test.com`. 

One way this can be solved is to ensure that each test suite runs in a separate namespace-ed db (e.g. `postmangovsg_test_auth`) by passing an argument to `seqeulizeLoader`. This db can be torn down at the end of each test suite. We want to avoid setting `--runInBand` due to the performance hit to tests since all test suite are now run serially. 

Found this [link](https://walrus.ai/blog/2020/04/testing-database-interactions-with-jest/) that might be useful. They namespace-ed the test database based on the Jest worker ID.

.travis.yml
```diff
@@ -16,6 +17,9 @@ jobs:
         - npx lockfile-lint --type npm --path package-lock.json -o "https:" -o "file:" --allowed-hosts npm
       install:
         - npm ci
+      before-script:
+        - psql -c 'create database postmangovsg_test_1;' -U postgres
+        - psql -c 'create database postmangovsg_test_2;' -U postgres
```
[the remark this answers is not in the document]
lamkeewei: Would it make sense to do this initialisation and the subsequent teardown in the `globalSetup` and the `globalTeardown` step? Just thinking along of lines of making it easier when running in dev so that he/she wouldn't have to remember to create/drop the two test db.

backend/src/core/routes/tests/auth.routes.test.ts
```diff
@@ -0,0 +1,132 @@
...
+  })
+
+  test('Invalid otp provided', async () => {
+    const res = await request(app)
+      .post('/auth/login')
+      .send({ email: 'user@agency.gov.sg', otp: '000000' })
+    expect(res.status).toBe(401)
+  })
+
+  test('OTP is invalidted after retries are exceeded', async () => {
```
lamkeewei, line 73:
```suggestion
  test('OTP is invalidated after retries are exceeded', async () => {
```

lamkeewei (approved):
lgtm! Left a minor remark about the creation of the test databases. Noted that this is just the scaffold, so I didn't check if all the required flows were addressed.

backend/package.json
```diff
@@ -4,14 +4,14 @@
...
-    "build": "rimraf build && tsc",
+    "build": "rimraf build && tsc -p tsconfig.build.json",
     "dev": "npm run postbuild && tsc-watch --onSuccess \"node ./build/server.js\"",
     "lint-no-fix": "tsc --noEmit && eslint --ext .js,.ts --cache .",
     "lint": "npm run lint-no-fix -- --fix",
     "postbuild": "npm run copy-assets",
     "start": "node build/server",
     "copy-assets": "copyfiles -u 1 src/assets/* src/**/*.sql build",
-    "test": "jest",
+    "test": "jest --maxWorkers=2",
```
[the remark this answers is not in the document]
lamkeewei: This should be ok for now. We can revisit this in the future if we have more tests and it's getting slow.

miazima:
I've refactored the initial setup to include a globalSetup file, however environmental variables are not loaded during globalSetup so I defined them within the file.

miazima:
@lamkeewei made some changes to the db setup, could you have a final look before we merge this

lamkeewei:
> @lamkeewei made some changes to the db setup, could you have a final look before we merge this

lgtm! Tested the creation of test database in global setup function and works as expected.

PR #952, opened 2021-01-19
chore(robots): guide NLB crawler

LoneRifle:
## Problem

Crawlers may be overly aggressive and parse bundle.js for possible links.
This results in the crawler trying to request paths from Postman that do not
actually exist

## Solution

None of the JavaScript source files will have links that crawlers can
follow, so prevent them from crawling them in the first place

### DIFF
--- a/frontend/public/robots.txt
+++ b/frontend/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /static/js/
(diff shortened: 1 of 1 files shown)


lamkeewei (approved):
lgtm. tested this on amplify and `robots.txt` is served correctly.

PR #955, opened 2021-02-01
fix(telegram): return 200 to skip Telegram API retries

lamkeewei:
## Problem

When an error occurs within `handleUpdate`, an error message is sent back to the user through Telegram to notify him/her. However, when this error occurs, our callback handler endpoint does not return a `200` response to the Telegram API. This causes the Telegram API to repeatedly retry the webhook. 

This results in the user being spammed with the same error message. 

## Solution

**Bug Fixes**:
- When an error occurs, log error and return 200 to the Telegram API to prevent retries
- Include `contact` and `from` in logs for debugging 

## Tests
- Setup and verify a bot on Postman
- Send a contact that is not your own contact
- Check in logs that `200` is returned by the callback endpoint
- Check that error message is only sent once

## TODO
- [x] Include additional information from the incoming message into the error log

### DIFF
--- a/backend/src/telegram/middlewares/telegram-callback.middleware.ts
+++ b/backend/src/telegram/middlewares/telegram-callback.middleware.ts
@@ -34,7 +34,7 @@ const verifyBotIdRegistered = async (
 const handleUpdate = async (
   req: Request,
   res: Response,
-  next: NextFunction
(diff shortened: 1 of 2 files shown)


miazima (approved):
Tested with the steps stated above, all checks passed.

PR #977, opened 2021-02-08
chore(localstack): pin localstack version to 0.11.6

### LINKED ISSUE #882: Presigned URL upload doesn't work for localstack versions >0.11.6
**Describe the bug**
Presigned URL upload fails for localstack versions > 0.11.6

**To Reproduce**

1. Pull localstack and run localstack image for versions > 0.11.6 (modify `docker-compose.yml` to set specific version)
3. Try to upload a CSV file
4. Error will be thrown

**Possible solutions**
- Pin localstack version for `docker-compose.yml`
- Investigate root cause for incompatibility (could be related to this [issue](https://github.com/localstack/localstack/issues/3266))

lamkeewei:
## Problem

Closes #882 

## Solution

**Bug Fixes**:

- Pin localstack to `0.11.6`. This is a temporary fix first so that `npm run dev:localstack` works. Will be taking a deeper look into what is causing the incompatibility. However, given that some features (custom from email) cannot be replicated with localstack, we might eventually deprecate it entirely.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
     ports:
       - "6379:6379"
   localstack:
-    image: localstack/localstack:latest
+    image: localstack/localstack:0.11.6
     ports:
       - "4566-4599:4566-4599"
     environment:


lamkeewei:
@zwliew could you help review to see if localstack works for you with version `0.11.6`

zwliew (approved):
This works!

So this is why I couldn't upload CSVs locally earlier...

PR #982, opened 2021-02-09
Increase request timeout on server and cloud run for verify-backup

miazima:
## Problem

The verify-backup service deployment was failing as it was not able to pass the readiness check by cloud run. The startup of the server was blocked by the verify-backup process which timed out as it required a long time for the db backup to be restored.

## Solution

Approaches taken to attempt to resolve this issue:
- increase the request timeout for cloud run (time within which a response must be returned by the verify-backup service)
- increasing cpu and memory allocation (by observation of the utilisation metrics we concluded this was not the issue)

However the verify-backup process did not run to completion and an error with HTTP 503 was thrown and the server would be restarted. This seemed to suggest that the server was aborting the request.

**Improvements that resolved the issue**:

- Increase the node request timeout to 20 min - this allows the the verify-backup process to run to completion before the server aborts the request
- Increase the request timeout for cloud run to 15 min

## Tests

Deploy the service to cloud run and observe logs for successful completion of database restoration.


lamkeewei:
@miazima was just looking at the logs in GCP console. It showed that the request completed in `263.4` seconds, which is roughly ~ 4mins. Was wondering if you tried with smaller values for the timeout and whether that worked.

miazima:
> @miazima was just looking at the logs in GCP console. It showed that the request completed in `263.4` seconds, which is roughly ~ 4mins. Was wondering if you tried with smaller values for the timeout and whether that worked.

I tested with a request timeout of 5 mins for cloud run, and it deployed successfully. I have updated the request timeout!

lamkeewei:
> > @miazima was just looking at the logs in GCP console. It showed that the request completed in `263.4` seconds, which is roughly ~ 4mins. Was wondering if you tried with smaller values for the timeout and whether that worked.
> 
> I tested with a request timeout of 5 mins for cloud run, and it deployed successfully. I have updated the request timeout!

Thanks for checking this out. Just wanted to make sure that the request timeout values are correlated with the length of time taken for the restoration. Would be good to add a bit of buffer, so that we don't have to change it so often.

serverless/verify-backup/scripts/verify-backup.sh
```diff
@@ -53,4 +53,4 @@ RESTORE_LOG="Successfully restored $DUMP_VERSION db dump! \
 notify "$RESTORE_LOG"
 
 # Completed Cronitor job
-curl "https://cronitor.link/$CRONITOR_CODE/complete" -m 10 || true
\ No newline at end of file
+curl "https://cronitor.link/$CRONITOR_CODE/complete" -m 10 || true
```
lamkeewei, line 56: Maybe we should move the corresponding Cronitor `/run` call into this script also? Otherwise, consecutive POST to the server endpoint will have a complete event without a corresponding start event on Cronitor.

Although, with that said, this mostly only happen in development since the container should also be started when the request from the scheduler comes in.

serverless/verify-backup/src/server.ts
```diff
@@ -1,14 +1,33 @@
 import express, { Request, Response } from 'express'
 import bodyParser from 'body-parser'
+import { spawn } from 'child_process'
```
lamkeewei, line 3: Another dev only issue. 

Environment variables in `.env` are not loaded on the first POST request. `config` is only imported when `npm run decrypt-dump` is called. Consequently, `dotenv.config()` is also only called at that time. As such when the child process is spawned, environment variables (e.g. `CRONITOR_CODE`) is not set/available.

serverless/verify-backup/src/server.ts
```diff
@@ -1,14 +1,33 @@
...
-app.post('/', (req: Request, res: Response) => {
-  console.log('Incoming pubsub message', JSON.stringify(req.body))
-  res.sendStatus(200)
+app.post('/', async (req: Request, res: Response) => {
+  // set request timeout to prevent server from aborting
+  // before verify-backup process is complete
+  req.setTimeout(20 * 60 * 1000) // 20 mins
+
+  // run verify-backup script
+  const verifyBackup = await spawn('sh', ['scripts/verify-backup.sh'])
```
lamkeewei, line 16:
```suggestion
  const verifyBackup = spawn('sh', ['scripts/verify-backup.sh'])
```

`await` has no effect here since `spawn` does not return a Promise.

serverless/verify-backup/scripts/deploy.sh
```diff
@@ -15,9 +15,11 @@ echo 'Deploying build to Cloud Run...'
...
   --region asia-southeast1 \
-  --no-allow-unauthenticated  \
+  --no-allow-unauthenticated \
   --image gcr.io/postmangovsg/verify-backup \
   --service-account $GCLOUD_SERVICE_ACCOUNT \
   --platform managed \
   --set-env-vars `grep -v '^#' .env | awk -v ORS=, 'NF { print $1 }'` \
-  --memory 4G
\ No newline at end of file
+  --timeout 5m \
```
lamkeewei, line 23: Perhaps we should [cap the maximum container](https://cloud.google.com/run/docs/configuring/max-instances#command-line) instance by setting `--max-instances 1` also? When testing it out, I noticed that when the response takes awhile to return, Cloud Run starts up another instance. Might be good to explicitly hard limit this to prevent a runaway auto-scale since the default is 1000 instance.

serverless/verify-backup/src/server.ts
```diff
@@ -1,14 +1,33 @@
...
+  verifyBackup.stdout.on('data', (data) => {
+    console.log(data.toString())
```
lamkeewei, line 18:
```suggestion
    console.log(data.toString().trim())
```

Minor nitpick, might want to add `trim` to prevent additional empty lines in the logs.

lamkeewei (approved):
Left some minor suggestions. But otherwise, lgtm!

Deployed and tested for production backups and works as expected.

serverless/verify-backup/scripts/verify-backup.sh
miazima: Yes I think it makes more sense to start the cronitor job within the verify-backup script, else if there is a lag between the server start and the verify-backup script start, we might get false job failures.

serverless/verify-backup/src/server.ts
miazima: Noted, I have imported `dotenv` in `server.ts` instead

serverless/verify-backup/scripts/deploy.sh
miazima: Yes that'll be a good precaution to put in, will set that

miazima: After testing, fixing the max instances at 1 increases the time taken for the request to complete. Its ~5.5mins now, so I think we can raise the request timeout from 5 mins to 8 mins?

lamkeewei:
@miazima changes looks good! Squash and merge whenever.

PR #984, opened 2021-02-09
fix(rich-text): unhandled error when removing image from popover

### LINKED ISSUE #983: Image removal generating an error and blank screen
**Describe the bug**
The screen went blank when I try to remove the image. 

**To Reproduce**

1.  Insert an image
2. Upload a CSV
3. Go back to step 1 to remove the image
4. See errors if you open the console (see screenshot)
5. If you refresh the screen, it jumps to the preview step (see screenshot)

**Expected behavior**
The image should be removed instead of the screen going blank. 

**Screenshots**
Screenshot 2021-02-09 at 4 13 54 PM
Screenshot 2021-02-09 at 4 14 42 PM

Chrome Version 88.0.4324.146 (Official Build) (x86_64) 
MacBook Pro

lamkeewei:
## Problem

If there was other text within the block that the image was inserted in, the image block will not be removed correctly through the "Remove" button in the popover. 

Closes #983 

## Solution

**Bug Fixes**:

- Residual characters in image block causes editor to crash when removing image with tooltip (#983)

## Tests
- Insert an image in the message and proceed to upload a CSV file
- Return to step 1 to remove the image
- There should not be a crash

### DIFF
--- a/frontend/src/components/common/rich-text-editor/blocks/ImageBlock.tsx
+++ b/frontend/src/components/common/rich-text-editor/blocks/ImageBlock.tsx
@@ -1,6 +1,5 @@
 import React, { useRef, useState, useContext } from 'react'
 import {
-  CharacterMetadata,
   ContentBlock,
   ContentState,
   EditorState,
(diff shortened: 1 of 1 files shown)


miazima (approved):
Tested on staging and works as expected

Issue #945, opened 2021-01-15
Horizontal padding missing on mobile view 


pettypychen:
**Issue**: No horizontal padding on mobile view
**Device**: Mobile
**Browser**: Edge
![Screenshot_20210114-205314](https://user-images.githubusercontent.com/39106042/104674301-5fa1b280-571e-11eb-9878-4e8c10c41679.jpg)
![Screenshot_20210114-205321](https://user-images.githubusercontent.com/39106042/104674316-67615700-571e-11eb-8ab9-332c1417cd9a.jpg)

Issue #911, opened 2020-12-16
Post-sending page Date format


NatMaeTan:
Date format on post-sending screen and dashboard are different. Post-sending screen date format should follow dashboard date format.

<img width="1000" alt="Screenshot 2020-12-16 at 2 13 51 PM" src="https://user-images.githubusercontent.com/63220614/102312067-f0af2d80-3fa8-11eb-9dd0-9ca4be8095d7.png">

<img width="1000" alt="Screenshot 2020-12-16 at 2 13 45 PM" src="https://user-images.githubusercontent.com/63220614/102312071-f278f100-3fa8-11eb-970a-e622d6760528.png">

Issue #877, opened 2020-12-02
Tall button pressed/selected states in creation modals


NatMaeTan:
<img width="332" alt="Screenshot 2020-12-01 at 4 45 22 PM" src="https://user-images.githubusercontent.com/63220614/100820740-d5f99680-3489-11eb-866c-b6de46060a90.png">

Tall pressed/selected buttons on the create modals (_Create new campaign_ and _Create demo campaign_) should not show a hover state. Since they're already selected, and they arent unselectable – You cannot unselect them, but only change your choice by selecting another channel so that at all times, 1 channel is selected. Hence there shouldnt be a hover state to indicate that it is selectable.

Hover states indicate that something can be interacted with. Now in production, when clicking on an already pressed/selected button in these modals, a hover state shows but nothing actually happens. It's misleading as it tells the user that you can click on this button, so they expect that something will happen but nothing does.

If the button is outlined in style, it should show that style for all states and not switch to a solid button style.

Issue #876, opened 2020-12-01
Text button + Solid Button text color


NatMaeTan:
<img width="469" alt="Screenshot 2020-12-01 at 5 38 16 PM" src="https://user-images.githubusercontent.com/63220614/100726525-0399fd80-3400-11eb-83e5-b9f839ab29c2.png">
 
To standardise text color in the _Text button + Solid Button_ component. Text button should be Deep Blue 040651

Note places to change this:
1. Creation flow previous/next. _Previous_ text button should be Deep Blue 040651
<img width="400" alt="Screenshot 2020-12-01 at 5 36 34 PM" src="https://user-images.githubusercontent.com/63220614/100726996-9dfa4100-3400-11eb-8cba-1f3209d88af3.png">

2. Create demo campaign WatchTheVideo/CreateDemoCampaign. _Watch the video_ text button should be Deep Blue 040651
<img width="400" alt="2  Test_CreateTestCampaign" src="https://user-images.githubusercontent.com/63220614/100727009-a6527c00-3400-11eb-9253-38bc6a65fce0.png">

PR #987, opened 2021-02-15
Fix some P3 UI issues

zwliew:
## Problem

Addresses various miscellaneous UI issues.

Addresses issues #875, #876, #877, #911, #945.

## Tests

For #875:
- For the TextArea component:
  - Border width should become 1.5px on hover
  - Border color should become #2C2CDC on hover

For #876:
- For the DemoVideoModal and CreateDemoModal components:
  - In desktop mode, margin between the secondary TextButton and the PrimaryButton components should be $spacing-7 (3rem).
  - In mobile mode, the secondary TextButton should above the PrimaryButton, and the both of them should be aligned horizontally

For #877:
- For the CreateModal and CreateDemoModal components:
  - The actively selected campaign type button should not have any hover state
  - The actively selected campaign type button should not do anything on click

For #911:
- For the ProgressDetails component:
  - The sent date format should be MMM DD YYYY, HH:MM

For #945:
- For the CreateModal and CreateDemoModal components:
  - In mobile mode, there should be horizontal margins between the content and the edge of the display.
- Note: this decreases the horizontal padding on all modals in general (see: the change in Modal.tsx), but this should not degrade the UI since the original padding was a bit excessive compared to the Figma design.

### DIFF
--- a/frontend/src/components/common/modal/Modal.module.scss
+++ b/frontend/src/components/common/modal/Modal.module.scss
@@ -43,7 +43,7 @@ $title-height: $height-4;
 .close {
   position: absolute;
   top: 2rem;
-  right: 2rem;
+  right: $spacing-1;
   &.modalTitleClose {
     color: $green-900;
     top: calc(#{$title-height} / 2 - 1.5rem); // close button has height of 3rem
@@ -58,6 +58,6 @@ $title-height: $height-4;
   min-height: $title-height;
   text-align: center;
   @include mobile {
-    padding-right: 3rem;
+    padding-right: 3.5rem;
   }
 }
--- a/frontend/src/components/common/modal/Modal.module.scss
+++ b/frontend/src/components/common/modal/Modal.module.scss
@@ -33,7 +33,7 @@ $title-height: $height-4;
 
 .content {
   max-width: 100vw;
-  padding: 3rem ($modal-width - 528px) / 2;
+  padding: 3rem ($modal-width - 600px) / 2;
   margin: auto;
   @include mobile() {
     padding: 6rem 3rem; // Need additional top padding to clear the close button
--- a/frontend/src/components/common/text-area/TextArea.module.scss
+++ b/frontend/src/components/common/text-area/TextArea.module.scss
@@ -21,9 +21,18 @@
       padding-bottom: 0.7rem;
     }
 
+    &:hover,
     &:focus {
       outline: none;
       border-color: $primary;
+    }
+
+    &:hover {
+      border-width: 1.5px;
+    }
+
+    &:focus {
+      border-width: 1px;
     }
 
     &::placeholder {
--- a/frontend/src/components/dashboard/create/create-modal/CreateModal.tsx
+++ b/frontend/src/components/dashboard/create/create-modal/CreateModal.tsx
@@ -96,7 +96,11 @@ const CreateModal = ({
                 className={cx(styles.button, {
                   [styles.active]: selectedChannel === ChannelType.SMS,
                 })}
-                onClick={() => setSelectedChannel(ChannelType.SMS)}
+                onClick={
+                  selectedChannel === ChannelType.SMS
+                    ? undefined
+                    : () => setSelectedChannel(ChannelType.SMS)
+                }
               >
                 SMS
                 <i
@@ -127,7 +131,11 @@ const CreateModal = ({
                 className={cx(styles.button, {
                   [styles.active]: selectedChannel === ChannelType.Telegram,
                 })}
-                onClick={() => setSelectedChannel(ChannelType.Telegram)}
+                onClick={
+                  selectedChannel === ChannelType.Telegram
+                    ? undefined
+                    : () => setSelectedChannel(ChannelType.Telegram)
+                }
               >
                 Telegram
                 <i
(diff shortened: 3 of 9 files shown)


miazima:
Great work on the changes, I noticed a two issues that would be good to resolve with this UI fix.

1. There is a hover state for tall buttons which is not currently implemented.
Design system: https://www.figma.com/file/Wd6jTqpdaqaItVLoyk1Me2/%F0%9F%8E%A8Postman-Design-System?node-id=3178%3A5
image

2. The header in the mobile view for demo modal is a little cramped.

zwliew:
> Great work on the changes, I noticed a two issues that would be good to resolve with this UI fix.
> 
> 1. There is a hover state for tall buttons which is not currently implemented.
>    Design system: https://www.figma.com/file/Wd6jTqpdaqaItVLoyk1Me2/%F0%9F%8E%A8Postman-Design-System?node-id=3178%3A5
>    image
> 2. The header in the mobile view for demo modal is a little cramped.
> 
> 

Thanks for the review! I'll get these fixed shortly. 

However, there is 1 issue regarding the hover state for tall buttons: in the Figma design, the on-hover background-color of the buttons is the same as the background-color of the modal (#f5f6fc). This makes it pretty difficult to visually identify the hover state.

frontend/src/components/dashboard/create/create-modal/CreateModal.tsx
```diff
@@ -96,7 +96,11 @@ const CreateModal = ({
                 className={cx(styles.button, {
                   [styles.active]: selectedChannel === ChannelType.SMS,
                 })}
-                onClick={() => setSelectedChannel(ChannelType.SMS)}
+                onClick={
+                  selectedChannel === ChannelType.SMS
+                    ? undefined
+                    : () => setSelectedChannel(ChannelType.SMS)
+                }
```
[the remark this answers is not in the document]
zwliew: It's not strictly necessary, since the `setSelectedChannel()` calls are no-op for active buttons. I only added it for completeness -- so that clicking on an active button literally does not do anything.

frontend/src/components/common/text-area/TextArea.module.scss
```diff
@@ -21,11 +21,20 @@
...
     }
 
+    &:hover {
+      border-width: 1.5px;
+    }
+
+    &:focus {
+      border-width: 1px;
+    }
+
```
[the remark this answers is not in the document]
zwliew: Yup, I've checked both TextInput and TextInputWithButton, and they already have the correct hover state.

miazima:
> However, there is 1 issue regarding the hover state for tall buttons: in the Figma design, the on-hover background-color of the buttons is the same as the background-color of the modal (#f5f6fc). This makes it pretty difficult to visually identify the hover state.
> 
> Perhaps a darker shade of grey should be used for this specific scenario instead?

Could I trouble you to check with @NatMaeTan on which color would be suitable?

frontend/src/components/common/modal/Modal.module.scss
```diff
@@ -58,6 +58,6 @@ $title-height: $height-4;
   min-height: $title-height;
   text-align: center;
   @include mobile {
-    padding-right: 3rem;
+    padding-right: 3.5rem;
```
miazima, line 61: Minor suggestion, I tried this with `padding: 0 2rem` to make it centered.

zwliew: I just tried `padding: 0 2rem`, but the close button still overlaps with the text at some widths.

miazima: Ah okay, I didn't test across different widths, we can stick to `padding-right: 3.5rem;`then

zwliew:
> > However, there is 1 issue regarding the hover state for tall buttons: in the Figma design, the on-hover background-color of the buttons is the same as the background-color of the modal (#f5f6fc). This makes it pretty difficult to visually identify the hover state.
> > Perhaps a darker shade of grey should be used for this specific scenario instead?
> 
> Could I trouble you to check with @NatMaeTan on which color would be suitable?

She didn't reply to me :( I'll just change the background color to match the Figma design.

sarahsalim:
@zwliew as discussed, we'll change the hover state background to 50% of #B5C4FF (pale blue). Here's a screenshot of the updated Design System. Thanks!

Screenshot 2021-02-25 at 5 31 19 PM

frontend/src/components/dashboard/demo/create-demo-modal/CreateDemoModal.module.scss
```diff
@@ -42,14 +35,21 @@ $modal-width: $tablet;
...
 
       &:not(.active) {
         color: $primary;
-        background: $neutral-light;
+        background: transparent;
 
         &:hover {
-          background: $primary;
-          color: white;
+          background: rgba(181, 196, 255, 0.5);
```
miazima, line 49: Could we rewrite this as `rgba($blue-100, 0.5)`so that we can infer the color from our variables

miazima (approved):
Lgtm!

frontend/src/components/dashboard/demo/create-demo-modal/CreateDemoModal.module.scss
zwliew: I don't think it's possible to write it that way. Perhaps we should define a `$blue-100-translucent` variable in `src/styles/base/_colours.scss` instead?

zwliew: Ahh, disregard what I said. I didn't know that that was a SASS feature.

PR #988, opened 2021-02-15
feat: add lambda to clean up orphaned secrets

### LINKED ISSUE #956: Clean up orphan credentials in Secrets Manager
Credentials are currently never deleted even when a user disassociate it with their account. These credentials cost us money to store in Secrets Manager.

- Identify the credentials that are:
  - Not associated to any user in the database (found in `credentials` but not in `user_credentials`)
  - All campaigns using the credentials have already been sent 
  - Exists in Secrets Manager but not in the database
- Delete these credentials programmatically

lamkeewei:
## Read me first
For testing locally, **USE LOCALSTACK** to avoid accidentally deleting credentials in the actual AWS Secrets Manager.

In addition, while the PR is being reviewed, I have only granted read-only permissions (for both database and Secrets Manager to the lambda) to prevent any potential bug from accidentally deleting secrets. Once we are sure that this script function as expected, do the following to grant the lambda write permissions
- `GRANT DELETE ON credentials TO serverless_cleanup_secrets_iam;`
- Change Secrets Manager permission for serverless role to RW

## Problem

Over time, secrets might become orphaned after users disassociate a credential from their user account. The secrets are still stored in Secrets Manager and continue to incur cost.

Closes #956 

Note that this PR does not clean up secrets that exists in secrets manager but not the database. The main difficulty for this is that not all our secrets are tagged with the environment that they belong to. As such to figure out which secrets are safe to delete, we will have to compare against both staging and production database. This will likely have to be done manually and can be pursued separately if required.

## Solution

**Features**:

- Add a lambda that can be triggered manually to clean up secrets. 
- Lambda can also be ran with dry run to allow manual verification before actual deletion occurs.

## Tests

**Dry run flag**
- Run the lambda with both example events in `examples/` folder. Ensure that unless `confirmRemoval` is set explicitly to `true`, actual deletion wouldn't occur. 

**Orphaned credentials**
- Add a new credential (Telegram/Twilio)
- Send a campaign
- Run lambda to check that credential is not being removed
- Remove credential under settings
- Re-run lambda and see that the credential is being removed from both the table and Secrets Manager

**Orphaned credentials with unsent campaign**
- Add a new credential (Telegram/Twilio)
- Create a new campaign using the credential but do not send yet
- Remove credential under settings
- Run lambda to check that credential is not being removed

## Deploy Notes
**New environment variables**
The following environment variables have already been set in Travis
- Staging:
   - `STAGING_CLEAN_SECRETS_ROLE`
   - `STAGING_CLEAN_SECRETS_DB_URI`
- Production:
   - `PRODUCTION_CLEAN_SECRETS_ROLE`
   - `PRODUCTION_CLEAN_SECRETS_DB_URI`

**New scripts**
- Create database user for `cleanup-secrets` in production after release
```sql
CREATE USER serverless_cleanup_secrets_iam;
GRANT rds_iam TO serverless_cleanup_secrets_iam;

GRANT SELECT, DELETE ON credentials TO serverless_cleanup_secrets_iam;
GRANT SELECT ON user_credentials TO serverless_cleanup_secrets_iam;
GRANT SELECT ON statistics TO serverless_cleanup_secrets_iam;
GRANT SELECT ON users TO serverless_cleanup_secrets_iam;
GRANT SELECT ON campaigns TO serverless_cleanup_secrets_iam;
```

## TODOs
- [x] Add check to ensure that no unsent campaigns are using the credential
- [x] Create IAM users for staging and production
- [x] Create database user
- [x] Add lambda deployment to travis
- [x] Test dry run on staging


lamkeewei:
Once approved, will merge this in next week so that we have sufficient time to test in staging first. Wouldn't be releasing this for this week's release.

serverless/cleanup-secrets/src/cleanup.ts
```diff
@@ -0,0 +1,124 @@
...
+  // 2. Does not belong to an user
+  // 3. Not used in an unsent campaign
+  const results = await sequelize.query(`
+    SELECT
+      name
+    FROM
+      credentials
+      LEFT OUTER JOIN user_credentials ON credentials.name = user_credentials.cred_name
+    WHERE
+      LOWER(credentials.name) NOT SIMILAR TO '%_(default|demo)'
```
miazima, line 105: Postgres pattern matching docs state that % matches 0 or more characters and _ matches a single character, did you mean to escape the underscore here?

lamkeewei: Good catch! Will fix this.

lamkeewei:
While re-testing this, I noticed an issue. 

If a credential in Secrets Manager have been marked for deletion by the cleanup lambda, a user adding the same credential again in a 7 days period will have an error. This is due to the fact that Secrets Manager don't immediately delete the secret. So while trying to add the secret again within the 7 days period, an error will be thrown as the secret has already been marked for deletion. 

To reproduce:
1. Create a new Telegram/Twilio credential under settings
2. Delete the credentials under settings
3. Run cleanup (`dryRun` should be `true`). The credential should be deleted from both Secrets Manager and the database
4. Attempt to add and validate the same credential again in the settings page
5. An error should be thrown

Will think a little on how to fix this.

serverless/cleanup-secrets/src/cleanup.ts
```diff
@@ -0,0 +1,150 @@
...
+  // 3. Not used in an unsent campaign
+  const results = await sequelize.query(
+    `
+    SELECT
+      name
+    FROM
+      credentials
+      LEFT OUTER JOIN user_credentials ON credentials.name = user_credentials.cred_name
+    WHERE
+      LOWER(credentials.name) NOT SIMILAR TO '%\\_(default|demo)'
```
jeantanzj, line 129: When I ran this query using TablePlus, it still returned the demo / default query. It looks like this line should have been  `LOWER(credentials.name) NOT SIMILAR TO '%_(default|demo)'`

jeantanzj:
@lamkeewei  Would this PR be better off as two manual scripts:
1) tags secrets as staging or production
2) removes unused secrets (as in this PR) 

I'm wary that deploying this lambda (as a job) adds unnecessary complexity / scenarios where a bug accidentally deletes secrets, and so would rather not run it as frequently, but perhaps only when someone manually triggers it. 

It works, except for the sql query, and I leave it up to your discretion whether to deploy it or not.

jeantanzj (changes requested):
as above

lamkeewei:
> @lamkeewei Would this PR be better off as two manual scripts:
> 
> 1. tags secrets as staging or production
> 2. removes unused secrets (as in this PR)
> 
> I'm wary that deploying this lambda (as a job) adds unnecessary complexity / scenarios where a bug accidentally deletes secrets, and so would rather not run it as frequently, but perhaps only when someone manually triggers it.
> 
> It works, except for the sql query, and I leave it up to your discretion whether to deploy it or not.

Yep. Agree with you. Running this automatically feels risky. Moreover, once lambdas are started, there's no way to cancel it. Let me make this into manual scripts and add some docs on how to run this. Will also add the script to tag the secrets.

jeantanzj:
Oh and with regard to your comment on deleted secrets being readded in 7 days - perhaps we can set up a Sentry alert when this exception is thrown? I think the probability of this happened is pretty low so may not be worth it to automatically restore a secret (https://docs.aws.amazon.com/secretsmanager/latest/userguide/manage_delete-restore-secret.html)

jeantanzj:
Waiting on the change for the sql query otherwise lgtm

PR #989, opened 2021-02-16
chore(lingui): Update to lingui v3+

### LINKED ISSUE #864: Resolve warnings to update lingui to remove use of React Legacy API
**Describe the bug**
A warning is thrown by React that the `lingui` package is using a legacy api version of react. This legacy api will not be supported in the next major release of react. We should figure out if we can transition `lingui` from this version.

image

**To Reproduce**

1. Load Postman landing/dashboard page and observe warnings in console

zwliew:
Notes:
- Remove @types/lingui packages as Lingui has native TypeScript support now.

## Problem

We're currently using Lingui v2+, which uses the legacy version of the React Context API. Lingui v3+ uses the modern API instead.

Closes #864

## Solution

- Updated Lingui to the latest v3+ version (v3.5.1).
- Removed @types/lingui dependencies as Lingui v3+ has native TypeScript types.
- Updated the `.linguirc` config to use the new config options, but with no functionality change. (see: https://lingui.js.org/ref/conf.html)
- Used the new `i18n` API in `src/config.ts` (see: https://lingui.js.org/releases/migration-3.html#lingui-core)
- Replaced the use of the old and removed `i18n._(...)` API. Used the `t` macro instead. (see: https://lingui.js.org/ref/macro.html#examples-of-js-macros)

## Tests

Re-extract and re-compile the translations with `npm run extract` and `npm run compile`.

- When running the frontend, ensure that no errors appear in stdout or in the developer console.
- All the text in the frontend should be translated as written in `messages.po`.
- Any change in the translations in `messages.po` should be reflected in the frontend after re-compiling.

**Changed dependencies**:

- @lingui/core, @lingui/react, @lingui/macro from v2.9.1 -> v3.5.1
- Removed @types/lingui__core, @types/lingui__react, @types/lingui__macro


lamkeewei:
@zwliew can you rebase and resolve conflict before I review? Thanks!

zwliew:
> @zwliew can you rebase and resolve conflict before I review? Thanks!

Done. Thanks for reviewing!

frontend/src/locales/index.tsx
```diff
@@ -1,15 +1,13 @@
...
+ * Add it to the `locales` array in `.linguirc` and run `npm run extract`
+ */
 
 import catalogEn from './en/messages.js'
-const catalogs: Catalogs = { en: catalogEn }
-export const i18n: I18n = setupI18n({
-  language: 'en', // language to display for the site. Currently this is hard-coded and cannot be changed via UI.
-  catalogs,
-})
+const catalogs: Record<string, Record<string, string>> = {
```
zwliew, line 9: I can't find an importable type declaration for Catalogs

lamkeewei: Seems to be the case. The current declaration is fine.

frontend/src/locales/index.tsx
```diff
@@ -1,15 +1,13 @@
...
+  en: catalogEn.messages,
+}
+i18n.load(catalogs)
```
lamkeewei, line 12:
```suggestion
i18n.load('en', catalogEn.messages)
```

How about this as suggested by lingui's [migration guide](https://lingui.js.org/releases/migration-3.html#i18n-load-loads-a-catalog-for-a-single-locale)? Though I think it doesn't really make a difference given we only have one language.

frontend/.linguirc
```diff
@@ -1,6 +1,10 @@
 {
-   "localeDir": "src/locales/",
-   "srcPathDirs": ["src/"],
+   "catalogs": [{
```
lamkeewei, line 2: Minor nitpick. How about using `lingui.config.ts` instead, so that our formatter/linter can pick it up.

frontend/src/locales/index.tsx
zwliew: I used `i18n.load(catalogs)` in case there were ever other languages since I reckon it'd be cleaner that way. It doesn't matter though so I'll change it.

frontend/.linguirc
zwliew: Sounds good

zwliew: I kept encountering the following error when using `export default { ... }` in `lingui.config.ts`:
```
/Users/zwliew/stuff/repos/postmangovsg/frontend/node_modules/@lingui/conf/index.js:461
    throw require("@endemolshinegroup/cosmiconfig-typescript-loader/dist/Errors/TypeScriptCompileError").fromError(error);
                                                                                                         ^
TypeError: require(...).fromError is not a function
    at loader (/Users/zwliew/stuff/repos/postmangovsg/frontend/node_modules/@lingui/conf/index.js:461:106)
    at ExplorerSync.loadFileContentSync (/Users/zwliew/stuff/repos/postmangovsg/frontend/node_modules/@lingui/conf/node_modules/cosmiconfig/src/ExplorerSync.ts:80:26)
    at ExplorerSync.createCosmiconfigResultSync (/Users/zwliew/stuff/repos/postmangovsg/frontend/node_modules/@lingui/conf/node_modules/cosmiconfig/src/ExplorerSync.ts:89:30)
    at ExplorerSync.loadSearchPlaceSync (/Users/zwliew/stuff/repos/postmangovsg/frontend/node_modules/@lingui/conf/node_modules/cosmiconfig/src/ExplorerSync.ts:64:25)
    at ExplorerSync.searchDirectorySync (/Users/zwliew/stuff/repos/postmangovsg/frontend/node_modules/@lingui/conf/node_modules/cosmiconfig/src/ExplorerSync.ts:49:32)
    at fn (/Users/zwliew/stuff/repos/postmangovsg/frontend/node_modules/@lingui/conf/node_modules/cosmiconfig/src/ExplorerSync.ts:28:27)
    at cacheWrapperSync (/Users/zwliew/stuff/repos/postmangovsg/frontend/node_modules/@lingui/conf/node_modules/cosmiconfig/src/cacheWrapper.ts:28:18)
    at ExplorerSync.searchFromDirectorySync (/Users/zwliew/stuff/repos/postmangovsg/frontend/node_modules/@lingui/conf/node_modules/cosmiconfig/src/ExplorerSync.ts:41:14)
    at ExplorerSync.searchSync (/Users/zwliew/stuff/repos/postmangovsg/frontend/node_modules/@lingui/conf/node_modules/cosmiconfig/src/ExplorerSync.ts:19:25)
    at getConfig (/Users/zwliew/stuff/repos/postmangovsg/frontend/node_modules/@lingui/conf/index.js:94:92)
```

I switched to using `module.exports = { ... }` in `lingui.config.js` and it worked fine. However, during pre-commit, ESLint keeps complaining about `module` not being defined. This is because the `node` environment isn't defined in the ESLint config (https://eslint.org/docs/user-guide/configuring/language-options#specifying-environments).

Should I add the `node` environment to the config? It doesn't sound very right since technically the frontend isn't a node environment.

I think it'd be better to just leave the config as `.linguirrc`.

lamkeewei: Hmm...ok. let's leave it as `.linguirrc`

frontend/src/locales/index.tsx
```diff
@@ -1,15 +1,10 @@
...
+ * Add it to the `locales` array in `.linguirc` and run `npm run extract`
+ */
 
 import catalogEn from './en/messages.js'
-const catalogs: Catalogs = { en: catalogEn }
-export const i18n: I18n = setupI18n({
-  language: 'en', // language to display for the site. Currently this is hard-coded and cannot be changed via UI.
-  catalogs,
-})
+i18n.load('en', catalogEn.messages)
```
lamkeewei, line 9:
```suggestion
import { en } from 'make-plural/plurals'

i18n.loadLocaleData('en', { plurals: en })
i18n.load('en', catalogEn.messages)
```

There's a warning in the console for `Locale data for locale "en" not loaded. Plurals won't work correctly.`. Although we don't use plurals (yet), we should probably load it.

zwliew: Oops, I'll properly commit this

lamkeewei (approved):
1. Check that translations are being interpolated correctly
2. Tested build on Amplify
3. Verify that warnings are resolved

PR #1011, opened 2021-02-25
feat(tesseract): extend recipients step to support Vault links

### LINKED ISSUE #1002: [Tesseract] New UI for Upload Recipients step to display Vault file option
As this feature is going to be rolled out to selective users, it is important that the **existing UI is retained and the updated UI is only shown on conditional basis**. 

- Add vault link option in Upload Recipients screen, together with file upload option
- Error messages to be shown on errors
- If file is uploaded using one option, other option should be cleared/reset
- Message preview is removed
- After the file upload step is completed, revisiting the upload recipients step should reflect the correct mode of file upload 

Design: https://www.figma.com/file/29FkTqWnA4yLKzstnvyY6X/%F0%9F%91%91Design-Master?node-id=5797%3A0

lamkeewei:
_Hosted at: https://tesseract.postman.gov.sg_

## Problem

Allow users to insert and use a Vault link to upload recipients.

Closes #1002 

## Solution

**Features**:
- Refactored CSV uploading logic across different channels to a custom hook `useUploadRecipient`
- Add `UrlUpload` component to support uploading of links
- Added recipient list type selector for campaigns
- Extend `/upload/status` endpoint to return `is_vault_link` so that frontend does not need knowledge of the vault bucket name

## Before & After Screenshots

**Upload CSV recipient list**

image

**Upload Vault link**

image

## Tests

**Hide recipient list selector for protected email campaign**
- Verify that there is no option to insert Vault after inserting Message B

**Hide recipient list selector for demo mode**
- Create a SMS/Telegram demo campaign
- Verify that there is no option to insert Vault

**Should reset csvInfo when recipient list type is changed**
- Create a new campaign
- Upload a CSV file as recipient list
- Switch to Vault recipient list type
- Preview and details should not be shown

**Upload vault link**
- For each channel:
   - Generate a pre-signed url using AWS CLI `aws s3 presign` as Vault UI is not available yet
   - Insert as recipient list
   - Ensure that recipients are added and campaign can be sent out

## TODOs
- [x] Integrate with backend when available


lamkeewei:
@sarahsalim included a minor variation from the design on Figma for the upload CSV component. I preserved the warning block as well as the previous copy writing that reminded the user that a `recipient` column is required.

**Original design**



**Implemented design**

sarahsalim:
The implemented design looks fine. Except that the font should be regular and not italics. 
Attached a screenshot for you from the Design System :)

Screenshot 2021-02-25 at 12 38 28 PM

lamkeewei:
> The implemented design looks fine. Except that the font should be regular and not italics.
> Attached a screenshot for you from the Design System :)
> 
> Screenshot 2021-02-25 at 12 38 28 PM

I realised why it's italic. It's because it is a link to the guide. Is that the expected style?

sarahsalim:
Oh... then only italicise the second sentence to be consistent with the other UI. Thanks!

frontend/src/components/dashboard/create/Create.module.scss
```diff
@@ -151,3 +151,30 @@
...
+      &:hover {
+        background-color: $blue-300;
+        color: $white;
+      }
+
+      &:not(:hover) {
+        background-color: $primary;
+        color: $white;
+      }
+    }
```
zwliew, line 178:
```suggestion
    cursor: default;

    &:.active {
      background-color: $primary;
      color: $white;
    }
    
    &:not(.active):hover {
      background-color: $primary;
      color: $white;
      cursor: pointer;
    }
```

lamkeewei: Thanks!

backend/src/core/services/campaign.service.ts
```diff
@@ -236,6 +236,10 @@ const getCampaignDetails = async (
       'demo_message_limit',
       [literal('cred_name IS NOT NULL'), 'has_credential'],
       [literal("s3_object -> 'filename'"), 'csv_filename'],
+      [
+        literal(`s3_object ->> 'bucket' = '${config.get('aws.vaultBucket')}'`),
```
miazima, line 240: I moved the config from `aws` to `tesseract` so all references to this config would be `tesseract.vaultBucket`

frontend/src/components/dashboard/create/sms/SMSRecipients.tsx
```diff
@@ -43,162 +43,189 @@ const SMSRecipients = ({
...
+            >
+              Use Vault link<i className={cx('bx', 'bx-link')}></i>
+            </PrimaryButton>
+          </div>
+        </StepSection>
+      )}
+
+      <StepSection>
+        {renderUploadInput()}
+        <ErrorBlock>{error || sampleCsvError}</ErrorBlock>
```
miazima, line 225: Is this ErrorBlock to be common between csv and url upload, it seems like maybe it belongs to the csv upload block? Since url upload has its own ErrorBlock component?

backend/src/core/services/campaign.service.ts
miazima: After a successful csv upload, and then we try switching to the vault url and input the vault url, if the vault url throws an error, the frontend would show the csv upload ui if we refresh as the bucket wasn't filled as vault yet. Which may be fine to default back to the last csv successful upload, but then the error for the vault link would have to be displayed on the vault UI. Can't see a straightforward solution to this, let's think about it!

miazima:
You may have to merge both the backend and frontend branches to test it out, or you could use the `tesseract-user-feature` branch which has the latest merged versions of frontend and backend

frontend/src/components/dashboard/create/sms/SMSRecipients.tsx
lamkeewei: `ErrorBlock` is shared across both CSV and URL upload. It's slightly confusing but the main gist is that the `ErrorBlock` within `CsvUpload` and `UrlUpload` displays the `csvError` while this `ErrorBlock` display API call error (e.g. network error, etc).

backend/src/core/services/campaign.service.ts
lamkeewei: As discussed, let's add a `temp_bucketname` field to track the type of the upload, even in the event of an error.

PR #1015, opened 2021-02-26
backend/routes/settings: Fix API doc for retrieving credentials

zwliew:
The affected API endpoint uses `channelType` as a parameter, not `campaignId`.

**Tests**
Go to the `/docs` path and check that the Swagger doc for `/settings/{channelType}/credentials` shows `channelType` instead of `campaignId` for its parameter.


zwliew:
Tested and works on staging.

lamkeewei (approved):
lgtm! Thanks for picking this up.

PR #1016, opened 2021-02-26
feat(single-send): Configure rate limiter middleware

### LINKED ISSUE #1013: [single-send] Implement rate limiting middleware
Write an express middleware for rate limiting requests for single-send messages.

*General:*
- Connect to [rate-limiter-flexible](https://github.com/animir/node-rate-limiter-flexible)
- Use Redis as the backend so that multiple backend instances can use it
- Respond with HTTP status 429 if exceeded, and a Retry-After header of 1s
- Use a standard rate limit of 10 req/s for single-send messages which is well above the current average rate of ~3 req/s.

*Email:*
Email credentials are shared between single-send messages and campaigns.
- Rate limit emails for single-send messages to 10 req/s

*SMS:*
We'll use dedicated credentials for SMS.
- Rate limit SMS for single-send messages to the default 10 req/s

zwliew:
## Problem

A rate limiter middleware needed to be added to rate limit single-send requests.

Closes #1013

## Solution

**Features**:

- Bring in the `express-rate-limit` and `rate-limit-redis` libraries.
- Configure `express-rate-limit` with a fixed window of 1 second.
- Configure `express-rate-limit` to use a Redis backend with the `rate-limit-redis` library.
- Add the `SINGLE_SEND_EMAIL_RATE` and `SINGLE_SEND_SMS_RATE` environment variables for configuring the rate limit threshold.
- Add the `REDIS_RATE_LIMIT_URI` environment variable for the rate limiter client to connect through.
 
## Tests

1. Install new dependencies and deploy the backend.
2. The new API endpoints should not accept more requests than specified by the rate limit,

## Deploy Notes

**Deploy steps**:
1. Install the new dependencies with `npm i`
2. Define the new environment variables
3. Run the backend

- `REDIS_RATE_LIMIT_URI`: Defines the Redis connection URI for the rate limiter client.
- `SINGLE_SEND_EMAIL_RATE`: Defines the rate limit threshold for single-send emails. All users shared the rate limit.
- `SINGLE_SEND_SMS_RATE`: Defines the rate limit threshold for single-send SMSes. Unique Twilio account SIDs have their own rate limits.

**New dependencies**:

- `express-rate-limit`: An Express middleware for rate limiting single-send requests.
- `rate-limit-redis`: A Redis client for `express-rate-limit`

**New dev dependencies**:

- `@types/express-rate-limit`: TypeScript definitions for `express-rate-limit`
- `@types/rate-limit-redis`: TypeScript definitions for `rate-limit-redis`


backend/src/sms/middlewares/sms-transactional.middleware.ts
```diff
@@ -48,6 +52,25 @@ async function sendMessage(
   }
 }
 
+const rateLimit = expressRateLimit({
+  store: new RedisStore({
+    prefix: 'sms-transactional:',
+    client: RedisService.rateLimitClient,
+    expiry: 1,
+  }),
+  keyGenerator(_: Request, res: Response) {
```
lamkeewei, line 61:
```suggestion
  keyGenerator: (_: Request, res: Response) => {
```

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -53,6 +57,25 @@ async function sendMessage(
   }
 }
 
+const rateLimit = expressRateLimit({
+  store: new RedisStore({
+    prefix: 'email-transactional:',
+    client: RedisService.rateLimitClient,
+    expiry: 1,
+  }),
+  keyGenerator() {
```
lamkeewei, line 66:
```suggestion
  keyGenerator: () => {
```

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -53,6 +57,25 @@ async function sendMessage(
   }
 }
 
+const rateLimit = expressRateLimit({
+  store: new RedisStore({
+    prefix: 'email-transactional:',
+    client: RedisService.rateLimitClient,
+    expiry: 1,
```
lamkeewei, line 64: Perhaps we can extract the window size to config. Makes it easier to test since we can set a longer window time and test if rate limit kicks in.

zwliew:
Thanks for the review! I've addressed the issues.

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -41,6 +45,27 @@ async function sendMessage(
...
+  keyGenerator: () => 'global',
+  windowMs: config.get('transactionalEmail.window') * 1000,
+  max: config.get('transactionalEmail.rate'),
+  draft_polli_ratelimit_headers: true,
+  message: {
+    status: 429,
+    message: 'Too many requests. Please try again later.',
+  },
+  handler: (_: Request, res: Response) => {
+    logger.warn({ message: 'Rate limited request to send transactional email' })
```
lamkeewei, line 63: Would be good to log information about the user to provide more context.

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -41,6 +45,27 @@ async function sendMessage(
   }
 }
 
+const rateLimit = expressRateLimit({
+  store: new RedisStore({
+    prefix: 'email-transactional:',
+    client: RedisService.rateLimitClient,
+    expiry: config.get('transactionalEmail.window'),
+  }),
+  keyGenerator: () => 'global',
```
lamkeewei, line 54: Just thinking, do you think it will make more sense to rate limit by user instead of imposing a system wide rate limit. 

Given the scenario with two senders, A and B. If sender A sends at a much faster rate than sender B, sender B will never have a chance to send.

zwliew: Good point. But I think a global rate limit is still a good thing to have alongside a user-level rate limit.

The user-level rate limit would prevent individual users from monopolizing the global rate limit. On the other hand, the global rate limit would prevent single-send emails as a whole from monopolizing the SES send rate.

What do you think?

lamkeewei: Yep. I think that makes sense.

zwliew: To implement this, I'll need to use a more flexible rate limiter instead, like `node-rate-limiter-flexible`. `express-rate-limit`'s API is not flexible enough.

backend/src/email/middlewares/email-transactional.middleware.ts
zwliew: I've logged user IDs. I think this is enough since we can just look up more details from the DB.

backend/src/email/middlewares/email-transactional.middleware.ts
lamkeewei: Alternatively, we can implement the global rate limit on CloudFlare. But that might make things confusing since we have rate limit at two different places. 

Another thought is that, given we have relatively small number of initial users for this. We can always ensure that `number of single send users` x `per user email rate limit` is less than the maximum number of concurrent single send we want to allow on Postman.

zwliew: I think the issue with having the global rate limit on CloudFlare is: if a sender gets past the global rate limit, but not the user rate limit, the global rate limit would still use a token even if it shouldn't.

This is the same issue with `express-rate-limit`. In general, I think that token consumption should be an all-or-nothing event -- either both the global and user rate limits use a token or, none of them do and the user is rate limited.

backend/src/core/config.ts
```diff
@@ -300,11 +300,18 @@ const config = convict({
...
+      doc:
+        'The max number of transactional emails that can be requested per window per user',
+      default: 5,
+      env: 'TRANSACTIONAL_EMAIL_USER_RATE',
+      format: 'int',
+    },
+    globalRate: {
       doc:
         'The max number of transactional emails that can be requested per window',
       default: 10,
```
lamkeewei, line 313: I think we can definitely bump this up to higher. We can send 120m/s. Most of the time the capacity is not utilised.

backend/src/email/services/email-transactional.service.ts
```diff
@@ -50,6 +54,53 @@ async function sendMessage({
...
+
+async function rateLimit(userKey: number, globalKey: string) {
+  const ACTION = 'rateLimit'
+
+  const userRateLimiterRes = await rateLimitClients.user.get(userKey)
+  const globalRateLimiterRes = await rateLimitClients.global.get(globalKey)
+  // If rateLimiterRes is null, the key has not been used before
+  const remainingUserTokens = userRateLimiterRes?.remainingPoints ?? 1
+  const remainingGlobalTokens = globalRateLimiterRes?.remainingPoints ?? 1
+  if (!remainingUserTokens || !remainingGlobalTokens) {
```
lamkeewei, line 80: I think this check does not guarantee that the `consume` will succeed later on. Assuming two requests that are processed in parallel, both of them will first see that they have remaining tokens. However, later on, only one `consume` will succeed. The get and set of the token value must be in one atomic operation. 

So this still does not solve the initial problem you had.

backend/src/core/config.ts
zwliew: I've set this to 10, but I guess we'll revise this after talking to users.

backend/src/email/services/email-transactional.service.ts
zwliew: After discussion, we decided that a global rate limit isn't that useful and is likely to have edge cases. Hence, I've change back to just using a per-user rate limit.

backend/src/core/config.ts
zwliew: I've increased this to 10, but I guess we'll increase this after understanding users' usage patterns.

lamkeewei (approved):
lgtm! Tested the following scenarios with a low rate limit threshold and works as expected:
1. Per user rate limit for transactional email
2. Per Messaging Service SID rate limit for transactional SMS

PR #1017, opened 2021-02-26
feat(single-send): Implement API endpoints

### LINKED ISSUE #1012: [single-send] Implement backend API endpoints
Create API endpoints to send single-send messages.

**Endpoint:**
- /single/email, /single/sms
- Request body:
  - recipient: string
  - body: string
  - subject: string, from: string|undefined (for email)
  - relevant credentials (for SMS)
- Connects to the rate limiter in #1013

**Errors:**
- Log all provider errors.
- On synchronous provider errors, proxy the response to the user.
- On asynchronous provider errors, trigger a CloudWatch notification.

zwliew:
## Problem

There needs to be an API endpoint for users to send single-send messages.

Closes #1012

## Solution

**Features**:

- Add 2 new API endpoints: `/transactional/sms/send` and `/transactional/email/send`
- Cache all Twilio credentials obtained from AWS Secrets Manager for 1 day in Redis.
  - A caveat is that verification of Twilio credentials also uses the cache as well, but there shouldn't be any downside to that.
- Install and configure `express-rate-limit` to use a Redis backend with the `rate-limit-redis` library.
- Add the environment variables for configuring rate limit thresholds.
- Add environment variables for the rate limiter client and credential cache client to connect to Redis.

## Tests

The following tests have been verified on staging.

**The API endpoints should respond with the appropriate status:**
- [x] Reject unauthorized requests (HTTP 401)
- [x] Reject requests with missing or extraneous parameters (HTTP 400)
- [x] Reject blacklisted email addresses (HTTP 400). Test with `suppressionlist@simulator.amazonses.com`.
- [x] Reject invalid phone numbers (done by Twilio) (HTTP 400). Test with `+6599999999`.
- [x] Reject invalid SMS body after sanitization (HTTP 400)
- [x] Reject invalid email body or subject after sanitization (HTTP 400)
- [x] If valid, send a single SMS message (HTTP 202)
- [x] If valid, send a single email message (HTTP 202)

**The API should respond with HTTP 429 when rate limited:**
- [ ] SMS: rate limit provided by the Twilio API; rate limited by `messagingServiceSid`.
- [x] Email: backend uses `express-rate-limit` middleware; rate limited by ID. Test with small `TRANSACTIONAL_EMAIL_RATE` and large `TRANSACTIONAL_EMAIL_WINDOW`

**The Twilio credential cache should try Redis cache before SecretsManager:**
- [x] Twilio credentials should be retrieved from AWS Secrets Manager the first time. (check logs for the `Retrieved secret from AWS secrets manager.` log)
- [x] The same credentials should subsequently be retrieved from Redis cache. (check logs for the absence of the aforementioned log)
- [x] The credentials should expire after `TWILIO_CREDENTIAL_MAX_AGE` milliseconds. Test with small expiry.

## Deploy notes

**New dependencies**
- `express-rate-limit`: An Express middleware for rate limiting single-send requests.
- `rate-limit-redis`: A Redis client for `express-rate-limit`

**New dev dependencies**:

- `@types/express-rate-limit`: TypeScript definitions for `express-rate-limit`
- `@types/rate-limit-redis`: TypeScript definitions for `rate-limit-redis`

**New environment variables**
- `REDIS_RATE_LIMIT_URI`: Connection URI to the Redis cache for rate limiting transactional requests
- `REDIS_CREDENTIAL_URI`: Connection URI to the Redis cache for storing credentials
- `TRANSACTIONAL_EMAIL_RATE`: The max number of transactional emails that can be requested per window per user
- `TRANSACTIONAL_EMAIL_WINDOW`: The duration, in seconds, of each window for transactional emails in seconds
- `TWILIO_CREDENTIAL_CACHE_MAX_AGE`: The maximum age, in milliseconds, of each item in the credential cache

**CloudWatch alarms (todo)**
- Added CloudWatch metric filter for HTTP 429 responses for production beanstalk instance
- Added an alarm to notify Slack on every trigger

## Todo

- [ ] Add CloudWatch filters and alarms for delivery failures
- [x] Use SNS fallback once the PR is merged

## Possible future extensions

- Track delivery reports


backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -0,0 +1,58 @@
...
+      recipient,
+      replyTo,
+    })
+    if (!messageId) {
+      errored({ next, action: ACTION, message: ERR_MESSAGE })
+      return
+    }
+    res.status(202).json({ message: 'Accepted' })
+  } catch (error) {
+    errored({ next, action: ACTION, message: ERR_MESSAGE, error })
```
lamkeewei, line 52: This returns an `HTTP 500 Internal server error` when an errors occurs. Perhaps return more meaningful HTTP error code depending on the error.

backend/src/sms/services/sms.service.ts
```diff
@@ -53,6 +53,28 @@ const getHydratedMessage = async (
...
+  credential: TwilioCredentials,
+  recipient: string,
+  message: string
+): Promise<string | void> => {
+  const twilioService = new TwilioClient(credential)
+  recipient = PhoneNumberService.normalisePhoneNumber(
+    recipient,
+    config.get('defaultCountry')
+  )
+  return twilioService.send(recipient, message)
```
lamkeewei, line 75: Same here. Make sure `message` is sanitized.

backend/src/core/services/credential.service.ts
```diff
@@ -19,6 +20,14 @@ import { UserSettings } from '@core/interfaces'
 const secretsManager = new AWS.SecretsManager(configureEndpoint(config))
 const logger = loggerWithLabel(module)
 
+const twilioCredentialCache = new LruCache<string, string>({
+  max: 1000000, // 1M chars or ~2MB
```
lamkeewei, line 24: Perhaps move these variables into `config` so that we can tune it without code changes.

lamkeewei (changes requested):

lamkeewei:
> This reduces latency by ~600ms during tests

How was this value derived? Was this tested on staging?

zwliew:
> > This reduces latency by ~600ms during tests
> 
> How was this value derived? Was this tested on staging?

My bad, this was completely wrong.

backend/src/email/services/email-transactional.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  recipient,
+  replyTo,
+}: {
+  subject: string
+  body: string
+  from: string
+  recipient: string
+  replyTo?: string
+}): Promise<void> {
+  const sanitizedSubject = EmailTemplateService.client.replaceNewLinesAndSanitize(
```
lamkeewei, line 26: Another thing we will have to handle is to check whether the `recipient` is in the `email_blacklist` table. If so, we should not even attempt to send this message as it risk increasing the bounce rate of our SES credentials. 

With that said, SES bounce notifications from single send will not be processed with this current implementation since processing [is skipped](https://github.com/opengovsg/postmangovsg/blob/e73f09603b4a8eea37d11c58aec3921bfb29a13c/backend/src/email/utils/callback/parsers/ses.ts#L115) if there is no message ID. This will need to be addressed.

zwliew: I've addressed these concerns in the latest 2 commits.

backend/src/sms/routes/sms-transactional.routes.ts
```diff
@@ -0,0 +1,61 @@
...
+ *         "429":
+ *           description: Too many requests. Please try again later.
+ *         "500":
+ *           description: Internal Server Error
+ */
+router.post(
+  '/send',
+  celebrate(sendValidator),
+  SmsMiddleware.getCredentialsFromLabel,
+  SmsTransactionalMiddleware.rateLimit,
```
lamkeewei, line 57: I was thinking, do we actually need to rate limit SMS explicitly? Can we just forward Twilio's rate limit response to the user directly?

zwliew: I think we can. We can throw a specific RateLimitError for the relevant Twilio error codes. I believe they are 20429 [1] and 21611 [2].

[1]: https://www.twilio.com/docs/api/errors/20429
[2]: https://www.twilio.com/docs/api/errors/21611

lamkeewei: Cool. I think we should do that. Given that there are actually [two mechanisms](https://www.twilio.com/blog/programmable-sms-api-basics-best-practices) at play, API rate limit and message queue. It would be easier to just forward the rate limit response back to the user than to try to replicate that on our side.

backend/src/sms/services/twilio-client.class.ts
```diff
@@ -11,40 +13,53 @@ export default class TwilioClient {
...
+        return this.send(recipient, message, true)
+      }
+
+      // 20429 - REST API rate limit exceeded
+      // 21611 - Message queue limit exceeded
+      if ([20429, 21611].includes(error.code)) {
+        throw new RateLimitError()
+      }
+
+      throw new TwilioError({
```
lamkeewei, line 58: Do we really need to introduce `TwilioError`? `error.status` will be undefined for the error thrown on line 35 and 39.

backend/src/sms/services/sms-transactional.service.ts
```diff
@@ -0,0 +1,47 @@
...
+  logger.info({
+    message: 'Sending transactional SMS',
+    action: 'sendMessage',
+  })
+  const sid = await SmsService.sendMessage(
+    credentials,
+    recipient,
+    sanitizedBody
+  )
+  if (!sid) {
```
lamkeewei, line 40: Will this line ever be reached, given that `twilioClient.send` already throws an error if `sid` is undefined.

zwliew: Ah, that's right. I'll do that.

backend/src/sms/services/twilio-client.class.ts
zwliew: Ah my bad about `error.status` being undefined.

I was originally intending for the HTTP status code to vary based on the metadata of the TwilioError.

However, on second thought, there are too many error types to accurately map to HTTP status codes. And...the implementation is broken 😅

I think I'll just return a generic HTTP 500 status code along with the error message.

backend/src/sms/services/sms-transactional.service.ts
zwliew: Resolved in https://github.com/opengovsg/postmangovsg/pull/1017/commits/689500ac36d16ac189ae5e2dc112ed3bf0d1143e

backend/src/sms/services/twilio-client.class.ts
zwliew: Reverted https://github.com/opengovsg/postmangovsg/pull/1017/commits/100c283591a26128e7b8e423bdd9a09d40362e9d

backend/src/sms/routes/sms-transactional.routes.ts
zwliew: Resolved in https://github.com/opengovsg/postmangovsg/pull/1017/commits/5d10738f9cdbf87a8e62b9b413d645d7772acf9f

backend/src/sms/middlewares/sms-transactional.middleware.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (BAD_REQUEST_ERRORS.some((errType) => err instanceof errType)) {
+      res.status(400).json({ message: err.message })
+      return
+    }
+
+    if (err instanceof RateLimitError) {
+      logger.warn({
+        message: 'Rate limited request to send transactional SMS',
+        userId: req?.session?.user.id,
+        label: credentials.label,
```
zwliew, line 44: I've decided to log just the `userId` and `label` here, since the rest of the credentials can be determined from just these 2 if needed. Also, I was a bit unsure about logging the credentials directly since they reveal somewhat sensitive information

lamkeewei: I think User ID and label are good enough!

jeantanzj:
Why do we need an LRU cache?
Would redis's  expire command be insufficient? 
https://github.com/opengovsg/postmangovsg/blob/a971bd302024e1071d644f3f68c6ac344fc1187d/backend/src/core/services/auth.service.ts#L37-L41

zwliew:
> Why do we need an LRU cache?
> Would redis's expire command be insufficient?
> https://github.com/opengovsg/postmangovsg/blob/a971bd302024e1071d644f3f68c6ac344fc1187d/backend/src/core/services/auth.service.ts#L37-L41

I think Redis would also work, but that could cause each request to incur some additional latency.

That said, I guess it would to try using Redis for now instead of introducing another dependency.

jeantanzj:
> > Why do we need an LRU cache?
> > Would redis's expire command be insufficient?
> > https://github.com/opengovsg/postmangovsg/blob/a971bd302024e1071d644f3f68c6ac344fc1187d/backend/src/core/services/auth.service.ts#L37-L41
> 
> I think Redis would also work, but that could cause each request to incur additional latency.
> 
> That said, I guess it might be better to try using Redis for now instead of introducing another dependency. Do you think we should switch to using Redis for this instead?

How were you measuring the latency

zwliew:
> > > Why do we need an LRU cache?
> > > Would redis's expire command be insufficient?
> > > https://github.com/opengovsg/postmangovsg/blob/a971bd302024e1071d644f3f68c6ac344fc1187d/backend/src/core/services/auth.service.ts#L37-L41
> > 
> > 
> > I think Redis would also work, but that could cause each request to incur additional latency.
> > That said, I guess it might be better to try using Redis for now instead of introducing another dependency. Do you think we should switch to using Redis for this instead?
> 
> How were you measuring the latency

I haven't measured it, just wondering 😅. But there wasn't a noticeable latency difference in how long the POST request took to respond when I switched over to Redis.

I'll do some actual measurements by measuring how long the `redisClient.get()` API takes to call the callback, using the perf_hooks Node.js module. I suspect there shouldn't be a significant difference though.

zwliew:
Oh wow, I should've seen this coming -- reading and writing to Redis takes at most 2ms.

jeantanzj:
isFromAddressAccepted doesn't seem like a sufficient check for whether the 'from' email can be used.  Was this expected?

```
{"level":"error","message":"Unexpected error occured","metadata":{"label":"core/loaders/express.loader","error":"Error: Error: Message failed: 554 Message rejected: Email address is not verified. The following identities failed the check in region EU-CENTRAL-1: jean@open.gov.sg\n at /usr/home/postmangovsg/src/core/services/mail-client.class.ts:69:18\n at
```

backend/src/sms/routes/sms-transactional.routes.ts
```diff
@@ -0,0 +1,60 @@
...
+ *           application/json:
+ *             schema:
+ *               required:
+ *                 - body
+ *                 - recipient
+ *                 - label
+ *               properties:
+ *                 body:
+ *                   type: string
+ *                 recipient:
```
jeantanzj, line 37: should make it clear that the recipient should contain the country code

jeantanzj: oh actually it already adds the country code, my bad

jeantanzj (approved):
Rate limit for email appears to be lower than 10/s, more like 7/s, but other than that, works fine.

PR #1019, opened 2021-02-26
feat: allow activation of SNS as fallback for all SMS campaigns

lamkeewei:
## Problem

If Twilio goes down, critical SMS campaigns cannot be sent out. We currently have no fallback mechanism. 

## Solution

**Features**:

- Add `SnsSmsClient` to send SMS using SNS
- Create new IAM policy, `postmangovsg-SNSPublishSMS` and attached to Beanstalk instance role and ECS task role
   - Configured explicit deny for SNS topic (`arn:aws:sns:*:*:*`) and allow for others. Only allow sending SMS but not publishing to other SNS topics.
- Updated incident response guide with [instructions](https://docs.google.com/document/d/1Trcqr17UZBjBoCOUKuzkk7vbbFR99QkqRobhtQ-sVIc/edit#heading=h.jme2h8uts4jm) on activating SMS fallback

## Tests
- Follow the instructions in the incident response guide and set `SMS_FALLBACK_ACTIVATE ` to `true` for `backend` and `worker`
- Send out SMS campaign
- Messages should be sent using SNS. The sender ID should either be the default value `GovSg` or the configured value.

**Credential validation tests:**
- [x] Validating a new credential should return a HTTP 400 error with a descriptive message
- [x] Validating an existing credential should return a HTTP 400 error with a descriptive message
- [x] Adding a new credential to a campaign should return a HTTP 400 error with a descriptive message

## Deploy Notes

In both backend and worker:
- `SMS_FALLBACK_ACTIVATE` : Switch to true to use SNS as fallback for all SMS campaigns
- `SMS_FALLBACK_SENDER_ID`: Sender ID to use for SNS SMSes

**New dependencies**:

- `@aws-sdk/sns-client` : AWS SNS Client

**IAM policy changes**
- Attach `postmangovsg-SNSPublishSMS` to production Beanstalk instance role
- Attach `postmangovsg-SNSPublishSMS` to production ECS task role

## TODOs
- [x] Add necessary IAM permissions to beanstalk
- [x] Update incident response doc to explain how to turn on the fallback


backend/src/sms/services/sns-sms-client.class.ts
```diff
@@ -0,0 +1,33 @@
+import { SNSClient, PublishCommand, PublishInput } from '@aws-sdk/client-sns'
+import config from '@core/config'
+
+const region = config.get('aws.awsRegion')
+
+export default class SnsSmsClient {
+  private client: SNSClient
+  private senderId: string
+
+  constructor(senderId = 'postman') {
```
jeantanzj, line 10: I'm not super sure whether we should default to Postman. will it seem like phishing if messages from say MOM suddenly started coming from Postman?

jeantanzj:
also note this caveat
https://aws.amazon.com/sns/sms-pricing/

> SMS Origination Identities
> When sending SMS using Amazon SNS, depending on the destination country, you will need to purchase and use dedicated origination identities. For more information refer to Origination identities for SMS messages in the Amazon SNS documentation. For country specific requirements when sending SMS, please refer to special requirements pages for India, and the US.
> 

Singapore is currently not one of the countries that require Origination identities.

The impact of not buying one of these identities is that smses to the foreign numbers may fail

backend/src/sms/services/sns-sms-client.class.ts
lamkeewei: Maybe we can have a fallback on the UI also to allow them to enter a sender ID.

lamkeewei: As discussed, we'll default to Gov.sg

lamkeewei:
Will tackle logging of final delivery status in a separate PR.

backend/package.json
```diff
@@ -18,6 +18,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
+    "@aws-sdk/client-sns": "^3.7.0",
     "@sentry/node": "^5.16.1",
     "async-retry": "^1.3.1",
     "aws-sdk": "^2.814.0",
```
zwliew, line 24: Since you're using the v3 SDK in this change, are there any plans to migrate from v2 to v3 entirely in the future?

lamkeewei: Good point. Guess we can open a new issue to track the effort required to move to the new V3 sdk.

zwliew (commented):
I think this works as a first pass. However, since `sendValidationMessage` and `sendCampaignMessage` are both used in validating and storing users' credentials, this may cause us to store credentials which are actually invalid.

Perhaps in the future, we should prevent users from validating credentials during the fallback? We could also skip the credential validation page when users create campaigns.

lamkeewei:
> I think this works as a first pass. However, since `sendValidationMessage` and `sendCampaignMessage` are both used in validating and storing users' credentials, this may cause us to store credentials which are actually invalid.
> 
> Perhaps in the future, we should prevent users from validating credentials during the fallback? We could also skip the credential validation page when users create campaigns.

You're right. Let me add this in for this PR.

lamkeewei:
> I think this works as a first pass. However, since `sendValidationMessage` and `sendCampaignMessage` are both used in validating and storing users' credentials, this may cause us to store credentials which are actually invalid.
> 
> 
> 
> Perhaps in the future, we should prevent users from validating credentials during the fallback? We could also skip the credential validation page when users create campaigns.

I gave some thought about this and I think I'll only prevent users from adding new Twilio credentials when fallback is activated to avoid saving potentially invalid credentials. 

However, users should still go through the flow to select their own previously validated credentials. These credentials will be set for their campaign. 

In the event, the fallback is deactivated mid-send, the workers can continue sending using the users' own credentials when the job is resumed after the workers are restarted.

zwliew (approved):
lgtm! Tested the following on staging:

1. Creating a new SMS campaign, "validating" an existing credential in it, and sending it

Expected error states:
1. Adding a new credential within a campaign
2. Adding a new credential in settings
3. Re-validating an existing credential in settings

zwliew:
To reduce code duplication, I was wondering if we could refactor the SMS clients into a separate module, which is then imported into the workers and backend. Maybe that could be explored in a future PR?

lamkeewei:
> To reduce code duplication, I was wondering if we could refactor the SMS clients (and other clients) into a separate module, which is then imported into the workers and backend. Maybe that could be explored in a future PR?

Yes, definitely. I think we should have a `shared` folder in place of the local modules. Similar to [vacops clinic registration](https://github.com/datagovsg/vaccinegovsg-clinic).

lamkeewei:
Will merge in after #1117 is merged into this

lamkeewei:
Will merge this in first. Since we can always manually reconcile, #1117 is optional.

PR #1023, opened 2021-03-01
fix: display error page if campaign id is invalid; improve layout

### LINKED ISSUE #1022: Navigating to an invalid campaign ID page leaves the UI "loading" forever
**Describe the bug**
When navigating to https://staging.postman.gov.sg/campaigns/9999123123123123123/, the UI stays "loading" forever.

Specifically, this happens in 3 cases:
1. when the logged in user doesn't have access to the specified campaign (API returns 403)
2. when the specified campaign doesn't exist, but the ID is a valid positive integer (API returns 404)
3. when the specified campaign doesn't exist and the ID is not a valid positive integer (API returns 400)

**To Reproduce**
1. Log in
2. Navigate to https://staging.postman.gov.sg/campaigns/9999123123123123123/
3. The UI should show an indefinite loading spinner.

**Expected behavior**
The website should redirect the user to the main campaigns dashboard (https://staging.postman.gov.sg/campaigns) when the specified campaign is invalid/inaccessible.

**Screenshots**
Screenshot 2021-03-01 at 3 22 18 PM

zwliew:
## Problem

The UI is stuck in the loading state whenever the user navigates to an invalid campaign page (/campaigns/<invalid campaign id>).

Closes #1022 

## Solution

**Features**
- Navigate to the error page if the campaign ID is invalid
- Center-align text on the error page
- Add navigation buttons to the previous page and to the dashboard to the error page

**Screnshots**
Screenshot 2021-03-22 at 10 13 26 AM
Screenshot 2021-03-22 at 10 13 45 AM

## Tests

1. Deploy the app.
2. Navigate to an invalid campaign URL, such as /campaigns/12378123461248124 (does not exist) or /campaigns/notapositiveinteger (not a positive integer).
3. The frontend should redirect to the main campaigns dashboard: /campaigns


frontend/src/components/dashboard/create/Create.tsx
```diff
@@ -24,16 +24,27 @@ const Create = () => {
     finishLaterContent,
   } = useContext(FinishLaterModalContext)
   const [isLoading, setLoading] = useState(true)
+  const [isInvalid, setIsInvalid] = useState(false)
 
   useEffect(() => {
-    if (!id) return
+    // Campaign ID is invalid
+    if (!id || !/^[1-9]\d*$/.test(id)) {
+      setIsInvalid(true)
```
zwliew, line 32: Another (more imperative?) way to implement this would be to use `history.replace` API, but the current method seems more declarative and React-like.

miazima:
I was wondering if it would be better user experience to display an error page like `Page not found! Please check your link again.` and maybe a button/link to return to the dashboard. So that the user has some feedback before being navigated away.

zwliew:
> I was wondering if it would be better user experience to display an error page like `Page not found! Please check your link again.` and maybe a button/link to return to the dashboard. So that the user has some feedback before being navigated away.

Good point! I think I'll add the link to the current 404 page and use that page here instead.

miazima:
lgtm code wise, I think Sarah might have some design changes to suggest

sarahsalim:
Guess linking to previous page makes sense - this will get user to try again instead of restarting the whole process all over again. 

Also I would like to use an error page graphic for this page. Will give the graphic asap this week. 

For now can remove the logo. Just have the title ‘Page not found’ without ‘!’

And can we have solid buttons instead of text buttons to make the UI look leas wordy:  ‘Back to previous page’ and ‘Go to dashboard’ 

Thanks!

sarahsalim:
We can keep both buttons:

'Go to previous page' can remain solid button. 'Go to dashboard' can be text button. Like how we did for the unsubscribe page. 

Also, can we have a bit more space between the buttons and the text above it. The spacing I saw at unsubscribe page is more accurate. Thanks!

P.S. Will try get the graphic to you asap, sorry for the delay. a bit hectic last week!

zwliew:
> We can keep both buttons:
> 
> 'Go to previous page' can remain solid button. 'Go to dashboard' can be text button. Like how we did for the unsubscribe page.
> 
> Also, can we have a bit more space between the buttons and the text above it. The spacing I saw at unsubscribe page is more accurate. Thanks!
> 
> P.S. Will try get the graphic to you asap, sorry for the delay. a bit hectic last week!

Thanks, I've updated the margins and text weights to match those of the unsubscribe page.

sarahsalim:
Graphic for error page [here](https://drive.google.com/file/d/1Ctc-OQNP80T89EjDoHW195LKcco2h3cm/view?usp=sharing) in gdrive

zwliew:
> Graphic for error page [here](https://drive.google.com/file/d/1Ctc-OQNP80T89EjDoHW195LKcco2h3cm/view?usp=sharing) in gdrive

Thanks! I've added it in.

miazima (approved):
lgtm!

PR #1027, opened 2021-03-04
chore: add robots.txt to backend

lamkeewei:
## Problem

Technically, our backend routes should not be publicly indexable. But no harm adding `robots.txt` to disallow all crawlers.

## Solution

**Improvements**:
- Add `robots.txt` to backend server

## Tests
- Visit `/robots.txt`. It should return the following:
```
User-agent: *
Disallow: /
```


lamkeewei:
Ok. Decided we should deal with the alerts instead.

PR #1030, opened 2021-03-08
Selective access to Tesseract user feature

### LINKED ISSUE #1003: [Tesseract] Implement selective feature access for users
- Add new tesseract field in user_features table to indicate if user is eligible to access the new feature
- Control access to the feature on frontend and API using user features data
- Test on multiple accounts to ensure feature is only accessible to eligible user

miazima:
## Problem

We want to feature gate Tesseract to only selected users.

Closes #1003 

## Solution

**Features**:

- Add `tesseract` boolean type column in `user_features` table
- Add middleware for user_features to handle endpoints associated with selective access
- Create sql function to whitelist users for tesseract

## Tests

User must not be able to see updated UI or access backend tesseract route 
- Test with an user who does not have an entry in user features table
- Test with user who exists in user features table and has tesseract value set to false

User is able to see updated UI and can access backend tesseract route 
- Test with user who exists in user features table and has tesseract value set to true

## Deploy Notes

- Add a new column `tesseract` in `user_features` table
```sql
ALTER TABLE user_features ADD tesseract boolean;
```
- Add sql function to whitelist tesseract users
```sql
CREATE OR REPLACE FUNCTION public.add_tesseract_user (email text)
	RETURNS void
	LANGUAGE plpgsql
	AS $function$
DECLARE
	tesseract_user_id integer;
BEGIN
	SELECT id FROM users WHERE users.email = add_tesseract_user.email INTO tesseract_user_id;
	INSERT INTO user_features (user_id, tesseract, created_at, updated_at)
		VALUES(tesseract_user_id, TRUE, clock_timestamp(), clock_timestamp()) ON CONFLICT (user_id)
		DO UPDATE SET
			tesseract = TRUE,
			updated_at = clock_timestamp();
END
$function$;
```


lamkeewei (approved):
lgtm! Tested the following:
- Flow for each channel type for non-tesseract user
- Flow for each channel type for tesseract user
- `add_tesseract_user` function

PR #1042, opened 2021-03-11
fix: allowed image sources variable not loaded correctly from lingui

lamkeewei:
## Problem

After upgrading to `lingui 3.x` in #989, we switched to importing `i18n` from `@lingui/core` as it automatically binds to the default `i18n` instance in the new version of the library. 

However, this results in the catalog not being loaded yet when we instantiate `templateClient` in `validate-csv.service.ts`. Previously, we were able to guarantee that the catalog was loaded because we referenced the exported `i18n` from `locales`. 

## Solution
**Bug Fixes**:
- Instantiate `templateClient` within the function calls

## Tests
- Insert an image tag within the Message B, referencing an image hosted on `file.go.gov.sg`
- Upload recipient list and note that no error should be thrown

### DIFF
--- a/frontend/src/services/validate-csv.service.ts
+++ b/frontend/src/services/validate-csv.service.ts
@@ -13,12 +13,14 @@ export interface ProtectedCsvInfo {
 
 export const PROTECTED_CSV_HEADERS = ['recipient', 'password']
 
-// Using default xss options for registered mail
-const templateClient = new TemplateClient({
-  allowedImageSources: i18n._(ALLOWED_IMAGE_SOURCES).split(';'),
-})
+const getTemplateClient = (): TemplateClient => {
+  return new TemplateClient({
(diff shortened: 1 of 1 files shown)


zwliew:
Perhaps it'd be better to move `import 'locales'` before all other imports instead? For example, I moved `import 'locales'` to the very top of `App.tsx` and it seems to work.

zwliew (approved):

lamkeewei:
> Perhaps it'd be better to move `import 'locales'` before all other imports instead? For example, I moved `import 'locales'` to the very top of `App.tsx`, and a`console.log(i18n._(ALLOWED_IMAGE_SOURCES))` in `validate-csv.service.ts` prints the correct value.
> 
> Other areas of the code may be broken as well because of the late import.

As spoken. Would prefer not to rely on the implicit ordering of the imports for things to work. It's easy for someone to accidentally override later on.

PR #1043, opened 2021-03-11
fix(rich-text): allow paste within table cells

lamkeewei:
## Problem

Pasting in table cells were not handled correctly previously as the default paste handler inserts a new block. 

## Solution

**Features**:

- Insert text when pasting within a table cell instead of creating a new block

## Tests
**Inserting text in table cell**
- Create a new email campaign
- Create a table
- Copy and paste text within a single cell
- Text should be pasted

**Replacing text in table cell**
- Create a new email campaign
- Create a table and enter some text within the cell
- Select text within the cell and paste text over selection
- Text should replaced by pasted text

**Multiple cell selected**
- Create a new email campaign
- Create a table with multiple cells
- Select multiple cells and paste text 
- Text should only be pasted in the first cell

### DIFF
--- a/frontend/src/components/common/rich-text-editor/RichTextEditor.tsx
+++ b/frontend/src/components/common/rich-text-editor/RichTextEditor.tsx
@@ -8,6 +8,7 @@ import {
   convertToRaw,
   DefaultDraftBlockRenderMap,
   RichUtils,
+  Modifier,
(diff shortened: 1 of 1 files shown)


zwliew (approved):

zwliew:
lgtm and also works fine for me (locally)!

PR #1046, opened 2021-03-14
fix(rich-text): line breaks not preserved and line breaks within blocks

### LINKED ISSUE #1044: Rich text editor issues
**Describe the bug**

cannot copy paste text into table cells.

text deletion occasionally doesnt work.

line spacing between paragraphs doesn't work - defaults to single line spacing.

**To Reproduce**

table copy paste:
try to copy paste text into a cell or bunch of cells

buggy backspace:
type a few paragraphs
highlight a chunk and copy paste it
try to delete it

lamkeewei:
## Problem

Address two issues:
1. Line breaks not preserved within table cells
2. Not able to insert line break without starting a new paragraph

Closes #1044 

## Solution

**Bug Fixes**:
- Replaced `\n` with `<br />` when converting to HTML so that line breaks are preserved
- Allow insertion of soft new line without starting a new paragraph when shift + enter is pressed 

## Tests
**Line breaks in table**
- Create table and enter text with line breaks within
- Send test message and campaign
- Line breaks should be preserved in the table cell

**Insert line break within paragraph**
- Insert a line of text and insert new line with `Shift + Enter`
- Line break should be entered without a new paragraph created
- Repeat and test for each type of block types (`Title`, `Subtitle`, `Header`)

### DIFF
--- a/frontend/src/components/common/rich-text-editor/RichTextEditor.tsx
+++ b/frontend/src/components/common/rich-text-editor/RichTextEditor.tsx
@@ -10,6 +10,7 @@ import {
   RichUtils,
   Modifier,
   SelectionState,
+  KeyBindingUtil,
 } from 'draft-js'
 import { Editor } from 'react-draft-wysiwyg'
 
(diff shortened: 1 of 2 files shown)


zwliew (approved):

zwliew:
lgtm and also works fine for me (locally)!

PR #1061, opened 2021-03-17
chore(frontend): upgrade react-scripts to v4.0.3

zwliew:
## Problem

react-scripts was outdated at v3.4.3 and snyk keeps prompting to update it.

## Solution

The main goal of this PR is to upgrade react-scripts and make minimal changes to address all resulting linting errors and warnings. Further upgrades to other linter plugins can be revisited in the future.

**Features**:

- Upgrade `react-scripts` to v4.0.3
- Upgrade `eslint` to v7.22.0 since react-scripts now uses v7+
- Upgrade `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` to v4.18.0 since react-scripts now uses v4+
- Remove `camelcase` rule which has been removed in ESLint v7
- Remove `@typescript-eslint/eslint-recommended` ruleset which is now redundant
- Fix new errors and warnings due to new/updated `ban-types` and `no-unused-vars` rules
- Disable new/updated rules where they make sense

react-scripts migration doc: https://github.com/facebook/create-react-app/blob/master/CHANGELOG.md#breaking-changes
ESLint migration doc: https://eslint.org/docs/user-guide/migrating-to-7.0.0
typescript-eslint release notes: https://github.com/typescript-eslint/typescript-eslint/releases/tag/v3.0.0 and https://github.com/typescript-eslint/typescript-eslint/releases/tag/v4.0.0

## Tests
The following tests have been verified on Amplify on the staging endpoint.

- Run the linter and the build. No errors or warnings should appear.
- Conduct general smoke tests
  - Creating and sending email campaign
  - Creating and sending SMS campaign
  - Creating and sending Telegram campaign
  - Creating and sending protected email campaign
  - Viewing and decrypting protected email page
  - Unsubscribing

## Deploy Notes

**New dependencies**:

- `react-scripts` from v3.4.3 -> v4.0.3

**New dev dependencies**:

- `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` from v2.26.0 -> v4.18.0
- `eslint` from v6.8.0 -> v7.22.0


frontend/src/components/common/file-input/FileInput.tsx
```diff
@@ -22,7 +22,7 @@ const FileInput = ({
         accept=".csv"
         disabled={disabled || isProcessing}
         onClick={(e) => ((e.target as HTMLInputElement).value = '')} // reset target value to allow selecting of new files
-        onChange={(e) => onFileSelected(e.target.files)}
+        onChange={(e) => onFileSelected(Array.from(e.target.files ?? []))}
```
zwliew, line 25: I'm a bit iffy about this...`e.target.files` has type `FileList | null` and we want to convert it to a type `File[]`.

lamkeewei: I think let's refactor `onFileSelected` to take in a `FileList` instead. I've looked through the code and `FileInput` is only used in the recipients pages.

frontend/src/components/dashboard/demo/create-demo-modal/CreateDemoModal.tsx
```diff
@@ -12,6 +12,9 @@ import { ModalContext } from 'contexts/modal.context'
 import { useHistory } from 'react-router-dom'
 import { createCampaign } from 'services/campaign.service'
 import DemoVideoModal from 'components/dashboard/demo/demo-video-modal'
+
+import type { Campaign } from 'classes/Campaign'
```
lamkeewei, line 16: Out of curiosity, what's the difference between `import` and `import type`?

zwliew: I'm not fully clear on the specifics of this, but what I understand is:

`import type` only imports the declarations for type checking, but doesn't introduce a runtime dependency on what was imported. So, after transpilation, whatever was imported via `import type` shouldn't appear in the code.

There's a short doc on TypeScript's website on it as well: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html

frontend/src/components/common/file-input/FileInput.tsx
zwliew: Good point -- I'll do that. Also, `e.target.files` may be `null` as well. I think I'll do a null check on this line instead of in the `onFileSelected` callbacks, since it doesn't really make sense for a caller to want to do anything with a `null` parameter.

frontend/src/components/dashboard/demo/create-demo-modal/CreateDemoModal.tsx
lamkeewei: Makes sense. Guess it should have _some_ effect on the bundle size. Was originally concerned with the following from happening:

```ts
import type { Campaign } from 'classes/Campaign'
...
const campaign = new Campaign(...)
```

But the compiler will catch this and recognize that only the type was imported.

lamkeewei (approved):
lgtm! I've tested the different user flows for each channels and it seems fine.

PR #1062, opened 2021-03-17
Log email bounce and complaint subtype

### LINKED ISSUE #1033: Log bounce subtype metadata to easier debug reason for blacklisting
Currently, all emails with `bounceType` equals to `Permanent` are added to the blacklist. The specific reason for the permanent bounce is not captured as `bounceSubType` is not logged. This prevents us from knowing the actual reason for an email being blacklisted. 

This information is useful when we are trying to remove an email from the blacklist. For e.g. if the `bounceSubType` is `Suppressed`, we know that we will have to remove the email address from both our internal `email_blacklist` table and AWS SES's global suppression list. 

Log additional `bounceSubtype` information when adding to the blacklist. Refer [here](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html#bounce-types) for the full list of bounce subtypes.

miazima:
## Problem

It would be useful to log the bounce/complaint subtype for email bounces to understand the reason behind the bounce. This also allows us to know if a blacklisted email is part of SES's global suppression list, in the case we need to remove a blacklisted email.

Closes #1033 

## Solution

**Improvements**:

- Add a new column in `email_messages` to log error sub types
- Log the email bounce subtype and complain subtype

## Tests

- Ensure the following recipients do not exist in the `email_blacklist` table
- Send an email campaign to the recipients
- Check the db for the `error_sub_type` of the email messages sent

```
recipient

// SES
bounce@simulator.amazonses.com
complaint@simulator.amazonses.com
suppressionlist@simulator.amazonses.com

// SendGrid
bouncetest@tribulant.com
bounce-test@service.socketlabs.com
```

Documentation on these mailbox simulator addresses can be found [here](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/send-email-simulator.html)

**Note**
I expected a different error_sub_type for `suppressionlist@simulator.amazonses.com` but it was classified as `General` too

## Deploy Notes

- Create a new column `error_sub_type` of type `varchar(255)` and is nullable to the `email_messages` and `email_ops` tables

**Migration query:**
```sql
ALTER TABLE email_messages
  ADD error_sub_type VARCHAR(255);

ALTER TABLE email_ops
  ADD error_sub_type VARCHAR(255);
```


zwliew:
I asked Kee Wei about this, and he mentioned that:
1. the `email_ops` table needs to mirror the `email_messages` table
2. the logger worker needs to account for the new column when logging from the `email_ops` table back to the `email_messages` table (I think the relevant query is in `log-job-email.sql`)

Also, I thought that maybe sometime in the future, we should have DB schema migration process like detailed in [this doc](https://sequelize.org/master/manual/migrations.html)? It should make modifying the DB schema less error-prone.

That said, Kee Wei also asked to consider whether we need to add the new column.

zwliew:
I considered maybe modifying the `error_code` column to include the bounce subtype for hard bounces. Something like `Hard bounce: Suppressed`. However, this should make querying for hard bounces slower.

zwliew:
Also, maybe sometime in the future, we should have a DB schema migration process like detailed in [this doc](https://sequelize.org/master/manual/migrations.html)? It should make modifying the DB schema less error-prone, and make mistakes more reversible.

lamkeewei:
> Also, maybe sometime in the future, we should have a DB schema migration process like detailed in [this doc](https://sequelize.org/master/manual/migrations.html)? It should make modifying the DB schema less error-prone, and make mistakes more reversible.

@zwliew good suggestion. Could you open an issue for this? We recently did this for CheckFirst. Can reference this [issue](https://github.com/opengovsg/checkfirst/issues/282).

zwliew:
This works fine.

However, I had another look at the original issue (#1033) and am wondering of an alternative solution: to log the error subtype in the `email_blacklist` table when adding an email to it.

Since the issue is that we want to know why an email address was blacklisted, I think this solution would more directly solve the issue while using less memory and operations.

What do you think?

jeantanzj (approved):
Tested on staging. Works as expected, hopefully the subtypes will be be more informative.

PR #1064, opened 2021-03-18
Render rich text link popover in top level html

### LINKED ISSUE #978: Prevent horizontal scroll from editor tooltip
Particularly long link will cause horizontal scroll for text area.

- Use [React createPortal](https://reactjs.org/docs/portals.html) to attach the tooltip as a sibling to the top level rich text editor `div`

miazima:
## Problem

When linking long links in the rich text editor, the link popover will overflow the textarea and introduced horizontal scroll to the textarea.

image

Closes #978 

## Solution

**Improvements**:

- Use [ReactDOM createPortal](https://reactjs.org/docs/portals.html) to attach the tooltip as a child of html body tag

I chose to attach the popover element to the html body tag instead of the top level rich text editor div as
1. relative and absolute positioning from the text block to popover does not work anymore (since they are not parent-child anymore)
2. We have to calculate the position of the text block relative to the top level rich text editor div in order to set the position of the popover div
3. Hence I felt it was more straightforward to calculate the position of the text block relative to the browser window

## Before & After Screenshots

**AFTER**:
image

## Tests

Add long links to text in rich text editor (especially to text situated at the right/left most side of the editor). Observe the position of the rendered popover.

### DIFF
--- a/frontend/src/components/common/rich-text-editor/RichTextEditor.module.scss
+++ b/frontend/src/components/common/rich-text-editor/RichTextEditor.module.scss
@@ -254,28 +254,6 @@
     }
   }
 
-  .popover {
-    @extend %caption;
-    position: absolute;
-    z-index: 999;
-    background: white;
-    width: max-content;
-    border: solid 1px $neutral;
-    box-shadow: 0px 0px 10px rgba(193, 199, 205, 0.5);
-    border-radius: 5px;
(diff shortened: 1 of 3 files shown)


lamkeewei (approved):
lgtm! Thanks for picking this up! Added ellipsis for long links to prevent overly large popovers.

image

Issue #1083, opened 2021-03-24
react-lottie is abandoned and uses deprecated lifecycle methods


zwliew:
**Describe the bug**
[react-lottie](https://github.com/chenqingspring/react-lottie) is the library we use to render the landing page animation. It appears to be abandoned and uses deprecated lifecycle methods (`componentWillUpdate`).

This is an issue because the lifecycle method will be removed in future React versions.

**To Reproduce**

1. Run the frontend in development mode locally
2. There should be a console warning (see: https://github.com/chenqingspring/react-lottie/issues/119#issuecomment-770424147)

**Expected behavior**
There should be no console warning.

**Possible Solution**
We can migrate to the [lottie-react](https://github.com/Gamote/lottie-react) library. It fixes the issue and is expected to have React 17 support [soon](https://github.com/Gamote/lottie-react/issues/29). Furthermore, the library is written in TypeScript and seems to have better code quality compared to other alternatives like [react-lottie-player](https://github.com/mifi/react-lottie-player).

zwliew:
The foremost motivation behind this issue is that during tests, react-testing-library spews a warning whenever `componentWillUpdate` is used.

zwliew:
Fix merged

PR #1084, opened 2021-03-24
chore(lottie): migrate to lottie-react library

zwliew:
## Problem

During tests and development, React warns about using deprecated React lifecycle methods (`componentWillUpdate`). This is due to the `react-lottie` library which is abandoned and still uses the deprecated method [[1]]. 

This is fine for now, but mildly annoying. Furthermore, the method will be removed in future React versions.

A fix would be to migrate to a more well-maintained library like `lottie-react`.

Closes issue #1083 

## Solution

Migrate to [lottie-react](https://github.com/Gamote/lottie-react) which is written in TypeScript and has generally better code quality. It is also expected to have React 17 support soon [[2]].

I considered [react-lottie-player](https://github.com/mifi/react-lottie-player) as well, but decided upon `lottie-react` since it is written in TypeScript.

[1]: https://github.com/chenqingspring/react-lottie/issues/119
[2]: https://github.com/Gamote/lottie-react/issues/29

**Improvements**:

- Removed `react-lottie` and `@types/react-lottie`
- Installed `lottie-react`

## Tests

These tests have been verified on Amplify on the `chore-migrate-react-lottie` branch, but not on the `staging` endpoint.

1. Deploy the change on staging.
2. Visit the landing page
3. The landing animation should play just fine with no errors/warning in the console.

Lottie isn't used anywhere else, so this is the only test needed.

## Deploy Notes

**New dependencies**:

- `lottie-react`

**Removed dependencies**:

- `react-lottie`
- `@types/react-lottie`


zwliew:
The foremost motivation behind this issue is that during tests, React spews a warning whenever `componentWillUpdate` is run.

I considered silencing the warnings by mocking them, but it sounds unnecessary given that the fix is so simple.

lamkeewei (approved):
lgtm! Thanks for fixing this. Checked and the bird still flies on both desktop and mobile.

Issue #1078, opened 2021-03-23
Set up frontend test fixtures and deployment processes


zwliew:
**Is your feature request related to a problem? Please describe.**
We want to start testing important workflows in the frontend.

**Describe the solution you'd like**
- Set up test fixtures
- Run the tests during deployment on Amplify and Travis

PR #1086, opened 2021-03-24
feat(frontend): set up test fixtures and deployment processes

zwliew:
## Problem

We want to start testing the frontend.

Closes issue #1078

## Solution

**Features**:

- Mock the Canvas API for Lottie
- Install a shim for the WebCrypto API since jsdom doesn't implement it
- Start listening on the `mswjs` server before all tests
- Use fake Jest timers
- Mock the `muted` property for `HTMLMediaElement` to workaround a React issue
- Pass in optional parameters to configure `react-ga`
- Add script to run all frontend tests during deployment on Amplify and Travis

## Before & After Screenshots

There are no visual changes.

## Tests

The tests have been verified locally, but not on staging yet.

1. Deploy the app on staging.
2. Verify that the automated tests run without errors on Travis and Amplify.
3. Conduct smoke tests to ensure that overall functionality is unaffected.

- [x] Write a simple unit test for rendering the landing page when unauthenticated

## Deploy Notes

The aforementioned tests have been added and should be run during every deployment on Travis and Amplify. If any test fails, the deployment should not go through.

Specifically, Amplify should not publish the new build if any test fails.

**New scripts**:

- `test`: Runs the frontend tests. If `CI=true` is not set, the script will be interactive and auto-reloaded. Otherwise, the script will run all tests exactly once and exit. (ref: https://create-react-app.dev/docs/running-tests/#continuous-integration)

Note: Travis runs with `CI=true` by default, but Amplify doesn't.

**New dev dependencies**:

- `msw`: to mock backend API responses
- `@testing-library/react`: for idiomatic react testing APIs
- `@testing-library/jest-dom`: for idiomatic DOM assertion APIs
- `@testing-library/user-event`: for idiomatic DOM user event APIs
- `jest-canvas-mock`: to mock Canvas for Lottie as jsdom doesn't implement it
- `@peculiar/webcrypto`: as jsdom doesn't implement the WebCrypto API


frontend/src/setupTests.ts
```diff
@@ -0,0 +1,44 @@
...
+  // Workaround for issue with 'muted' property not being treated as a default attribute
+  // see: https://github.com/testing-library/react-testing-library/issues/470
+  Object.defineProperty(HTMLMediaElement.prototype, 'muted', {
+    set: jest.fn(),
+  })
+})
+
+beforeEach(() => {
+  // Fake timers using Jest
+  jest.useFakeTimers()
```
lamkeewei, line 29: Not sure about calling this at the global level. It would probably be better to make explicit calls to `jest.useFakeTimers` in the tests itself. Also there might be instances where you want to call `jest.runOnlyPendingTimers` and make assertions after the pending timers have executed.

frontend/src/services/ga.service.ts
```diff
@@ -19,9 +19,10 @@ export const GA_USER_EVENTS = {
   COMPLETE_DUPLICATE: 'Complete campaign duplication',
 }
 
-export function initializeGA() {
+export function initializeGA(options?: ReactGA.InitializeOptions) {
   ReactGA.initialize(GA_TRACKING_ID, {
```
lamkeewei, line 23: This currently throws a warning as GA_TRACKING_ID is not provided. Might be good to mock the entire ga service during tests. This way we also don't need to pass `testMode: true` to `AuthContextProvider`.

frontend/src/App.test.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import { server, rest, within, render, screen } from './test-utils'
+import App from 'App'
+
+test('renders with no errors', async () => {
```
lamkeewei, line 5: This would probably be a better test if:
1. More specific about the behaviour you're testing for. E.g. should show landing page when route is `/`.
2. Try to test for only one thing per test. So it will be better to have multiple tests for each expected route, rather than one test for all routes

The general idea is that if the test fails, you should be able to quickly pinpoint what broke. In this case, if `renders with no errors` failed, it's not immediately obvious what needs to be fixed.

lamkeewei (commented):
Left some suggestions.

frontend/src/App.test.tsx
zwliew: Thanks! I originally included this test just to ensure that tests run fine on Travis and Amplify, so I didn't put much thought into the naming. I've renamed the test to "displays landing page when unauthenticated".

About having multiple tests, I feel like there isn't really a reason to do that for this test. I think this test is simple enough to be able to pinpoint errors quickly from the console error. Furthermore, having to render the route multiple times would increase the overall running time of the test suite.

I will keep these in mind for future tests though!

frontend/src/setupTests.ts
zwliew: I've removed this line and the corresponding `jest.useRealTimers()` call. Thanks!

frontend/src/services/ga.service.ts
zwliew: I've decided to inject the options for ReactGA using an exported variable from `config.ts`. So now `initializeGA` looks like this: https://github.com/opengovsg/postmangovsg/pull/1086/commits/01718ba811a86a35309cb370d4076ffb63de37b9#diff-1791205ed2d87458629a79c784c32fb6f18ab0a797f42c8635b11b589ec72371R23

lamkeewei:
@zwliew just noticed that #940 stores tests in a separate test folder. Would be good to discuss with @miazima to standardise what sort of directory structure we will be using.

lamkeewei (approved):
lgtm as the base for frontend tests! Though I wasn't able to reproduce the error you had on Travis. 

I'll leave it up to you to decide when to merge into develop once the other tests are done.

zwliew:
I just rebased the branch atop `develop`. Then realised I probably should've merged `develop` instead to minimize confusion 😅

PR #1091, opened 2021-03-26
feat(rich-text): allow use of variables in link url

lamkeewei:
## Problem

Users might want to insert a link with variables as the URL (e.g. `{{protectedlink}}`). This is currently not possible as the `react-draft-wysiwyg` tries to coerce the link value to a valid URL.

## Solution

**Features**:
- Override the default `linkCallback` and set link value as-is if it is a variable.

## Before & After Screenshots
image
image
image

## Tests
**Insert protectedlink**
- Insert a link using the toolbar and set the value as `{{protectedlink}}`
- Validation should recognize that `{{protectedlink}}` is set and allow you to proceed to the next step in the campaign creation flow.
- On the preview step, the link should be correctly substituted by a protected email link

**Variables within link**
- Create an email campaign and pass in a variable in the link. E.g. `google.com/q?={{recipient}}`
- Send campaign and observe that the variable should be substituted correctly 

**Conversion from HTML**
- Create an email and insert a link with a variable as the value. E.g. `{{protectedlink}}`
- Proceed to upload recipient list
- Go back to the "Create message" step and check that the popover shows the correct value of the link (e.g. `{{protectedlink}}`. 

**Links without protocol**
- Create an email campaign and insert a link with url `postman.gov.sg`
- Upon insertion, the link should be automatically prefixed with `http`

**Allow use of variable as link title**
- Create an email with a variable as the link title (see above screenshot)
- Sent messages should have their links titles and urls correctly templated


miazima:
When inserting a link in email campaign with a variable as the link value e.g. `{{link}}`, the link popover shows the correct link value. However in the preview the substituted link value is not correct, and opens up the campaign page again. This affects the Conversion from HTML and Allow use of variable as link title scenario. 

The rest of the scenarios work as expected.

lamkeewei:
> When inserting a link in email campaign with a variable as the link value e.g. `{{link}}`, the link popover shows the correct link value. However in the preview the substituted link value is not correct, and opens up the campaign page again. This affects the Conversion from HTML and Allow use of variable as link title scenario.
> 
> The rest of the scenarios work as expected.

As discussed, the links provided in the CSV file has to be complete links with `http://` or `https://` prefixed. Otherwise, `xss` lib [will filter out](https://github.com/leizongmin/js-xss/blob/c52ed1264202f46bd79e6292bca1b0c05dd102ca/lib/default.js#L151) the value for `href`, resulting in an empty link that points to the current page.

miazima (approved):
lgtm! We can put in an issue to update the guide on allowed url format for variable links.

PR #1092, opened 2021-03-26
chore(frontend): upgrade typescript to v4.2.3

zwliew:
## Problem

We want to install `msw` is a useful dependency for mocking API endpoints during tests. However, `msw` requires TypeScript v4+ while we currently only use v3.9.7.

Reference: https://github.com/mswjs/msw/issues/491

## Solution

**Improvements**:

- Upgrade the `typescript` dependency from v3.9.7 to v4.2.3
- Fix a minor compilation error resulting from the upgrade

## Before & After Screenshots

There are no visual changes.

## Tests

These tests have been verified locally, but not on the Amplify staging endpoint yet.

1. Compile the app. There should be no errors.
2. Deploy the app on staging.
3. Conduct smoke tests.
4. The app should look and function exactly like before.

Smoke tests:
 - Creating and sending email campaign
 - Creating and sending SMS campaign
 - Creating and sending Telegram campaign
 - Creating and sending protected email campaign
 - Viewing and decrypting protected email page
 - Unsubscribing

## Deploy Notes

**Updated dependencies**:

- `typescript`: upgraded from v3.9.7 to v4.2.3

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -21266,9 +21266,9 @@
       "integrity": "sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c="
     },
     "typescript": {
-      "version": "3.9.7",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-3.9.7.tgz",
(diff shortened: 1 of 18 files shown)


lamkeewei (approved):
lgtm!

PR #1094, opened 2021-03-29
style: fix font size for links in protected email preview

lamkeewei:
## Problem

Links were of a different font size compared to the normal text in protected preview.

## Solution

**Bug Fixes**:
- Explicitly apply same font size for links in protected preview

## Before & After Screenshots
**Before**
image

**After**
image

## Tests
- Create and send protected email with link
- Ensure that font size is the same for both Message B preview and the protected message page


zwliew (approved):
LGTM and works fine for me locally! Can't deploy it on Amplify to test but it should be fine there too

PR #1103, opened 2021-03-31
chore: fix build warnings

zwliew:
## Problem

There are a few build warnings that come up in Amplify. These are all minor warnings which require zero functional changes.

## Solution

**Improvements**:

- Move the definition of custom headers for Amplify from `amplify.yml` to `customHttp.yml` (ref: https://docs.aws.amazon.com/amplify/latest/userguide/custom-headers.html#migrate-custom-headers)
- Explicitly specify --no-rewrite for source map uploads to Sentry

## Before & After Screenshots

There are no functional or visual changes.

## Tests

These changes have been verified on the branch deployed at https://fix-build-warnings.d3k3b7zs07zjio.amplifyapp.com/

**Test 1: Amplify custom headers should remain unchanged**
1. Deploy the change to Amplify.
2. Ensure that the following warnings no longer appear:
```
2021-03-30T01:45:46.599Z [WARNING]: [DEPRECATION WARNING] using amplify.yml to modify custom headers is being deprecated.
2021-03-30T01:45:46.599Z [WARNING]: Going forward, in order to configure your custom headers via deploy, please use customHttp.yml instead.
2021-03-30T01:45:46.599Z [WARNING]: Learn more: https://docs.aws.amazon.com/amplify/latest/userguide/custom-headers.html
```
3. Visit the landing page. (this should be the same for any page)
4. Under the network tab of Chrome devtools, look at the response headers of the initial GET request.
5. Ensure that the response headers that are specified in `customHttp.yml` are exactly the same as those in the GET request.
6. Also check the response headers for a static file like https://fix-build-warnings.d3k3b7zs07zjio.amplifyapp.com/static/js/runtime-main.f46cbea8.js

**Test 2: Sentry warning no longer appears**
2. Verify that the following build warnings no longer appear:
```
2021-03-30T01:45:40.429Z [WARNING]: WARN    2021-03-30 01:45:40.429564075 +00:00 The default --no-rewrite will disappear. Please specify --rewrite or --no-rewrite explicitly during sourcemap upload.
```


frontend/sentry.js
```diff
@@ -14,7 +14,8 @@ async function createReleaseAndUpload() {
     await cli.releases.uploadSourceMaps(release, {
       include: ['build/static/js'],
       urlPrefix: '~/static/js',
-      validate: true
+      validate: true,
+      rewrite: false,
```
zwliew, line 18: It appears that this is true by default starting v1.59.0 [1], and we are currently using v1.54.0. It may be worthwhile to either set this to true, or upgrade to a later version of `@sentry/cli`. I'll look into these options.

[1]: https://github.com/getsentry/sentry-cli/blob/master/CHANGELOG.md#sentry-cli-1590

zwliew: Oh haha , snyk upgraded `@sentry/cli` to the latest version: https://github.com/opengovsg/postmangovsg/pull/1100

zwliew:
I was testing some other changes and decided to move to a different branch: https://github.com/opengovsg/postmangovsg/commits/improve-frontend-deployment

zwliew:
I think i'll restore this branch since the other changes are not as trivial

lamkeewei:
Getting @miazima to help review this since she did the initial implementation for frontend Sentry.

frontend/sentry.js
```diff
@@ -14,7 +14,8 @@ async function createReleaseAndUpload() {
     await cli.releases.uploadSourceMaps(release, {
       include: ['build/static/js'],
       urlPrefix: '~/static/js',
-      validate: true
+      validate: true,
```
miazima, line 17: I think we can remove this, refer Sentry-cli [docs](https://docs.sentry.io/product/cli/releases/#sentry-cli-sourcemaps) on uploading source maps

Regarding the default rewrite option
> It automatically validates source maps before upload very accurately which can spot errors you would not find otherwise until an event comes in. This is an improved version of what --validate does otherwise.

Regarding the validate option 
> This attempts source map validation before upload when **rewriting is not enabled.**

zwliew: Oh I see, so we can remove both `validate: true` and `rewrite: true`? That sounds good, but would we want to keep the options simply for clarity?

zwliew: Oh I see. Alright, I've removed it.

miazima (approved):
Build warnings stated have been resolved in amplify build, lgtm!

PR #1109, opened 2021-04-01
fix: avoid calling setState when components are unmounted

zwliew:
## Problem

During tests, React logs errors like so:
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
        in Campaigns (created by Context.Consumer)
```

This is because some components are calling `setState` or its derivatives even when they are unmounted.

## Solution

**Improvements**:

- Check if the components are still mounted before calling `setState`

## Before & After Screenshots

There are no visual changes.

## Tests

These tests have been verified locally but not on the staging branch on Amplify.

**The demo and announcement modal should be unchanged**
1. Load the campaigns dashboard.
2. The demo banner, number of demo SMS campaigns, and number of demo Telegram campaigns should appear just as before.

**Uploading recipients should be unchanged**
1. Upload a valid recipients file to an SMS campaign.
2. The uploading state, loading state, and the processed state should appear in order just as before.
3. Upload an invalid recipients file to the campaign.
4. The uploading state, loading state, and error state should appear in order just as before.
5. Do the same for the Telegram campaign.

### DIFF
--- a/frontend/src/components/dashboard/create/sms/SMSRecipients.tsx
+++ b/frontend/src/components/dashboard/create/sms/SMSRecipients.tsx
@@ -127,9 +127,8 @@ const SMSRecipients = ({
       setCsvInfo((info) => ({ ...info, tempCsvFilename }))
     } catch (err) {
       setErrorMessage(err.message)
-    } finally {
-      setIsUploading(false)
     }
+    setIsUploading(false)
   }
 
   // Hide csv error from previous upload and delete from db
--- a/frontend/src/components/dashboard/create/telegram/TelegramRecipients.tsx
+++ b/frontend/src/components/dashboard/create/telegram/TelegramRecipients.tsx
@@ -124,9 +124,8 @@ const TelegramRecipients = ({
       setCsvInfo((info) => ({ ...info, tempCsvFilename }))
     } catch (err) {
       setErrorMessage(err.message)
-    } finally {
-      setIsUploading(false)
     }
+    setIsUploading(false)
   }
 
   // Hide csv error from previous upload and delete from db
--- a/frontend/src/components/dashboard/campaigns/Campaigns.tsx
+++ b/frontend/src/components/dashboard/campaigns/Campaigns.tsx
@@ -27,10 +27,12 @@ import { getUserSettings } from 'services/settings.service'
 import DuplicateCampaignModal from '../create/duplicate-campaign-modal'
 import AnnouncementModal from './announcement-modal'
 import { ANNOUNCEMENT, getAnnouncementVersion } from 'config'
+import useIsMounted from 'components/custom-hooks/use-is-mounted'
 
 const ITEMS_PER_PAGE = 10
 
 const Campaigns = () => {
+  const isMounted = useIsMounted()
   const { email } = useContext(AuthContext)
   const modalContext = useContext(ModalContext)
   const [isLoading, setLoading] = useState(true)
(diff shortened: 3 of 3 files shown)


frontend/src/components/dashboard/create/telegram/TelegramRecipients.tsx
```diff
@@ -124,9 +124,8 @@ const TelegramRecipients = ({
       setCsvInfo((info) => ({ ...info, tempCsvFilename }))
     } catch (err) {
       setErrorMessage(err.message)
-    } finally {
```
zwliew, line 127: I originally thought this would be a noop, but it isn't.

The finally block will always execute even if there is a return statement within the try block.

frontend/src/components/dashboard/create/sms/SMSRecipients.tsx
```diff
@@ -127,9 +127,8 @@ const SMSRecipients = ({
       setCsvInfo((info) => ({ ...info, tempCsvFilename }))
     } catch (err) {
       setErrorMessage(err.message)
```
lamkeewei, line 129: Should we also check if the component is still mounted before choosing to call `setErrorMessage`?

frontend/src/components/dashboard/create/sms/SMSRecipients.tsx
```diff
@@ -127,9 +127,8 @@ const SMSRecipients = ({
...
-    } finally {
-      setIsUploading(false)
     }
+    setIsUploading(false)
```
lamkeewei, line 131: I'm think we might still need to reset `uploading` to `false` even on error. Otherwise, when the S3 upload fails, the upload button will be stucked in a loading state even when there's an error (haven't tested this, but seems to be the case from reading the code).

frontend/src/components/dashboard/create/sms/SMSRecipients.tsx
zwliew: Possibly, but I haven't noticed a real issue with it so far. That said, I might run into the error when unit testing error states for this component. I'll try it out and let you know how it goes.

frontend/src/components/dashboard/create/sms/SMSRecipients.tsx
zwliew: I think `setIsUploading(false)` will still be called even on errors.

zwliew: I think `setIsUploading(false)` will be called even on error. To be sure, I tested this locally by throwing an error in the `try` block. Upon error, the error message is shown and the file input is no longer in the processing state.

lamkeewei: Ah ok. You're right. My bad.

frontend/src/components/dashboard/create/sms/SMSRecipients.tsx
lamkeewei: Sounds fair.

lamkeewei (approved):
- Uploading a recipient list and navigate before processing is completed. This shouldn't produce warnings like before.
- On dashboard page, navigate away before the announcement version and demo campaign info is loaded from backend. No warnings were triggered.

Issue #1079, opened 2021-03-23
Write frontend tests for the campaign creation workflow


zwliew:
**Is your feature request related to a problem? Please describe.**
The most important part of the frontend is the campaign creation and sending workflow. Hence, we should test those first.

**Describe the solution you'd like**
In order of decreasing priority, write tests for:
- Creating and sending email campaigns
- Creating and sending SMS campaigns
- Creating and sending Telegram campaigns
- Creating and sending protected email campaigns
- Viewing protected email
- Unsubscribing from mailing list

**Checklist**
- [ ] Look into adding code coverage reports to GitHub PRs (possibly use Codecov GitHub actions like testing-library does)
- [x] (ongoing as more tests are written) Reimplement some functionality of the backend to make tests more realistic
- [x] Write integration test for creating and sending a new email campaign
- [x] Write integration test for creating and sending a new SMS campaign
- [x] Write integration test for creating and sending a new Telegram campaign
- [x] Write integration test for creating and sending a new protected email campaign
- [x] Write unit test for viewing a protected email
- [x] Write unit test for unsubscribing from a mailing list
- Write unit tests for the following error paths:
  - SMS
    - [x] Saving an invalid template (no body)
    - [x] Uploading an invalid CSV file
    - [x] Attempting to validate an invalid credential
  - Telegram
    - [x] Saving an invalid template (no body)
    - [x] Uploading an invalid CSV file
    - [x] Attempting to validate an invalid credential
  - Email
    - [x] Saving an invalid template (no subject)
    - [x] Uploading an invalid CSV file
  - Protected email
    - [x] Saving an invalid template (no subject) (same as Email)
    - [x] Saving an invalid template (extraneous subject params)
    - [x] Saving an invalid template (missing body params)
    - [x] Saving an invalid template (extraneous body params)
    - [x] Uploading an invalid CSV file
    - [x] Viewing a protected email with an invalid ID
    - [x] Viewing a protected email with an invalid password

zwliew:
Done in #1110 and #1118

PR #1110, opened 2021-04-01
feat: add tests for happy paths in critical workflows

zwliew:
## Problem

The following critical workflows are not tested. If one of them breaks, it would be difficult to realise unless we go through the tedium of testing each one manually. Hence, let's add automated tests for them.

Workflows:
1. Creating and sending a new email campaign
2. Creating and sending a new SMS campaign
3. Creating and sending a new Telegram campaign
4. Creating and sending a new protected email campaign
5. Viewing a protected email message
6. Unsubscribing from a mailing list

Partially resolves issue #1079 

## Solution

**Features**:

- Added integration tests for happy paths in critical workflows
- Added unit tests

## Before & After Screenshots

There are no visual changes.

## Tests

These tests have been verified on Travis and Amplify.

1. Deploy the change on Travis and Amplify.
2. All the tests should be run and should pass:
- [x] Write integration test for creating and sending a new email campaign
- [x] Write integration test for creating and sending a new SMS campaign
- [x] Write integration test for creating and sending a new Telegram campaign
- [x] Write integration test for creating and sending a new protected email campaign
- [x] Write unit test for viewing a protected email
- [x] Write unit test for successfully unsubscribing from a campaign
- [x] Write unit test for user choosing to not unsubscribe

## Deploy Notes

The aforementioned tests have been added and should be run during every deployment on Travis and Amplify. If any test fails, the deployment should not go through.

Specifically, Amplify should not publish the new build if any test fails.


frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -27,10 +27,12 @@ import { getUserSettings } from 'services/settings.service'
 import DuplicateCampaignModal from '../create/duplicate-campaign-modal'
 import AnnouncementModal from './announcement-modal'
 import { ANNOUNCEMENT, getAnnouncementVersion } from 'config'
+import useIsMounted from 'components/custom-hooks/use-is-mounted'
 
 const ITEMS_PER_PAGE = 10
 
 const Campaigns = () => {
+  const isMounted = useIsMounted()
```
zwliew, line 35: This change along with. those in `SMSRecipients.tsx` and `TelegramRecipients.tsx` are explored in greater detail in https://github.com/opengovsg/postmangovsg/pull/1109

frontend/src/components/unsubscribe/tests/Unsubscribe.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    {
+      router: {
+        initialIndex: 0,
+        initialEntries: ['/unsubscribe/v1?c=1&r=testEmail@open.gov.sg&h=testH'],
+      },
+    }
+  )
+}
+
+test('successfully unsubscribes user from a campaign', async () => {
```
miazima, line 43: It would be good to also include a test on the option where the user chooses not to unsubscribe

frontend/src/components/dashboard/tests/Dashboard.test.tsx
```diff
@@ -0,0 +1,893 @@
...
+const renderDashboard = () =>
+  _render(<Dashboard />, {
+    router: { initialIndex: 0, initialEntries: ['/campaigns'] },
+  })
+
+test('successfully creates and sends a new email campaign', async () => {
+  // Setup
+  jest.useFakeTimers()
+  server.use(...mockApis())
+  renderDashboard()
```
miazima, line 56: Could this be part of beforeEach() since they are the same for all the tests?

frontend/src/test-utils/api/index.ts
```diff
@@ -1,19 +1,61 @@
...
+        id: 1,
+      },
+    ],
+    curUserId: 0, // Start unauthenticated; 1-indexed
+
+    // Campaigns
+    campaigns: [], // Start with zero campaigns
+
+    // Protected messages
+    protectedMessages: [], // Start with zero protecd messages
```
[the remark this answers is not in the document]
zwliew: Fixed, thanks!

frontend/src/components/unsubscribe/tests/Unsubscribe.test.tsx
zwliew: I've added a test for that. I'm not sure if there's much to test other than to assert that the necessary UI elements are present after clicking the "I prefer to stay" button.

frontend/src/components/dashboard/tests/Dashboard.test.tsx
zwliew: Yup, but I'm a bit hesitant in doing so, because I'm not certain if all Dashboard tests in the future will definitely use these lines for setup.

frontend/src/components/unsubscribe/tests/Unsubscribe.test.tsx
miazima: Yup that's fine, thought since we asserted that the elements are rendering correctly for the protected mail page we could do that here too.

miazima:
Since the Dashboard component is actually a collection of routes, I was thinking the campaign creation tests could fall under the respective channel Create components, e.g. EmailCreate or any other component specific to the channel.

zwliew:
> Since the Dashboard component is actually a collection of routes, I was thinking the campaign creation tests could fall under the respective channel Create components, e.g. EmailCreate or any other component specific to the channel.

Yup, we could do that. However, the issue is that there is logic in the common `Create` component (rendered from `Dashboard`) which loads the campaign from the API.

For these tests, I thought that it would be better if they were complete integration tests, which would all the components being used in the campaign creation workflow. Hence, I think it would be better if we also included the `Dashboard` and `Create` components in the tests.

Going along the same line of thought, I also considered placing these tests in the `App` component to really test the entire component pipeline. However, since `App` does very little (it only acts as a router switch), I figured it would be fine to leave it out of the integration test.

miazima:
Yup I think these tests are integration tests rather than individual component unit tests. Perhaps you can create a new folder to house integration tests similar to this?

zwliew:
> Yup I think these tests are integration tests rather than individual component unit tests. Perhaps you can create a new folder to house integration tests similar to this?

Sounds good. I'm not very sure what would be a good way to do this though. My initial thought would be to split the tests up into smaller files like this:

```
frontend/
  - src/
    - components/
      - dashboard/
        - Dashboard.tsx
        - tests/
          - integration/
            - utils.ts
            - email.test.tsx
            - sms.test.tsx
            - telegram.test.tsx
```

I feel like it is quite an inadequate solution though. How would you suggest we do this?

miazima:
I think its fine to place it as you suggested for now, I can't think of a strong reason to place it elsewhere. We can always move around our tests as needed in the future!

zwliew:
I've split the Dashboard integration tests to separate files

miazima (approved):

zwliew:
Thanks for the review!

PR #1123, opened 2021-04-12
Remove middleware with duplicated logic and handle expired urls during processing

miazima:
Refer 
https://github.com/opengovsg/postmangovsg/pull/1106#discussion_r607438380
https://github.com/opengovsg/postmangovsg/pull/1106#discussion_r607440405

### DIFF
--- a/backend/src/telegram/routes/telegram-campaign.routes.ts
+++ b/backend/src/telegram/routes/telegram-campaign.routes.ts
@@ -11,8 +11,10 @@ import {
   TelegramStatsMiddleware,
   TelegramTemplateMiddleware,
 } from '@telegram/middlewares'
+import config from '@core/config'
 
 const router = Router({ mergeParams: true })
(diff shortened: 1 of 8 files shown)


backend/src/telegram/routes/telegram-campaign.routes.ts
```diff
@@ -76,7 +76,9 @@ const tesseractCampaignValidator = {
     url: Joi.string()
       .trim()
       .custom((value: string, helpers: any) => {
-        const url = value.match(/^https:\/\/storage(-test)?.vault.gov.sg\/.+$/g)
+        const url = value.match(
+          /^https?:\/\/storage(-test)?.vault.gov.sg\/.+$/g
```
lamkeewei, line 80: Shall we make this a variable in `config` with current value as the default value? In case, there's a change we wouldn't need to update the code.

lamkeewei (approved):
lgtm!

PR #1129, opened 2021-04-13
chore: upgrade React; use new JSX transform; sort imports

zwliew:
## Problem

Let's upgrade to the new JSX transform to stay modern.

React has introduced the new JSX transform to enable improvements in future React releases. Although it does nothing significant as of now, it might be a good idea to upgrade anyway to stay up-to-date.

Reference on transitioning to the new JSX transform: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

## Solution

**Improvements**:

- Upgrade React to v16.14.0, which adds support for the new JSX transform
- Use the new JSX transform
- Uninstall TypeScript from the root directory as it causes VSCode/CoC's tsserver to use it for static analysis instead of the one in the `frontend` directory
- Convert all default React import statements to destructured imports (`import React from 'react'` -> `import { ... } from 'react'`)

## Before & After Screenshots

No visual changes.

## Tests

These tests have been verified locally but not on Amplify staging yet.

2. Conduct smoke tests:
  - [x] Create and send SMS campaign
  - [x] Create and send Telegram campaign
  - [x] Create and send email campaign
  - [x] Create and send protected email campaign
3. All pages and components should work fine without errors.

## Deploy Notes

**Upgraded dependencies**:

- `react`, `react-dom`: 16.31.1 -> 16.14.0
- `@types/react`: 16.9.29 -> 16.14.5
- `@types/react-dom`: 16.9.5 -> 16.9.12


package.json
```diff
@@ -49,7 +49,6 @@
     "husky": "^4.2.5",
     "lint-staged": "^10.2.6",
     "prettier": "^2.0.5",
-    "typescript": "^3.8.3",
```
zwliew, line 52: coc-tsserver was using this installation of typescript instead of the one in `frontend`. I don't see this being used anywhere, so I removed it.

frontend/tsconfig.json
```diff
@@ -18,11 +18,11 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react",
+    "jsx": "react-jsx",
```
zwliew, line 21: This enables the new JSX transform.

frontend/src/components/common/action-button/ActionButton.tsx
```diff
@@ -1,14 +1,16 @@
-import React, { useState } from 'react'
+import { useState } from 'react'
 import cx from 'classnames'
 
 import styles from './ActionButton.module.scss'
 
+import type { MouseEvent as ReactMouseEvent } from 'react'
```
miazima, line 6: Let's move this up top together with the react import, combine them if possible?

frontend/src/components/common/body-wrapper/BodyWrapper.tsx
```diff
@@ -1,8 +1,9 @@
-import React from 'react'
 import cx from 'classnames'
 
 import styles from './BodyWrapper.module.scss'
 
+import type { FunctionComponent } from 'react'
+
```
miazima, line 6: To follow the import order convention, external module imports should be at the top. We could also look into if we want to use this plugin to sort import order 
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#importorder-enforce-a-convention-in-module-import-order

miazima (commented):
Lgtm, just need to shift the modules import order

frontend/src/components/common/body-wrapper/BodyWrapper.tsx
zwliew: Thanks! I agree about sorting the import order with a plugin. I've looked into both `eslint-plugin-import` and `eslint-plugin-simple-import-sort` and settled on the latter since it seems to do a better job at ordering the imports out of the box.

zwliew:
I've just rebased the branch atop `develop`.

Just to clarify, eslint-plugin-import sorts imports into the following groups, with newlines between each group:
1. builtin modules (e.g. Node.js modules)
2. external modules (e.g. react)
3. parent relative models (e.g. ../xyz)
4. sibling modules (e.g. ./abc)
5. index file (e.g. ./index.ts)
6. import-mapped modules (e.g. classes, components)

miazima:
@lamkeewei I had suggested using an eslint plugin to auto sort import order for our frontend, wanted to get your go ahead too

lamkeewei:
> @lamkeewei I had suggested using an eslint plugin to auto sort import order for our frontend, wanted to get your go ahead too

Sure! Will make our imports neater. Might also be useful to think about what is a good sort order for our imports, beyond the default rules (e.g. assets should be first/last, etc). Can [refer](https://github.com/datagovsg/vaccinegovsg-clinic/blob/dev/frontend/.eslintrc) to what vaccine ops did for their sorting.

zwliew:
Rebased the branch.

Todo: look into making assets a group and moving them to the end of the imports list. This doesn't seem possible with eslint-plugin-import/order, so we will likely need to use eslint-plugin-simple-import-sort for this. vaccine ops uses it as well.

PR #1141, opened 2021-04-15
feat: support whitelisting domains through `agencies` table

### LINKED ISSUE #1115: Whitelisted email domains table
**aim**
This will facilitate user adoption by wider public sector including healthcare agencies later on without running into env var character limit and having to devise another solution later.

**update**
as of last discussion, we would move this to the db first without mapping to agencies.

lamkeewei:
## Problem

Closes #1115 

## Solution

**Features**:

- Add `Agency` model and table in database
- Validate whitelisted domain using both config domains and agencies table domains

**Improvements**:
- Log matched domain for each authentication request

## Tests
- Add email domain (e.g. `@gmail.com`) to agencies table using the following query:
```sql
INSERT INTO agencies VALUES('@gmail.com', CURRENT_TIMESTAMP, CURRENT_TIMESTAMP);
```
- Check that you will be able to request for the OTP with an email belonging to the newly inserted domain
- Delete the domain in `agencies` table
- Check that you will not be able to request for an OTP with an email belonging to the deleted domain


miazima (approved):
Tested adding and removing whitelisted agencies, lgtm

PR #1144, opened 2021-04-20
chore: replace local module with shared folder

### LINKED ISSUE #1139: Replace local modules with `shared` folder
The current local modules approach of sharing code proves hard to use. For instance, upgrade or additional of new dependencies requires [manual intervention](https://github.com/opengovsg/postmangovsg/blob/develop/docs/local-module.md). 

Use a shared folder approach to share code. Reference to [vaccingovsg-clinic](https://github.com/datagovsg/vaccinegovsg-clinic).

lamkeewei:
## Problem

Closes #1139 

### Use Typescript project references to reference shared module
I've chosen to take a slightly different approach to referencing the shared module in `tsconfig` as compared to the [vaccinegovsg-clinic](https://github.com/datagovsg/vaccinegovsg-clinic/tree/dev/frontend) repository. Instead of adding the shared folder path in the `include` option, I've chosen to make use of [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) instead. 

The main advantage for this approach is that we can avoid having to manually keep in sync the dependencies for the `shared` folder and the other modules.  For example, if we were to use the `include` approach, the `xss` library has to be in each of the `package.json` for `backend`, `frontend` and `worker`. Keeping all the different `package.json` in sync is a non-trivial problem and is prone to developer error.

The drawback of this approach is that we have to manually `npm install` inside the shared folder to install its dependencies. The necessary changes to `.travis.yml` and the respective `Dockerfile` have been made.

### Manually patch `moduleNameMapper` rather than use `aliasJest`
The default `aliasJest` function provided by [react-app-rewire-alias](https://github.com/oklas/react-app-rewire-alias) does not work for our setup. The main reason for this is that the library assumes that our aliases would follow the pattern of `module/submodule`. However, there are quite a number of instances in our code that references aliases directly without a submodule (e.g. `import config from 'config'`).

## Tests
Apart from changing the structure of the code, this PR should not modify or introduce any behaviour. To verify that everything works as expected:
1. Ensure all tests for backend and frontend passes
2. Ensure deployment succeeds without errors for backend and frontend

In addition, hot reload should also work now for the shared module. Changing code in the shared module should automatically trigger a recompilation for any other module depending on it.


.travis.yml
```diff
@@ -17,8 +17,10 @@ jobs:
         - cd $TRAVIS_BUILD_DIR/backend
         - npx lockfile-lint --type npm --path package-lock.json -o "https:" -o "file:" --allowed-hosts npm
       install:
-        - npm ci
+        - cd $TRAVIS_BUILD_DIR/shared && npm ci
+        - cd $TRAVIS_BUILD_DIR/backend && npm ci
       script:
+        - npm run build
```
miazima, line 23: Could you help explain why we didn't need the build step previously but have included it now? I guess its to build the shared module, but where was the backend being built previously

miazima (approved):
Tested the following and lgtm
- All tests for backend and frontend passes
- Deploys successfully on backend and frontend
- Hot reload works for the shared module

Need to resolve merge conflicts

.travis.yml
lamkeewei: Yep - this is to build the shared module. Previously there was not need to separate build the backend because `npm test` handles it. However, `ts-jest` [does not](https://github.com/kulshekhar/ts-jest/issues/1648) automatically compile project references.

PR #1147, opened 2021-04-22
fix: prevent campaign names from causing dashboard rows to overflow

zwliew:
## Problem

This is a minor design issue I noticed during testing:

A particularly long campaign name causes the dashboard table to overflow its container. This doesn't look very good.

## Solution

**Improvements**:

- Only prepend `Copy of` to campaign names which don't already have it (explanation in commit 6a3db9a)

**Bug Fixes**:

- Prevent the dashboard table from overflowing its container

## Before & After Screenshots

**BEFORE**:
<details>
<summary>Before screenshots</summary>

Screenshot 2021-04-15 at 2 19 05 PM

Screenshot 2021-04-15 at 2 19 22 PM
</details>

**AFTER**:
<details>
<summary>After screenshots</summary>

Screenshot 2021-04-15 at 2 20 04 PM

Screenshot 2021-04-15 at 2 19 45 PM
</details>

## Tests

**Duplicating a duplicated campaign should not prepend `Copy of` to the campaign's name**
1. Try duplicating a campaign which already has `Copy of` at the start of its name.
2. Assert that the campaign duplication modal does not prepend yet another `Copy of` to the new campaign's name.

**Table should not overflow the container**
1. Create a campaign with a very long name.
2. Assert that the dashboard table does not have a horizontal scrollbar.

## Deploy Notes

No deploy notes.

## Todo

- [ ] Maybe add a unit test for ensuring the `Copy of` prepending functionality works as intended?

### DIFF
--- a/frontend/src/components/common/title-bar/TitleBar.module.scss
+++ b/frontend/src/components/common/title-bar/TitleBar.module.scss
@@ -15,7 +15,13 @@
 }
 
 .titleText {
+  flex: 1;
   color: $secondary;
   margin-top: 0;
   margin-bottom: 0;
+  margin-right: $spacing-3;
+
+  text-overflow: ellipsis;
+  overflow: hidden;
+  white-space: nowrap;
 }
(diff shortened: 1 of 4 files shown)


lamkeewei (approved):
Thanks for fixing this. I've also added another fix to add ellipsis for long campaign names in `TitleBar`.

image

PR #1148, opened 2021-04-22
feat: refactor msg template components; add telegram character limit

zwliew:
## Problem

Currently, the `SMSTemplate` and `TelegramTemplate` components are nearly identical. The only difference is that `SMSTemplate` has an error message (which I'll also add to the Telegram template in this PR)

Since they're nearly identical, some unit tests written for both components will be duplicated for both of them as well (ref: https://github.com/opengovsg/postmangovsg/pull/1118#discussion_r616323252).

To reduce code and test duplication, refactor both components to a common component.

## Solution

**Features**:

- Replace `exceedsCharacterThreshold` with `warningCharacterCount` and `errorCharacterCount` thresholds
  - `warningCharacterCount` merely displays an error message
  - `errorCharacterCount` displays an error message __and__ disables the Next button
- Add a set of warning and error thresholds for the Telegram template as well
  - 4096 is the max based on https://github.com/yagop/node-telegram-bot-api/issues/165
  - Also, during manual tests, trying to send a message with >4096 chars results in an error msg from the API

**Improvements**:

- Refactor `SMSTemplate` and `TelegramTemplate` to a common component
- Refactor the tests accordingly

## Before & After Screenshots

To be added

The only visual difference is the Telegram template should look exactly the same as the SMS template now.

## Tests

**SMS and Telegram templates should function just as before**
1. Clicking next button goes next
2. Exceeding the warn threshold of 1000 chars displays an error message

**Exceeding error threshold should display error msg and disable the next button**
1. Type >1600 chars for SMS and >4096 chars for Telegram
2. The next button should be disabled and an error msg should be displayed 

## Deploy Notes

No deploy notes.

## Todo
- [x] Refactor the tests


frontend/src/components/dashboard/create/common/BodyTemplate.tsx
```diff
@@ -11,40 +11,49 @@ import {
...
-    if (exceedsCharacterThreshold(body)) {
-      setErrorMsg(
-        (
-          <span>
-            Your template has more than 1000 characters. Messages which are
-            longer than <b>1600</b> characters (including keywords) can&apos;t
-            be sent. Consider making your message short and sweet to make it
-            easier to read on a mobile device.
-          </span>
-        ) as any
```
zwliew, line 42: We don't need the `as any` anymore after adding a type definition for the `useState` call in line 32

frontend/src/components/dashboard/create/common/BodyTemplate/BodyTemplate.tsx
```diff
@@ -11,40 +12,51 @@ import {
   StepSection,
 } from 'components/common'
 import SaveDraftModal from 'components/dashboard/create/save-draft-modal'
-import { exceedsCharacterThreshold, saveTemplate } from 'services/sms.service'
+import { saveTemplate } from 'services/sms.service'
```
lamkeewei, line 15: This component does not work with Telegram.

lamkeewei (changes requested):
Need to fix to make it work for Telegram. Maybe pass the `saveTemplate` function as a prop?

frontend/src/components/dashboard/create/common/BodyTemplate/BodyTemplate.tsx
zwliew: Ah my bad. I guess this is a case where frontend tests would've caught the error haha

zwliew:
I think I'll move the unit tests for this component to this PR

lamkeewei (approved):
lgtm! Tested the following an works as expected:
1. Existing Telegram/SMS template saving should works as expected
2. Warning shown when character limits are reached for both SMS/Telegram

PR #1153, opened 2021-04-27
chore: upgrade dependencies

lamkeewei:
## Upgraded

| Dependency | From | To |
| :------------ | :-----| :-- |
| redis | 3.0.2 | 3.1.2 |
| react-draft-wysiwyg | 1.14.5 | 1.14.6 |
| validator | 13.0.0 | 13.6.0 |
| @types/cheerio | 0.22.24 | 0.22.28 |
| @types/node | 14.14.31 | 14.14.42 |
| twilio | 3.55.1 | 3.58.0 |
| moment | 2.24.0 | 2.29.1 |
| @types/papaparse | 5.0.3 | 5.2.5 |

### DIFF
--- a/backend/src/core/middlewares/campaign.middleware.ts
+++ b/backend/src/core/middlewares/campaign.middleware.ts
@@ -102,7 +102,10 @@ const listCampaigns = async (
   res: Response,
   next: NextFunction
 ): Promise<Response | void> => {
-  const { offset, limit } = req.query
+  const { offset, limit } = (req.query as unknown) as {
+    offset: number
+    limit: number
+  }
   const userId = req.session?.user?.id
   try {
     const { rows, count } = await CampaignService.listCampaigns({
--- a/backend/src/core/middlewares/campaign.middleware.ts
+++ b/backend/src/core/middlewares/campaign.middleware.ts
@@ -102,10 +102,7 @@ const listCampaigns = async (
   res: Response,
   next: NextFunction
 ): Promise<Response | void> => {
-  const { offset, limit } = (req.query as unknown) as {
-    offset: number
-    limit: number
-  }
+  const { offset, limit } = req.query
   const userId = req.session?.user?.id
   try {
     const { rows, count } = await CampaignService.listCampaigns({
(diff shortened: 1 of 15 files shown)


zwliew:
Can I include the @types/express update (#1102) in this as well? A quick build works fine

lamkeewei:
> Can I include the @types/express update (#1102) in this as well? A quick build works fine

Sure!

backend/src/core/middlewares/campaign.middleware.ts
```diff
@@ -102,7 +102,10 @@ const listCampaigns = async (
   res: Response,
   next: NextFunction
 ): Promise<Response | void> => {
-  const { offset, limit } = req.query
+  const { offset, limit } = (req.query as unknown) as {
```
zwliew, line 105: This is a workaround for the change in type definitions of `Request.query` (info: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43434#issuecomment-607181516).

I also tried defining the type of `Request.query` using generics like so: `req: Request<{}, any, any, { offset: number; limit: number }>`. But this causes the type error to propagate up to the routes.

After defining the type for `routes.get('/campaigns')`, the `celebrate` library errors since it doesn't have generics support. We could upgrade to v13.0.4 which adds generics (https://github.com/arb/celebrate/commit/69a9b973860edb0e243467a7970353d74fa2c5e8), but that would be a major version upgrade and might require more looking into.

lamkeewei: I'm more inclined to put off this upgrade first in today's release and fix it properly rather than casting `req.query` to `unknown`.

zwliew: Alright, sounds good!

zwliew (approved):
lgtm!

Tested the following on staging backend:
1. Sending a validation SMS works
2. Parsing CSV file works
3. Creating and sending a campaign works

Tested the following on frontend locally:
1. Exported at time is formatted correctly
2. Inputting text in the rich text editor works

PR #1158, opened 2021-04-28
feat(rich-text): support image as links

lamkeewei:
## Problem

Users have been requesting to use images as links in their emails.

## Solution

**Features**:
- Add optional field to provide link when inserting image
- Allow conversion to and from Draft.js state and HTML

**Bug Fixes**:
- Fix missing styles for popover when clicking on image

## Tests
**Insert image with link**
- Create email campaign and insert an image with link
- Popover when clicking on image should show link to URL
- Saving template and revisiting template again should show that link is preserved
- Image in message preview should also be linked

**Insert image without link**
- Create email campaign and insert an image without link
- Popover should not show any link

**Insert image with variable link**
- Create protected email campaign and insert an image with variable (e.g. `{{protectedlink}}` as link)
- In sent test message, the URL should be correctly substituted


miazima:
Tested on staging and the email sent out contains linked image and successfully navigates to correct link 
https://rich-text-image.postman.gov.sg

Had issues with
- image does not contain link in preview 
- when going back to create message step, the image popover does not contain the link and email sent out does not have link for image

lamkeewei:
> Tested on staging and the email sent out contains linked image and successfully navigates to correct link
> https://rich-text-image.postman.gov.sg
> 
> Had issues with
> 
> * image does not contain link in preview
> * when going back to create message step, the image popover does not contain the link and email sent out does not have link for image

@miazima could you help provide the steps to reproduce the errors you were facing? Wasn't able to reproduce this locally. Backend and workers also need to be deployed to this branch, otherwise the `data-link` attribute will be stripped.

miazima:
Hmm I tested on staging again, only issue I had is that the image was not linked in the preview block

lamkeewei:
> Hmm I tested on staging again, only issue I had is that the image was not linked in the preview block

@miazima this should be fixed on staging! Realised the problem is that the previous Amplify build failed because of the tests timeout. This was fixed in #1162.

image

miazima (approved):
Tested on staging, lgtm!

PR #1162, opened 2021-04-28
fix(frontend): fix test flakiness

### LINKED ISSUE #1161: Flaky frontend tests
**Describe the bug**
Some of the integration tests fail sporadically with the following error:
```
thrown: "Exceeded timeout of 5000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."
```

Specifically, this was experienced commonly with `src/components/dashboard/tests/integration/protected-email.test.tsx`. 

**To Reproduce**

1. Run `npm test`. It should fail every now and then.

**Expected behaviour**
All tests should pass 100% of the time.

**Possible solution**
1. Maybe we need to increase the timeout by adding `jest.setTimeout(SOME_LONG_TIMEOUT)` to `setupTests.ts`?
2. Maybe the test are not ending when they should be?

zwliew:
We were experiencing flaky tests as described in issue #1161.

## Problem

2 issues:
1. Some tests were failing with an error saying that the test exceeded the timeout of 5000ms.

2. The integration tests were written under the assumption that `REACT_APP_ANNOUNCEMENT_ACTIVE != "true"`. However, if it is set to `true`, the integration tests will break due to missing API endpoints and the tests not accounting for the need to dismiss the announcement modal.

Closes #1161

## Solution

- Increase the timeout of each test to 20s. 10s works fine locally on a MBP 16", but given that CI servers are likely slower, 2x the timeout to be safe.
- Add a dummy API mock for `/settings/announcement-version`
- Disable announcements for tests to enforce the assumption that `REACT_APP_ANNOUNCEMENT_ACTIVE=false`. This is fine because none of our tests rely on announcements.

Note: a longer term fix would be to add support for `announcement_version` in the API mocks, and configure the initial `announcement_version` appropriately during each test.

## Tests

All CI tests should pass reliably.

## Deploy Notes

 - Amplify no longer runs the tests. Any test failures should be caught on Travis during development. If a test failure does get to deployment, we can manually redeploy an older release.

### DIFF
--- a/frontend/src/config.ts
+++ b/frontend/src/config.ts
@@ -7,6 +7,7 @@ import type { InitializeOptions } from 'react-ga'
 // Re-export these later on as constants
 let gaInitializeOptions: InitializeOptions
 let gaTrackingId: string
+let announcementActive: string
 
 /**
  * Configs differentiated between environments
@@ -17,6 +18,9 @@ if (process.env.NODE_ENV === 'test') {
   gaInitializeOptions = {
     testMode: true,
   }
+
+  // Disable announcements in test environments
+  announcementActive = 'false'
 } else {
   /**
    * React env vars are used for injecting variables at build time
@@ -44,6 +48,9 @@ if (process.env.NODE_ENV === 'test') {
   gaInitializeOptions = {
     debug: false, // Set to true only on development
   }
+
+  // `REACT_APP_ANNOUNCEMENT_ACTIVE` must be set to `true` in Amplify if we want the modal to display
+  announcementActive = process.env.REACT_APP_ANNOUNCEMENT_ACTIVE as string
 }
 
 /**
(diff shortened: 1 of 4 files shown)


zwliew:
Could possibly explore using `--runInBand` or `maxWorkers=4` based on comments here: https://github.com/facebook/jest/issues/1524

frontend/src/setupTests.ts
```diff
@@ -5,6 +5,8 @@ import { Crypto } from '@peculiar/webcrypto'
 import 'locales' // Locales necessary for I18nProvider
 import { server } from './test-utils'
 
+jest.setTimeout(20000)
```
lamkeewei, line 8: Perhaps we can introduce an environment variable with a fallback to a default for this? This way if we want to tune this for different environments, we do not need to make changes in the code.

zwliew: Good idea. I'll do that with a default of 20s.

lamkeewei: Do test this out on Amplify also with repeated runs.

zwliew: Yup, I'm doing so

lamkeewei:
Also remember to remove running tests on Amplify.

zwliew:
I'll continue running tests until deployment tonight, but this should fix the issue that you had locally as well as the timeout issues.

frontend/src/config.ts
```diff
@@ -17,6 +18,9 @@ if (process.env.NODE_ENV === 'test') {
   gaInitializeOptions = {
     testMode: true,
   }
+
+  // Disable announcements in test environments
+  announcementActive = 'false'
```
zwliew, line 23: What do you think about disabling announcements for tests? I think it's fine since none of our tests rely on announcements currently.

lamkeewei: I think it's fine. Given that for most of the flows, that's not the main thing we're testing for.

lamkeewei (approved):
lgtm! This fixed the issues I was facing previously when running tests. Merge whenever you're ready.

PR #1168, opened 2021-04-29
feat: set up sequelize migrations and seeders

### LINKED ISSUE #1111: Set up sequelize migrations instead of using `sync`

We currently use `sequelize.sync()` to modify our DB tables. This is [not recommended](https://sequelize.org/master/manual/model-basics.html#synchronization-in-production) because it is potentially destructive, assuming `alter` or `force` is enabled.

Consider using [sequelize migrations](https://sequelize.org/master/manual/migrations.html) instead.

**Additional context**

Check out this [great reference issue from CheckFirst](https://github.com/opengovsg/checkfirst/issues/282) for more context.

zwliew:
## Problem

We currently use `sequelize.sync()` to update our DB tables. However, this is not recommended for production use as it is potentially destructive if the `alter` or `force` options are enabled.

Furthermore, there is no way to do more complex migrations with `sequelize.sync()` without using `alter` or `force`. 

Closes #1111

## Solution

Manage database modifications using sequelize migrations instead. Currently, the goal is to manually use `sequelize-cli` to run migrations. (We may want to consider doing this programmatically using `umzug`)

Benefits of using sequelize migrations:
1. We can define a safe way to migrate from one DB schema to the next
2. We can define a safe way to undo migrations

**Improvements**:

- Removed `sequelize.sync()` from the backend and worker initialization code
- Removed function creation and credential seeding from the backend and worker initialization code
- Set up sequelize-cli migrations and seeders
  - Created migrations to create tables
  - Created migrations to create functions
  - Created seeder to seed demo and default credentials

## Tests

**The migrations and seeders should create a copy of the current database schema, functions, and default creds**
1. Create a new database
2. Modify DB_URI to point to that new database
3. Run `npx sequelize db:migrate` to run all migrations
4. Run `npx sequelize db:seed:all` to run all seeders
5. The database should have the same schema, functions, and default creds as the existing database
6. The SequelizeMeta and SequelizeData tables should be populated with the names of all the migration and seeder files

## Deploy Notes

**Some thoughts about doing common migrations**
1. Creating or replacing functions: manually run the migration before deploying new worker code
2. Deleting functions: deploy new worker code before manually running the migration
3. Creating and editing (add-only) tables: manually run the migration before deploying new backend code
4. Deleting tables: deploy new backend code before manually running the migration
5. Editing tables (destructive): 3 step process. 1, make additive changes. 2, migrate application code to use the new changes. 3, remove the old changes after all application code is migrated.

**New dependencies**:
- `umzug`: for running seeders on test environments

**New dev dependencies**:
- `@types/umzug`

## Todo
- [x] Set up sequelize-cli support
- [x] Create table migrations
- [x] Create worker function migrations
- [x] Seed default and demo credentials
- [x] Ensure that the migrated local database is exactly the same as the staging one (functions, tables, seeds)
- [x] Consider: should we move these to a separate top-level `database` folder (alongside `backend` and `worker`)?
  - Result: might be a good idea to enforce some sort of file-level separation of concerns, but I'll put it in `backend` to leverage the existing config definitions, at least for now.
- [x] Consider: should we have programmatic migrations using `umzug`? Do we want it to be automatic on worker/backend startup?
  - Result: `umzug` [doesn't support](https://github.com/sequelize/umzug/discussions/144) concurrent migrations via locks. It is safer to migrate manually. An alternative is to do something similar to [umzug-sync](https://github.com/SGrondin/umzug-sync) or [sequelize-migration-lock](https://github.com/DanielHreben/sequelize-migration-lock), which implements locks over `sequelize`
- [ ] Consider: how will NODE_ENV be defined? This is impt because e.g. the names of the demo credentials uses NODE_ENV. However, if we're using our local machines to seed the prod DB via jumphost, won't NODE_ENV be defined based on the env of our local machine? In that case, we would need to explicitly define NODE_ENV=production when migrating/seeding the prod DB.
- [x] Maybe use `umzug` to migrate the database for test environments?
  - Result: done


zwliew:
I'm using umzug v2.x here, which is supposed to be stable. On the other hand, umzug v3.x is in beta.

zwliew:
todo after discussion with Kee Wei: remove umzug dependency and run seeders for tests using built-in Node modules.

zwliew:
Rebased the branch.

jeantanzj:
I've diffed the sql functions, and checked foreign keys, primary keys, and defaults on a fresh local db, and they look correct.

I don't think it's worth the risk to run a fresh prod instance and migrate data. 

I propose to deploy this by adding two tables SequelizeMeta and SequelizeData manually, and inserting the migrations, then using the migrations only in the future.

backend/src/database/config.ts
```diff
@@ -0,0 +1,15 @@
+import '../setup'
+
+import { dbConfig } from './util'
+
+const commonConfig = {
+  dialect: 'postgres',
+  seederStorage: 'sequelize',
+  ...dbConfig,
```
lamkeewei, line 8:
```suggestion
migrationStorageTableName: 'sequelize_meta',
seederStorageTableName: 'sequelize_data',
  ...dbConfig,
```

Might want to rename to underscore case to match our other table names.

PR #1172, opened 2021-05-04
refactor: use shared function to initialize models

lamkeewei:
## Problem

When adding/removing a model, developer need to remember to update both test and runtime loaders. 

## Solution

**Improvements**:

- Introduce shared `initializeModels` function to be used in both test and runtime `sequelize` loaders.

**Bug fixes**
- Load `Agency` model for tests


miazima (approved):
Is this something we could extract out too to prevent inconsistencies?

https://github.com/opengovsg/postmangovsg/blob/f4c90e8259bd50f2bbebe1de4bfc907aa55f2893/backend/src/core/loaders/sequelize.loader.ts#L60-L68

lamkeewei:
> Is this something we could extract out too to prevent inconsistencies?
> 
> https://github.com/opengovsg/postmangovsg/blob/f4c90e8259bd50f2bbebe1de4bfc907aa55f2893/backend/src/core/loaders/sequelize.loader.ts#L60-L68

Yep definitely. But I'm keeping this here first because I think using the seeders from #1168 will be more appropriate for this purpose.

PR #1177, opened 2021-05-05
fix(backend): docker build and tsc output directory structure

lamkeewei:
## Problem
Deployments for backends were failing after merging #940 because:
1. `tsconfig.build.json` was not copied into the image during Docker build
2. `rootDir` was changed in `backend/tsconfig.json`, this results in a different build directory structure being created. `npm start` fails because `server.js` now resides in `build/src/server.js`.

## Solution

**Bug Fixes**:

- Copied `tsconfig.build.json` into the Docker image
- Reverted `rootDir` to `./src`. Also reverted `baseUrl` for consistency.

## Tests
- Deploy this branch and observe that the Elastic Beanstalk deploy succeeds on AWS console

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -19,6 +19,7 @@ RUN npm ci
 
 COPY src ./src
 COPY tsconfig.json ./
+COPY tsconfig.build.json ./
 
 RUN npm run build
 RUN npm prune --production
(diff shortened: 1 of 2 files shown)


miazima (approved):
Tested on staging and elastic beanstalk successfully deployed

Issue #1143, opened 2021-04-19
Unable to update Telegram ID for an existing phone number


lamkeewei:
**Describe the bug**

Currently, we only account for the case where a user with the same Telegram ID wants to update his/her number. This is done through the [query](https://github.com/opengovsg/postmangovsg/blob/9a1e4f7c18f5ba07966f0959df7f0300acfb7ef3/backend/src/telegram/utils/callback/handlers/contact.ts#L41) that updates the phone number if the Telegram ID already exists. However, the other scenario where the user has the same phone number but wants to change his Telegram ID is not handled. 

Insertion will result in a unique constraint conflict as the phone number is part of the primary key and has to be unique.

**To Reproduce**

1. Ensure that subscription is already captured in both `telegram_subscribers` and `bot_subscribers` table
2. Attempt to update the Telegram ID for the same phone number by making a POST to the `/callback/telegram/:botId` with the following payload:
```
{
   "message":{
      "from":{
         "id":[NEW TELEGRAM ID]
      },
      "contact":{
         "phone_number": "[OLD PHONE NUMBER]",
         "first_name": "John",
         "last_name": "Doe",
         "user_id":[NEW TELEGRAM ID]
      }
   }
}
3. An error should be thrown
```

**Expected behavior**

zwliew:
Would it be reasonable to just delete the existing row containing the phone number before inserting the new row?

zwliew:
Also a related question I have is: for telegram_subscribers, why was the phone number chosen to be the primary key instead of the telegram id?

PR #1178, opened 2021-05-06
fix: fix error when updating Telegram ID for an existing phone number

zwliew:
## Problem

Closes issue #1143

## Solution

**Bug Fixes**:

- Try updating either phone number or telegram ID. If no update occurs, insert a new row.

## Tests

These tests have been verified on staging.

**Users should be able to update telegram ID for an existing phone number**
1. Change your `telegram_id` to a different number in the `telegram_subscribers` table.
2. Run the `/updatenumber` command on the Telegram bot
3. The bot should say `Your phone number and Telegram ID have been updated.` and `telegram_id` should be updated in the `telegram_subscribers` table.

**Users should be able to update phone number for an existing telegram ID**
1. Change your `phone_number` to a different number in the `telegram_subscribers` table.
2. Run the `/updatenumber` command on the Telegram bot
3.  The bot should say `Your phone number and Telegram ID have been updated.` and `phone_number` should be updated in the `telegram_subscribers` table.

**Users should be able to create a new `telegram_subscribers` row**
1. Delete the row corresponding to your account in the `telegram_subscribers` table.
2. Run the `/updatenumber` command on the Telegram bot
3.  The bot should say `Your phone number and Telegram ID have been updated.` and there should be a new row reflecting your `phone_number` and `telegram_id` in the table.

### DIFF
--- a/backend/src/telegram/utils/callback/handlers/contact.ts
+++ b/backend/src/telegram/utils/callback/handlers/contact.ts
@@ -7,6 +7,14 @@ import { TelegramSubscriber, BotSubscriber } from '@telegram/models'
 
 const logger = loggerWithLabel(module)
 
+enum UpsertTelegramSubscriberResult {
+  Nothing,
+  Failed,
+  Created,
+  UpdatedPhoneNumber,
+  UpdatedTelegramId,
+}
+
 /**
  * Upserts a Telegram subscriber.
  *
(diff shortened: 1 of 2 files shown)


zwliew:
I thought of one additional edge case when thinking about the fix.

Imagine this scenario:
There are 2 rows in the `telegram_subscribers` table.

|`phone_number`|`telegram_id`|
----------
| A | 000 |
| B | 111 |

What if a user tries to update with the following parameters: `phone_number=B`, `telegram_id=000`?

backend/src/telegram/utils/callback/handlers/contact.ts
```diff
@@ -16,46 +22,56 @@ const logger = loggerWithLabel(module)
...
+    async (transaction) => {
+      /**
+       * Split into 2 scenarios:
+       * 1. a row with matching phone_number or telegram_id: update the row
+       * 2. no row matches: insert a new row
+       *
+       * Note: we have to use a raw query to update the phone_number.
+       * It is a primary key, and sequelize doesn't allow us to update
+       * primary keys in sequelize models.
+       */
```
zwliew, line 47: I tried a few fixes which used sequelize model methods (`findOrCreate`, `findOne` then `save`). However, they didn't work for updating `phone_number` due to this issue. I kept them as separate commits in this branch just for reference.

In the end, the simplest and probably the cleanest way to fix this seems to be the current iteration: trying to update, else insert.

lamkeewei:
> I thought of one additional edge case when thinking about the fix. However, given that it sounds like a really rare scenario, it might not be worth handling it.
> 
> Imagine this scenario:
> 
> There are 2 rows in the `telegram_subscribers` table.
> 
> `phone_number`	`telegram_id`
> A	000
> B	111
> What if a user tries to update with the following parameters: `phone_number=B`, `telegram_id=000`?
> 
> I couldn't think of a clear cut answer. One solution would be to delete one of the rows and update the other. Another would be to treat both rows as the same and merge them.
> 
> Steps to achieve this scenario:
> 
> 1. A user creates 2 Telegram accounts and subscribes them to the bot as shown above.
> 2. Then, the user with `telegram_id=111` changes their `phone_number` to `C` **on Telegram only**. But, the user doesn't update this in Postman. Now, there is a desync between Postman's `telegram_subscribers` table and the canonical Telegram mapping.
> 3. Lastly, the user with `telegram_id=000` changes their `phone_number` to `B`. This time, the user runs `/updatenumber` to update Postman. However, since `phone_number=B` and `telegram_id=000` both already exist in the table as 2 separate rows, this is an issue. It is unclear which row to update, and which row to delete/merge, in order to maintain uniqueness of the keys.

I think this seem reasonably rare that we don't have to handle it.

lamkeewei (approved):

PR #1193, opened 2021-05-12
feat: allow users to customise sender name

### LINKED ISSUE #1174: Allow users to customize sender name
**Aim**
to improve email legitimacy. A key reason why recipients perceive Postman's emails as suspicious is because the sender name is Postman whereas the email supposedly was sent by an agency.

**Todo**
- users should be able to customise the sender name
- "via Government Mail" should be appended to the end of the sender name regardless, and this should be communicated to the user. this is because savvy users will also realise and are trained to ensure that the sender name matches the domain.

lamkeewei:
## Problem

Closes #1174 

Hosted at https://sender-name.postman.gov.sg for testing.

## Solution
- Allow use of a custom sender name with both default and custom from addresses
- Automatically appending "via Government Mail" for custom from names

## Before & After Screenshots
image

## Tests
**Automated tests**
- Run test for `frontend` and `backend`

**Use default sender name with default address**
- Create email campaign and leave sender name as default
- Default sender name should be saved in template. Sender name should not be appended with "via Postman"
- Email messages sent out should be sent with default sender name

**Use custom sender name with default address**
- Create email campaign and use custom sender name (e.g. "Hello")
- Custom sender name should be saved in template. Sender name should be appended with "via Postman"
- Email messages sent out should be sent with custom sender name and default from address

**Use default sender name with custom from address** 
- Create email campaign and select custom from address (this should be selected by default). Leave sender name as default
- Default sender name should be saved in template. Sender name should not be appended with "via Postman"
- Email messages sent out should be sent with default sender name and custom from address

**Use custom sender name with custom from address** 
- Create email campaign and select custom from address (this should be selected by default). Replace sender name with a custom sender name
- Custom sender name should be saved in template. Sender name should be appended with "via Postman"
- Email messages sent out should be sent with custom sender name and custom from address

## Deploy Notes
**New environment variables**
- `BACKEND_MAIL_VIA` - Use to configure the via line for custom sender name. Defaults to `via Postman`


frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -124,25 +168,63 @@ const EmailTemplate = ({
...
     <>
       <StepSection>
         <StepHeader title="Create email message" subtitle="Step 1" />
 
         <div>
           <h4>From</h4>
-          <p>Emails will be sent from this address</p>
+          <p>
+            Sender details. Custom sender name will be automatically appended
+            with &quot;via Government Mail&quot;.
```
mjgui, line 214: This should be 'via Postman' I think

mjgui: Maybe we can consider replacing this with a string interpolation using the constant `mailVia`

lamkeewei: Good catch!

frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -31,6 +32,19 @@ import { FinishLaterModalContext } from 'contexts/finish-later.modal.context'
 import { saveTemplate } from 'services/email.service'
 import { getCustomFromAddresses } from 'services/settings.service'
 
+const parseFromAddress = (
```
jeantanzj, line 35: should probably go into shared folder now

frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -52,7 +66,15 @@ const EmailTemplate = ({
   const [replyTo, setReplyTo] = useState(
     initialReplyTo === userEmail ? null : initialReplyTo
   )
-  const [from, setFrom] = useState(initialFrom)
+  const {
+    fromName: initialFromName,
+    fromAddress: initialFromAddress,
+  } = initialFrom
+    ? parseFromAddress(initialFrom)
+    : { fromName: '', fromAddress: '' }
```
jeantanzj, line 74: There should never be an empty initialFrom, correct? If that's correct then this should throw an error

jeantanzj:
What were the reasons for setting the name on a campaign level instead of at the settings page? 
I think the code would have been cleaner if users could set sender name in settings, and  would also allow users to use a consistent name and email.  Worried about typos.

I'm not implying that this is incorrect, but trying to see if there's something  I'd missed. 

Also as a side note
This might be easier to understand/visual than the instruction "appended with"
Screenshot 2021-05-31 at 11 28 56 AM

lamkeewei:
> What were the reasons for setting the name on a campaign level instead of at the settings page?
> I think the code would have been cleaner and already supported by existing code if users could set sender name in settings, and would also allow users to use a consistent name and email. Worried about typos.
> 
> I'm not implying that this is incorrect, but trying to see if there's something I'd missed.
> 
> Also as a side note
> This might be easier to understand/visual than the instruction "appended with"
> Screenshot 2021-05-31 at 11 27 15 AM

Yep definitely doable. 

@viviennechong and @sarahsalim, any thoughts on this alternative user flow:
1. User goes to "Settings" and add a new from using a custom sender name
2. Existing email template page remains the same, user will select the added custom from in the dropdown

I think the added benefit is that users can save the list of froms that they've previously used. They don't have to re-type the sender name each time they want to use it. Could help avoid typos/inconsistency. Though, it does add slight bit of friction to using custom sender name.

Let me know what are your thoughts.

viviennechong:
I adopted the current tags for this
### "via Postman"
image

### "via Government Mail"
image


### Using "via Postman" as an example:

image

image
(Fixed height of 24 px)

### Text properties:
```
//styleName: overline;
font-family: Noto Sans;
font-size: 12px;
font-style: normal;
font-weight: 600;
line-height: 16px;
letter-spacing: 1.5px;
text-align: center;
```

### Grey container properties
```
height: 24px; //fixed
width: 122px;
left: 1046px;
top: 424px;
border-radius: 5px;
padding: 4px, 12px, 4px, 12px;
```

frontend/src/components/dashboard/create/email/EmailTemplate.tsx
lamkeewei: Turns out that `initialFrom` is undefined when it is a new campaign without a saved template. I've added a comment for this.

frontend/src/components/dashboard/create/email/EmailTemplate.tsx
lamkeewei: Done!

lamkeewei:
@jeantanzj updated the following:
- Moved common functions to `shared` module
- Added input tag to show when mail via is active

jeantanzj (approved):

PR #1196, opened 2021-05-18
feat(worker): remove threads and shutdown gracefully

### LINKED ISSUE #1137: [Deployment] Graceful termination of workers

- Restarts of the workers cause the API response for in-flight sending requests to be lost as we do not have graceful termination.
- The status for these messages will not be correctly reflected, resulting in erroneous or incorrect campaign statistics.

The requirement of the solution is as follows:
- Prevent messages from being stuck in the SENDING state if worker is terminated midway before receiving API responses from SES/Twilio. Not doing this will affect the accuracy of our delivery reports.
- Preserve the sending guarantees provided by the current sending logic

This can be achieved by:
1. Removing the threads library to make it possible for the worker process to directly handle `SIGTERM` or `SIGINT` signals
2. Ensure responses for the latest batch of in-flight messages is received before terminating worker

lamkeewei:
## Problem

Closes #1137 

## Solution

**Improvements**:
- Remove `threads` library
- Handle `SIGINT` or `SIGTERM` and flip `shouldRun` flag. 
- Use `exec` to replace shell process in `docker-entrypoint.sh` to allow proper signal handling. **tldr;** without `exec`, bash is the child process of `tini` not `npm`/`node`. See this [issue on tini's repo](https://github.com/krallin/tini/issues/72#issuecomment-274316094) for explanation. 
- In `development` mode, only load scripts using from sender. This is to prevent an error when both processes attempts to load the SQL scripts at the same time. The loading of the scripts will be eventually replaced with migrations. 

### Interrupt local workers while sending campaign
- Start sender and logger worker processes using `MESSAGE_WORKER_SENDER=1 npm start` and `MESSAGE_WORKER_LOGGER=1 npm start` respectively. `tsc-watch` interferes with signal handling
- Create and send a campaign with a number of recipients (they can all be pointing to the same address) and set send rate to `1`. This will allow you to have time to stop the worker mid-send.
- Attempt to terminate the sending worker with `Ctrl-C`
- Observe through the logs that worker does not terminate immediately and wait for the promises from `sendMessage` to resolve first.

### Testing on ECS
- Repeat the same test above on staging
- Observe that while re-deploying, it should log out line to indicate that shutdown signal is received and the worker is attempting to gracefully shutdown.

### Simulate ECS signal handling with Docker locally
ECS makes use of Docker engine to run containers. We can simulate what happens when a container is stopped on ECS by using `docker stop`. 

In order to test this out, follow these steps:

**Step 1:** Patch the worker code to listen for `SIGTERM` even in `development` mode [here](https://github.com/opengovsg/postmangovsg/blob/b8a4df2be1a7d7be834672ed86242dbb096f8e28/worker/src/core/loaders/message-worker.loader.ts#L13-L16).
```diff
process.on(
-    config.get('env') === 'development' ? 'SIGINT' : 'SIGTERM',
+   'SIGTERM',
    worker.shutdown
)
````

**Step 2:** Build the container image locally:
```bash
$ cd workers

# Docker build expects modules folder to be in the worker folder
$ cp -r ../shared .

# Build the container worker image locally
$ docker build -t postmangovsg-workers .
```

**Step 3:** Set `DB_URI` in your `.env` file to `postgres://[USERNAME]@host.docker.internal:5432/postmangovsg_dev`. The hostname `host.docker.internal` [resolves to the host machine IP address](https://docs.docker.com/docker-for-mac/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-). 

**Step 4:** Start the Docker container with your local `.env` file mounted and simulate stopping the container:
```bash
# We mount .env file in read-only mode to prevent it from being overwritten in docker-entrypoint.sh
$ docker run \ 
    -v $(pwd)/.env:/usr/home/postmangovsg/.env:ro \
    --network=host \
    --name sender \
    -e MESSAGE_WORKER_SENDER=1 \
    -d postmangovsg-workers

# Tail logs for your sender container
$ docker logs -f sender
```

On a separate shell, run `docker stop sender`. Observe in the logs that the shutdown signal is received and graceful shutdown is attempted.

Finally, to clean up the exited container so that you can test again, run `docker rm sender` to remove the stopped container.

## Deploy Notes

**After deploy**
- Update container stop timeout to `120` for sender and logger task definitions. This is the maximum amount of time permissible for instances running on Fargate before `SIGKILL` is sent.


worker/src/core/loaders/message-worker.loader.ts
```diff
@@ -1,51 +1,30 @@
...
+      config.get('env') === 'development' ? 'SIGINT' : 'SIGTERM',
+      worker.shutdown
+    )
+    await worker.start(workerId, isLogger)
   } catch (err) {
     logger.error({
       message: 'Worker died',
       workerId,
-      isLogger,
+      isLogger: false,
```
jeantanzj, line 22: isLogger should be `config.get('messageWorker.numLogger') > 0`

lamkeewei: Fixed!

lamkeewei:
TODO:
- [ ] Investigate if graceful termination works if there is an uncaught error
- [ ] Call `sequelize.close()` on termination

lamkeewei:
@jeantanzj took a quick look and it seems like the `SIGINT` being sent twice [only occurs when running starting the worker using `npm start`](https://stackoverflow.com/questions/54722158/node-and-npm-running-script-and-ctrl-c-triggers-sigint-twice). Starting the worker directly with `node build/server.js` doesn't cause `SIGINT` being sent twice. 

Running using Docker and `docker stop` to simulate the ECS environment also shows `SIGTERM` being handled only once.

jeantanzj:
Yep I realized that. Thanks!

jeantanzj (approved):

PR #1197, opened 2021-05-18
Backend tests 

mjgui:
## Tests
- Add tests to ensure adding a credential only succeds with valid API keys


mjgui:
Closing this. Will be done with PR #1204

PR #1199, opened 2021-05-19
chore(worker): parallel ecs deploys and exit deploy script on error

### LINKED ISSUE #1049: [Deployment] Improve the Travis deployment script for workers

There are 2 issues with the Travis deploy script for workers:
1. The exit code for the script depends solely on the exit code for the logger deployment. This masks potential errors in intermediate steps of the scripts.
2. The logger and sender deployments are run sequentially. This is potentially slower.

1. Return a non-zero exit code if any of the intermediate steps fail.
2. Run the deployments concurrently. (could test to see whether this actually reduces deployment time)

lamkeewei:
## Problem

Closes #1049 

## Solution

**Improvements**:
- Set shell script to exit on error and print out command that failed
- Run `ecs-deploy` in parallel to speed up deploy slightly

## Tests
- Ensure that build and deploy succeed for workers
- Observe that deploys for sender and logger runs in parallel

### DIFF
--- a/worker/deploy.sh
+++ b/worker/deploy.sh
@@ -24,5 +24,6 @@ docker build -t $TAG .
 docker push $TAG
 
 # Update ECS
-./ecs-deploy -c $CLUSTER -n $SENDING_SERVICE -i $TAG -r $AWS_DEFAULT_REGION --use-latest-task-def -t 300
-./ecs-deploy -c $CLUSTER -n $LOGGING_SERVICE -i $TAG -r $AWS_DEFAULT_REGION --use-latest-task-def -t 300
+./ecs-deploy -c $CLUSTER -n $SENDING_SERVICE -i $TAG -r $AWS_DEFAULT_REGION --use-latest-task-def -t 300 &
+./ecs-deploy -c $CLUSTER -n $LOGGING_SERVICE -i $TAG -r $AWS_DEFAULT_REGION --use-latest-task-def -t 300 &
+wait
--- a/worker/deploy.sh
+++ b/worker/deploy.sh
@@ -1,4 +1,11 @@
 #!/bin/bash
+# Exit on failure of any command
+set -e
+# Keep track of the last executed command
+trap 'LAST_COMMAND=$CURRENT_COMMAND; CURRENT_COMMAND=$BASH_COMMAND' DEBUG
+# Echo an error message before exiting
+trap 'echo "\"${LAST_COMMAND}\" command failed with exit code $?."' EXIT
+
 CLUSTER=$1
 SENDING_SERVICE=$2
 LOGGING_SERVICE=$3
--- a/worker/deploy.sh
+++ b/worker/deploy.sh
@@ -24,6 +24,6 @@ docker build -t $TAG .
 docker push $TAG
 
 # Update ECS
-./ecs-deploy -c $CLUSTER -n $SENDING_SERVICE -i $TAG -r $AWS_DEFAULT_REGION --use-latest-task-def -t 300 &
-./ecs-deploy -c $CLUSTER -n $LOGGING_SERVICE -i $TAG -r $AWS_DEFAULT_REGION --use-latest-task-def -t 300 &
+./ecs-deploy -c $CLUSTER -n $SENDING_SERVICE -i $TAG -r $AWS_DEFAULT_REGION --use-latest-task-def -t 600 &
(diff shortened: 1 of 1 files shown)


mjgui (dismissed):

jeantanzj:
```
# Keep track of the last executed command
trap 'LAST_COMMAND=$CURRENT_COMMAND; CURRENT_COMMAND=$BASH_COMMAND' DEBUG
# Echo an error message before exiting
trap 'echo "\"${LAST_COMMAND}\" command failed with exit code $?."' ERR
```
This doesn't seem to do anything? 

On branch fix-worker-deploy-err
Screenshot 2021-07-26 at 12 46 00 PM

mjgui:
Oh my bad - I tested it using a different command (not `ecs-deploy`) and it throws an error because I didn't include `& wait` at the end

mjgui:
Oh hm looked into this - `wait` will always return a value of 0 when no pid is specified
For it to properly throw an error, we need to either pass in the exact pids into `wait` or just run `wait -n` twice

That being said, in which case the logging actually won't work because it's the `wait` command that will throw the error - we probably need to loop through the `pid`s, run `wait` on them (which is blocking, but it's ok since both deployments are backgrounded), then log the command based on `pid` if exit code is nonzero

```bash
for pid in ${PIDS[@]}; do
  wait ${pid}
  if [[ $? -ne 0 ]]; then
    # Do something here
  fi
done
```

reference: https://stackoverflow.com/a/46212640

lamkeewei:
> ```
> # Keep track of the last executed command
> trap 'LAST_COMMAND=$CURRENT_COMMAND; CURRENT_COMMAND=$BASH_COMMAND' DEBUG
> # Echo an error message before exiting
> trap 'echo "\"${LAST_COMMAND}\" command failed with exit code $?."' ERR
> ```
> 
> This doesn't seem to do anything?
> 
> https://travis-ci.com/github/opengovsg/postmangovsg/jobs/526784526
> Screenshot 2021-07-26 at 12 46 00 PM

@jeantanzj The trap for `ERR` is only triggered when a bad command is called. If I'm not wrong, an `exit` called within a function in the same script will exit the current script instead of triggering an error. See [build](https://travis-ci.com/github/opengovsg/postmangovsg/jobs/526788437).

lamkeewei:
> Oh hm looked into this - `wait` will always return a value of 0 when no pid is specified
> For it to properly throw an error, we need to either pass in the exact pids into `wait` or just run `wait -n` twice
> 
> That being said, in which case the logging actually won't work because it's the `wait` command that will throw the error - we probably need to loop through the `pid`s, run `wait` on them (which is blocking, but it's ok since both deployments are backgrounded), then log the command based on `pid` if exit code is nonzero
> 
> ```shell
> for pid in ${PIDS[@]}; do
>   wait ${pid}
>   if [[ $? -ne 0 ]]; then
>     # Do something here
>   fi
> done
> ```
> 
> reference: https://stackoverflow.com/a/46212640

Good catch! Let me fix this.

worker/deploy.sh
```diff
@@ -25,5 +25,9 @@ docker push $TAG
 
 # Update ECS
 ./ecs-deploy -c $CLUSTER -n $SENDING_SERVICE -i $TAG -r $AWS_DEFAULT_REGION --use-latest-task-def -t 600 &
+SENDER_DEPLOY_PID=$!
 ./ecs-deploy -c $CLUSTER -n $LOGGING_SERVICE -i $TAG -r $AWS_DEFAULT_REGION --use-latest-task-def -t 600 &
-wait
+LOGGER_DEPLOY_PID=$!
+
+wait $SENDER_DEPLOY_PID
```
lamkeewei, line 32: @mjgui I'm manually storing the PID for each ECS deploy and waiting on them. Upon failure, the logs print either `"wait $SENDER_DEPLOY_PID" command failed with exit code` or `"wait $LOGGER_DEPLOY_PID" command failed with exit code`.

mjgui: Do we need a more informative error code? If not that should work

lamkeewei: I think given that the logs from `ecs-deploy` still prints, there should have sufficient information to determine any faults. The error printout at the end will point us to which deploy failed (also we can always checker ECS on AWS console).

jeantanzj (approved):

PR #1202, opened 2021-05-21
feat(backend): interruptible recipients list uploads

### LINKED ISSUE #1138: [Deployment] Offload handling of recipient list processing to separate process

- When the backend is restarted during a deployment, processing of an uploaded recipient list might be interrupted midway. 
- This would result in the upload to be stuck in a “processing” state until the user revisits the upload recipient page for the campaign after 10 minutes.
- Once the timeout is reached, the frontend will be “forced reset” to re-enable the upload button for the user to retry uploading.

The requirement of the solution is as follows:
- Processing of the recipient list should not be affected by a deployment or restart of the backend (from updating environment variables)

This can be achieved by:
- Processing the upload by invoking a lambda instead of processing the CSV file in the backend

lamkeewei:
## Problem

Closes #1138 

## Solution

**Features**:
- Use [bee-queue](https://github.com/bee-queue/bee-queue) to handle processing/resuming recipients list uploads. This allows for interruption to recipient list processing (e.g. from server restart) to be detected and resumed on server restart.

**Improvements**:
- Refactored out recipient list processing logic to a common generic function that's reusable across all channels

## Tests
Run unit tests to ensure that `processUpload` works as expected

**Test resumption of upload**
- Start up backend, frontend and workers separately
- Manually introduce delay in the upload processing by patching the code [here](https://github.com/opengovsg/postmangovsg/blob/441874ef62bad293abcb2d93c31093031df33e73/backend/src/core/services/upload.service.ts#L319-L329)
```diff
try {
+ await new Promise((resolve) => setTimeout(resolve, 30000))
  await ParseCsvService.parseAndProcessCsv(
    downloadStream,
    onPreview(params),
    onChunk(params),
    onComplete({
      ...params,
      key: s3Key,
      filename,
    })
  )
```
- Through the frontend, create a campaign and upload a recipient list
- While upload is being processed, terminate the backend process
- Restart the backend process and observe the logs that the upload restarts as expected

## Deploy Notes

- `UPLOAD_REDIS_URI` : URI to the redis database for recipient list upload job queue
- `UPLOAD_QUEUE_NAME` : Name of queue used to store upload jobs
- `UPLOAD_CONCURRENCY` : Maximum number of simultaneous active jobs
- `UPLOAD_CHECK_STALLED_INTERVAL` : How often to check for stalled jobs in milliseconds

**New dependencies**:
- `bee-queue`: Redis backed job queue

## TODOs
- [x] Backend tests for `processUpload`
- [x] ~Shift timeout mechanism to backend to ensure that overly long jobs are terminated and not left hogging resources.~ _Will tackle in a separate PR_


backend/src/core/config.ts
```diff
@@ -456,6 +456,32 @@ const config = convict({
...
     },
   },
+  upload: {
+    redisUri: {
+      doc: 'URI to the redis database for recipient list upload job queue',
+      default: '',
+      env: 'UPLOAD_REDIS_URI',
+      format: 'required-string',
+      sensitive: true,
+    },
```
mjgui, line 466: We should probably add this in `.env-example` since it's a required env var

mjgui:
@jeantanzj Tested resumption of uploads - works fine for me (once `UPLOAD_REDIS_URI` is set)

backend/src/core/services/upload.service.ts
```diff
@@ -11,19 +13,33 @@ import { MissingTemplateKeysError } from '@core/errors/template.errors'
...
+  Upload,
+  AllowedTemplateTypes,
+} from '@core/interfaces'
 import { CSVParams } from '@core/types'
-import { StatsService, CampaignService } from '.'
+import S3Client from '@core/services/s3-client.class'
+import { StatsService, CampaignService, ParseCsvService } from '.'
 
 const logger = loggerWithLabel(module)
 const MAX_PROCESSING_TIME = config.get('csvProcessingTimeout')
```
jeantanzj, line 27: do we still need this? is the job guaranteed to fail if the process promise is rejected?

jeantanzj:
One of my main concerns with bee-queue is that it provides at least once delivery. 
In our transaction (onPreview, onChunk and onComplete), onPreview will delete all existing messages for that campaign. If multiple transactions occurred and tried to delete the messages (DELETE * FROM email_messages WHERE campaign_id = :id), it could cause a deadlock. https://stackoverflow.com/questions/51969812/can-two-concurrent-but-identical-delete-statements-cause-a-deadlock

That being said, I tried to cause concurrency issues (by setting stalled interval to a very low value) and ran 5 backends simultaneously, and didn't manage to get multiple instances to pick up the same job. Is there anywhere in the documentation that you've read that guarantees/explains how only one worker would pick up the job from the queue at a time?

jeantanzj:
> UPLOAD_REDIS_URI 

Btw could we perhaps name this REDIS_UPLOAD_URI?

lamkeewei:
>  Is there anywhere in the documentation that you've read that guarantees/explains how only one worker would pick up the job from the queue at a time?

Based on the [docs](https://github.com/bee-queue/bee-queue#under-the-hood), the main mechanism that `bee-queue` uses to assign one worker to one job at a time is through the use of [`BRPOPLPUSH`](https://redis.io/commands/BRPOPLPUSH) on Redis. Given that Redis is single-threaded, only one command is executed at a given time. This ensures that only one worker will succeed at popping from the list. `BRPOPLPUSH` is used to ensure that popping from the `waiting` list and then pushing to the `active` list is an atomic operation. 

Once the job is assigned to a worker, it will start processing the job. While processing the job, the worker will regularly send heartbeats to indicate that it is still alive. This is done by [removing a key stored in the `stalling` set](https://github.com/bee-queue/bee-queue/blob/1f804b812470d7a439e5184c7b9af5d76b561714/lib/queue.js#L511-L520). 

So while the worker is alive (as indicated by the absence of the `jobId` in the `stalling` Redis et), the job that it is currently processing will not be marked as stalled. Stalled jobs are handled on a regular interval by calling the `checkStalledJobs` function. Only if a job is stalled, caused either by the worker dying or by loss of connectivity, then the job is returned to the `waiting` list for another worker to pick it up.

> In the parseAndProcessCsv transaction (onPreview, onChunk and onComplete), onPreview will delete all existing messages for that campaign. If multiple transactions occurred and tried to delete the messages (DELETE * FROM email_messages WHERE campaign_id = :id), it could cause a deadlock.

I believe (have not tested) this can happen now also. If there are two overlapping calls to `/complete`, there will be multiple transaction occurring trying to delete the messages. This is still possible in this implementation from repeated calls to `/complete`.

backend/src/core/services/upload.service.ts
lamkeewei: I didn't set an explicit timeout on the `bee-queue` job, so this means that if the promise is rejected, it is from the job failing. However, I think we should probably still keep this to prevent jobs from being stuck in the processing state. 

Didn't try to fix the timeout in this PR because it would need quite a bit of change. In order to handle it properly, we will need to be able to abort/interrupt the CSV processing mid-parse once the timeout is reached.

Issue #1167, opened 2021-04-29
[Backend tests] Default credentials 


miazima:
Demo campaigns and email campaigns make use of our default credentials to send out campaigns.

Write tests to ensure 
- retrieval of default credentials from db
- campaigns can be sent using default credentials

PR #1204, opened 2021-05-25
Backend tests for Credentials (Telegram/Twilio)

mjgui:
Closes [#1167]

- Implements manual mocks for `aws-sdk`, `telegraf`
- Note: Twilio hasn't been mocked because it's not needed here - the SMS tests were mocked at the SMSService level so I don't need to populate SMSTemplates and test the message hydration flow

## Tests
- `/credentials` and `/new-credentials` endpoint for Telegram and SMS
- `/settings/telegram/credentials` and `/settings/sms/credentials`

### DIFF
--- a/backend/jest.config.js
+++ b/backend/jest.config.js
@@ -9,6 +9,7 @@ module.exports = {
     '@telegram/(.*)': '<rootDir>/src/telegram/$1',
     '@test-utils/(.*)': '<rootDir>/src/test-utils/$1',
     '@shared/(.*)': '<rootDir>/../shared/src/$1',
+    '@mocks/(.*)': '<rootDir>/src/__mocks__/$1',
   },
   transform: {
     '^.+\\.(ts|tsx)$': 'ts-jest',
--- a/backend/jest.config.js
+++ b/backend/jest.config.js
@@ -11,6 +11,7 @@ module.exports = {
     '@shared/(.*)': '<rootDir>/../shared/src/$1',
     '@mocks/(.*)': '<rootDir>/src/__mocks__/$1',
   },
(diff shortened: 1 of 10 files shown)


backend/jest.config.js
```diff
@@ -9,7 +9,9 @@ module.exports = {
     '@telegram/(.*)': '<rootDir>/src/telegram/$1',
     '@test-utils/(.*)': '<rootDir>/src/test-utils/$1',
     '@shared/(.*)': '<rootDir>/../shared/src/$1',
+    '@mocks/(.*)': '<rootDir>/src/__mocks__/$1',
   },
+  modulePathIgnorePatterns: ['<rootDir>/build'],
```
mjgui, line 14: Not sure if I should do this or instead block typescript from building any mock files

lamkeewei: Sorry, could you explain more about what is it that you need to do here?

mjgui: Oops, my bad.

By default, my mock ts files (in `/src/__mock__`) get compiled into JS in `/build/__mock__`. When `jest` searches for mock modules to include, it will search both directories by default, resulting in a name collision.

The line above stops jest from looking inside `/build` for mock files to be included, but I'm not sure if i should instead add something to the `tsconfig` to stop it from building particular files (in this case mocks). The main downside of adding rules to `tsconfig` is that if I mock anything in the future in other folders (besides `/src/__mock__`), I'll need to manually exclude it too.

lamkeewei (approved):
lgtm! 

We should also exclude `__mocks__` in `tsconfig.build.json`. Though I realise the current `exclude` is also not valid (`tests/` doesn't match anything). I'll fix this in a separate PR.

PR #1214, opened 2021-06-02
fix(backend): support parentheses in email display name

lamkeewei:
## Problem

We did not explicitly install `validator` as a dependency in the backend and was using an outdated version that was a dependency of another library. This older version of `validator` did not support parentheses in email display names.

**Example runs**
- validator@10.11.0: [Runkit link](https://runkit.com/embed/igrnr5wykm8c)
- validator@13.6.0: [Runkit link](https://runkit.com/embed/4wdqn1h8v3ja)

## Solution

**Bug Fixes**:

- Explicitly install newer version of `validator`

## Tests
- Run unit tests
- Create new email campaign with custom sender name using parentheses (e.g. "User (Department)"). Email template should be saved correctly.

## Deploy Notes

**New dependencies**:

- `validator`: Used for email address validation


mjgui (changes requested):
Unit tests and saving template to DB works as expected
However, any text within the parentheses doesn't actually get sent out in the email
(e.g. `Hello (World) <donotreply@....>` is sent as `Hello <donotreply@...>`)

I've isolated this behavior to `nodemailer`. We need to the sender name in doublequotes (i.e. `"Sender (Name)" <donotreply@...>`), but since the value is taken wholesale from the frontend we'll need to do some parsing somewhere so it doesn't appear in the textbox

~~P.S. `nodemailer` actually doesn't seem to be following the relevant [rfc](https://datatracker.ietf.org/doc/html/rfc2822#appendix-A.1.2) which states that only special characters (commas, angled brackets) need to be wrapped in double quotes. Maybe we should do an upstream PR? lol~~

edit: Upon re-reading the [rfc, specifically the definition of an 'atom'](https://datatracker.ietf.org/doc/html/rfc2822#section-3.2.4), it's actually `validator` that's not following the rfc. In particular, parentheses aren't valid characters unless they're wrapped in doublequotes.

mjgui (approved):
Tested, works other than the stuff below!

**Stuff that breaks but might not be worth fixing:**

- `"User (Dept)"` (i.e. wrapped with quotes) becomes `User via Postman <...>"
- `"\"Test this out\""` passes validation in step 1 but fails when sending
- `Test \ test` (i.e. single backslash) becomes `Test test`

The solution for the above is to replace all `\` with `\\` and `"` with `\"` (backslash escape doublequote and backslash) in the `escapeFromAddress` function, but I'm not sure if that will cause any other issues

lamkeewei:
> Tested, works other than the stuff below!
> 
> **Stuff that breaks but might not be worth fixing:**
> 
> * `"User (Dept)"` (i.e. wrapped with quotes) becomes `User via Postman <...>"
> * `"\"Test this out\""` passes validation in step 1 but fails when sending
> * `Test \ test` (i.e. single backslash) becomes `Test test`
> 
> The solution for the above is to replace all `\` with `\\` and `"` with `\"` (backslash escape doublequote and backslash) in the `escapeFromAddress` function, but I'm not sure if that will cause any other issues

Yeah, was also considering adding escaping for those. But I think for now, going to keep this implementation constrained. We can expand `escapeFromAddress` function.

Issue #1216, opened 2021-06-02
HTML Email Templates


mjgui:
Currently, Postman emails are often mistaken for phishing emails as they are plaintext in nature.
As part of efforts to improve trustability of Postman emails, we should

- Append agency logos to the top of the email
- Add basic styling so emails don't look so plain

Issue #1203, opened 2021-05-25
Restructure agency collection to support agency logos


syan-syan:
probably need 3 keys: email domain, agency name and agency logo maybe a short name like formsg?

jeantanzj:
I'm not sure how we are hosting the logo. 
If we are using CID - we probably want to store the logo on disk so that we can load it into a buffer for attachment. Then the logo_uri would point to a location on disk. 
If we are using a file path (eg our own s3 bucket), then logo_uri should be a *relative* file path.
If we are using file.go.gov.sg, then logo_uri would be an *absolute* link to the file.

syan-syan:
ok to use CID since file size is small

mjgui:
Will do this with the PR for email templates

mjgui:
Currently, `agencies` table has a single column `domain`, with two possible formats:
- `.gov.sg` (wildcard), or 
- `@moe.edu.sg` (specific domain)

**Proposed Solution**

- `agencies` table will have 3 columns, `domain` (pri key), `name` and `logo_uri`
- `users` table will have new column `agency_domain` as foreign key on `agencies` table
- agency domain will be filled in with a `BeforeCreate` hook
- `agencies` table needs to be populated (manually) with all known agencies `@__.gov.sg`, 
- each `domain` will have a single row with assoicated name and logo

**Possible Issues**

- Fetching the agency/logo for a particular campaign requires either 3 queries, or a join across 3 tables (`email_templates` to `users` to `agencies`)
  - Unlikely to be that expensive since lookup is via primary key and user/agency tables are relatively small
- Currently, some users authenticate email with wildcard domain (`*@*.gov.sg`) that is not in the `agencies` table
  - Have a fallback row in `agencies` as default value for foreign key (e.g. all unrecognized gov.sg domains will point to generic gov.sg entry)
  - Might have to manually handle non gov.sg emails
  - If we choose to retroactively support a new domain, we need to backpatch all users of a particular domain to the new `agencies` entry
- There might be repetition between entries in `agencies` (e.g. `@open.gov.sg` and `@postman.gov.sg` are different rows but both have the same logo and name, Open Gov Products), resulting in potential inconsistencies in logo/name spelling
  - Might not be a significant problem since the list of domains is not very large

mjgui:
Closing this for now

**Potential extensions:**
- [ ] Having a "display name" when it needs to be displayed in a paragraph (e.g. "Supreme Court" vs "the Supreme Court")

PR #1217, opened 2021-06-02
Email template

mjgui:
## Problem

Current emails are mostly in plaintext, making them seem like phishing emails. Additionally, they carry no branding of the agency that sent the emails.

Closes [#1203, #1216]

## Solution

_Note: HTML theme is still TBC, waiting for design to be finalised_
- Implemented a HTML template for future emails which includes agency name (in the footer) and agency logo
- Updated `agencies` table to include `name` and `logo_uri` columns, add `domains` table to store mappings from domains to agencies
- Added backend logic to lookup agency details based on reply_to value
- Updated existing email preview to use same styles 

## Before & After Screenshots
_Note: Copy in screenshots is outdated_

**Desktop View**

**Mobile View**

**Rich text Edtior**

## Tests

- Send out any email - email should be sent using the new theme
- Draft a new email campaign - rich text editor should be using Arial font (in black), email preview (in step 2 and 4) should show the new theme

## Deploy Notes

- Motivation behind new schema is [here](https://github.com/opengovsg/postmangovsg/issues/1203#issuecomment-852788863)
_Note: If SQL migrations PR is deployed beforehand, will modify scripts below to use migrations instead_

**Pre-deployment Checklist**
- [ ] Remove pri key constraint from `domain` column (will be unused), add new columns to `agencies` table
```sql
ALTER TABLE agencies 
  DROP CONSTRAINT agencies_pkey,
  ALTER COLUMN domain DROP NOT NULL,
  ADD COLUMN id serial PRIMARY KEY,
  ADD COLUMN name varchar,
  ADD COLUMN logo_uri text;
```
**Note:** Do not drop `domain` column yet to prevent breaking old code

- [ ] Create new `domains` table
```sql
CREATE TABLE domains (
  domain varchar NOT NULL PRIMARY KEY,
  agency_id integer,
  created_at timestamptz,
  updated_at timestamptz,
  FOREIGN KEY (agency_id) REFERENCES agencies(id)
);
```

- [ ] Add `email_domain` column to User (FK on domain)
```sql
ALTER TABLE users
  ADD COLUMN email_domain varchar,
  ADD FOREIGN KEY (email_domain) REFERENCES domains(domain);
```

**When we want to add Logo support**
Note: For now, we can skip the following two steps (populate agencies and logos) until we finalise the logos. This PR only needs the above schema changes to run and will work even if the new tables are blank.

- [ ] Populate list of agencies and domains, populate existing users with domains

- [ ] After populating domains, backpatch existing users table with domains (if valid)
```sql
UPDATE users 
  SET email_domain = d.domain
  FROM domains d
  WHERE SUBSTR(users.email, POSITION( '@' in users.email ) )  = d.domain;
```

**After Deployment is successful (i.e. no more instances of code depending on the old schema)**
- [ ] Drop `domain` column from `agencies` table
_(Optional: Leave domain column for schema compatibility with previous versions of code)_
```sql
ALTER TABLE agencies DROP domain;
```

**New dev dependencies**:

- Added `copyfiles` package to `/shared` to copy static files (email template) to build folder - manually triggered in Dockerfile durin g build


shared/src/theme/index.ts
```diff
@@ -0,0 +1,17 @@
...
+import fs from 'fs'
+import path from 'path'
+
+const generateThemedHTMLEmail = (emailThemeFields: {
+  body: string
+  agencyName?: string
+  agencyLogoURI?: string
+  unsubLink: string
+}): string => {
+  const htmlTemplate = fs.readFileSync(path.resolve(__dirname, './email-theme.mustache'), 'utf-8')
```
lamkeewei, line 11: Might be better to use the async version of `readFile`. Synchronous read blocks the entire Node.js event loop.

mjgui: Makes sense, updated

shared/src/theme/index.ts
```diff
@@ -0,0 +1,20 @@
...
+const generateThemedHTMLEmail = async (emailThemeFields: {
+  body: string
+  agencyName?: string
+  agencyLogoURI?: string
+  unsubLink: string
+}): Promise<string> => {
+  const htmlTemplate = await fs.promises.readFile(
+    path.resolve(__dirname, './email-theme.mustache'),
+    'utf-8'
+  )
```
mjgui, line 14: Actually, should I move this out as a global constant? Seems wiser to only read the file once during initial run instead of for every message.
Or alternatively, refactor this entire file as a singleton class so it will get loaded a max of 1 time but will not be loaded if it isn't used.

lamkeewei: Makes sense. Though slight preference for the second method. Ideally, importing a module should not result in any side effect.

backend/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN aws configure set default.region ap-southeast-1
 WORKDIR /usr/home/postmangovsg
 
 COPY shared ../shared
-RUN cd ../shared && npm ci && cd ../postmangovsg
+RUN cd ../shared && npm ci
```
lamkeewei, line 16:
```suggestion
RUN cd ../shared && npm ci && npm run build
```

This should work too (haven't tested). Originally I didn't have to run `npm run build` in `shared` because the tsc command in `backend` will build all project references also. But in this case since there's post-build actions might be easier to explicitly run build.

backend/src/email/services/email.service.ts
```diff
@@ -61,11 +65,33 @@ const getHydratedMessage = async (
...
+        include: [
+          {
+            model: Domain,
+            include: [Agency],
+          },
+        ],
+      },
+    ],
+  })
+  const agency = campaign?.user?.domain?.agency
```
lamkeewei, line 84: Is there a situation where agency name or agency logo will be `undefined` and the hydrated message will still be considered valid? If not, it would be good to do a check here and throw an error if they are undefined. 

This way you can remove the `undefined` type for `agencyName` and `agencyLogoURI`. Benefits are that callers of this method no longer need to check for `undefined` from the return value and it better encodes the domain.

backend/Dockerfile
mjgui: Yep I was thinking about this, just that this will result in `shared` being built twice. I doubt it will have significant cost, but not sure if calling `npm run build` (and hence implicitly calling the postbuild script) is more readable or if running `postbuild` is more readable

backend/src/email/services/email.service.ts
mjgui: Actually yes, the database will allow for `NULL` values (which I guess will become `undefined` through sequelize?).

For `agencyLogoURI` for example, we're currently planning to leave it blank for agencies who don't have suitable logos (and potentially for the generic government agency too)

For `agencyName`, a useful side effect for leaving it optional as well is that if either `domain` or `agency` is undefined (such as for users in the current DB before backpatching), the new code will run perfectly too

shared/src/theme/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    // Take first 200 characters of the body (without html tags) to use as preheader (preview text of email content)
+    // Append whitespace after closing tags so paragraphs are spaced out in preheader
+    const { body } = emailThemeFields
+    const preheader = xss
+      .filterXSS(body, {
+        whiteList: {},                  // whitelist for tags is empty, i.e. strip all HTML tags 
+        stripIgnoreTagBody: ['script'], // remove contents of script tags entirely
+        onIgnoreTag: (_tag, _html, options) => {
+          // @ts-ignore types are missing, upgrade to js-xss 1.0.9 which fixes missing types
```
mjgui, line 36: had to leave a @ts-ignore here because the types for `js-xss` library are incomplete, see [this issue](https://github.com/leizongmin/js-xss/issues/224)

lamkeewei:
@mjgui sequelize migrations have been merged in. Would be good to change to use migrations instead of the manual SQL.

lamkeewei:
Probably want to use a different color for the placeholder text.

image

mjgui:
Removed all DB schema changes (and rebased on the latest `develop` branch) at #1242 
Will re-add schema changes after confirmation on implementation of logos

PR #1219, opened 2021-06-03
feat: support use of domain verification for custom from address

### LINKED ISSUE #1208: Support verification of custom from address using verified domain
Currently as part of the custom from verification process, we lookup the Dkim attributes [using the email address](https://github.com/opengovsg/postmangovsg/blob/develop/backend/src/email/services/custom-domain.service.ts#L50-L57) as the identity. As we move towards verifying the entire domain, we should also allow using the domain as the identity used for looking up.

We are preserving the existing email address verification to support pre-existing custom from addresses that do not have verified domains.

Reference: [GetIdentityDkimAttributes](https://docs.aws.amazon.com/ses/latest/APIReference/API_GetIdentityDkimAttributes.html)

lamkeewei:
## Problem

Closes #1208 

## Solution
**Features**
- Allow use domain verification for custom from address. For e.g. if agency.gov.sg is already verified, user@agency.gov.sg can be added without further verification on SES.

## Tests
Run unit tests to ensure that different scenarios works as expected. Email address with custom domain for testing is available in 1Password under "Domain verification test email".

Before each of this scenario, it's recommended to delete the domain/email address on SES first.

**Unverified domain and verified domain**
- On SES, ensure that:
   - Domain is not verified
   - Email address is verified
- Add custom from address through the Settings
- Verification should succeed and you should be able to send out a campaign using the custom from address

**Verified domain and unverified email**
- On SES, ensure that:
   - Domain is verified
   - Email address is not verified
- Add custom from address through the Settings
- Verification should succeed and you should be able to send out a campaign using the custom from address

**Verified domain and verified email**
- On SES, ensure that:
   - Domain is verified
   - Email address is verified
- Add custom from address through the Settings
- Verification should succeed and you should be able to send out a campaign using the custom from address


backend/src/core/utils/email-address.ts
```diff
@@ -0,0 +1,24 @@
+import psl from 'psl'
+
+/**
+ * Extract the domain of an email address
```
mjgui, line 4: Maybe it's more readable to explicitly mention that the extracted domain excludes the subdomain?
(If I didn't understand the code wrongly)
```suggestion
 * Extract the domain of an email address, excluding its subdomain
 * e.g. person@sub.agency.gov.sg -> agency.gov.sg
```

lamkeewei:
Discussed and merge if needed based on user demand

PR #1220, opened 2021-06-03
Standardise sql whitespace

mjgui:
## Problem

Only Whitespace changes
- replace all tabs with two spaces in SQL files
- remove trailing whitespace in SQL files
- fix random tab in config.ts

_Not sure if we should use tabs or spaces? I chose spaces in the end but we can easily change it I guess_


jeantanzj (approved):
Thank you

PR #1221, opened 2021-06-07
Add space after email salutation in the email template body

### LINKED ISSUE #1149: Automatically add a space after the first para
Cosmetic change allowing emails to appear more professional and legitimate in message previews. This is P1 as it could affect perception of email legitimacy.

**Problem**: the first para of an email is usually the salutation, which shows up in message previews. Without any special effort by the Postman user, the email could end up looking unprofessional or even phishy in the message preview today (bottom preview) due to no space being added after the first para, as compared to the top preview which is what most emails look like.

Screenshot 2021-04-26 at 5 37 39 AM

miazima:
## Problem

When there is no space following the email salutation, in templates such as below, the email preview tends to look phishy.

```
Dear form admin,You are receiving this email because
```

Closes #1149 

## Solution

**Improvements**:

- Add a space after the first comma in the email template, ensuring that there is a space followed by words
- Make this change after sanitising the template body 

e.g.
```
Dear form admin, You are receiving this email because
```

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- Enter template `Dear form admin,You are receiving this email because` 
- Save template and observe the preview to see that there is now a space after the email salutation

### DIFF
--- a/backend/src/email/services/email-template.service.ts
+++ b/backend/src/email/services/email-template.service.ts
@@ -179,7 +179,7 @@ const storeTemplate = async ({
 
   // Add whitespace, if not present, after email salutation
   // to prevent phisy looking email preview
-  sanitizedBody = sanitizedBody.replace(/(^.+?,)([^\s].+?<\/p>.+)/, '$1 $2')
+  sanitizedBody = sanitizedBody.replace(/(^.+?,)([^\s].+?<\/p>.*)/, '$1 $2')
 
   // Append via to sender name if it is not the default from name
   const {
--- a/backend/src/email/services/email-template.service.ts
+++ b/backend/src/email/services/email-template.service.ts
@@ -170,12 +170,16 @@ const storeTemplate = async ({
 }: StoreTemplateInput): Promise<StoreTemplateOutput> => {
   // extract params from template, save to db (this will be done with hook)
   const sanitizedSubject = client.replaceNewLinesAndSanitize(subject)
-  const sanitizedBody = client.filterXSS(body)
+  let sanitizedBody = client.filterXSS(body)
   if (!sanitizedSubject || !sanitizedBody) {
     throw new TemplateError(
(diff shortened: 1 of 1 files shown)


backend/src/email/services/email-template.service.ts
```diff
@@ -170,13 +170,17 @@ const storeTemplate = async ({
...
+  let sanitizedBody = client.filterXSS(body)
   if (!sanitizedSubject || !sanitizedBody) {
     throw new TemplateError(
       'Message template is invalid as it only contains invalid HTML tags!'
     )
   }
 
+  // Add whitespace, if not present, after email salutation
+  // to prevent phisy looking email preview
+  sanitizedBody = sanitizedBody.replace(/(^.+?,)([^\s].+?<\/p>.*)/, '$1 $2')
```
lamkeewei, line 182: This regex seems to have an issue if the next block after email salutation is not a `p` tag. See [here](https://regex101.com/r/J2YEpo/1) for example.

lamkeewei (changes requested):
Regex don't account for cases where the following block is not a `p` tag. Also might be good to extend out tests to cover different cases for this new functionality.

miazima:
> Actually, I had to separately implement this functionality while implementing the preheader feature as part of stripping the content of all HTML tags ([see this block](https://github.com/opengovsg/postmangovsg/pull/1217/files#diff-2527711556c19ec518189ee7b3162dc95c0de2605c84373bad4fca69252b170aR28-R41)), so we might no longer need this regex

@mjgui Oh was working on correcting the regex and adding tests. Perhaps I can help write tests for your implementation?

mjgui:
Oh yeah sure! Though I probably need to refactor the method into a separate function so it's more testable haha

lamkeewei:
@mjgui and @miazima is this PR still required?

PR #1231, opened 2021-06-10
fix(backend): extend keepalive timeout

lamkeewei:
## Problem

Node.js default `keepAliveTimeout` is [set to 5 seconds](https://nodejs.org/docs/latest-v12.x/api/http.html#http_server_keepalivetimeout). This is shorter than the idle timeout for ALB. This creates the potential for a race condition where ALB tries to send data through a connection that is in the processing of being closed by Node.js at the same time. This will result in a [tcp reset packet](https://www.pico.net/kb/what-is-a-tcp-reset-rst) to be sent, causing the "connection reset by peer" error we see on nginx error logs.

Can refer to this [article](https://shuheikagawa.com/blog/2019/04/25/keep-alive-timeout/) for a more detailed explanation.

## Solution

**Bug Fixes**:

- Set `keepAliveTimeout` to 110 seconds (greater than 100 seconds).

## Tests
The easiest way to test this is to manually make a HTTP call to the backend server using `nc` (netcat) and observe that after the response is sent, the connection remains open for the specified time (101 seconds). 

After starting up the backend, run the following in your terminal:
```bash
$ nc localhost 4000
GET / HTTP/1.1

HTTP/1.1 200 OK
...
Connection: keep-alive
```

You should observe that netcat does not immediately exit but instead wait for the timeout to be reached before closing the connection. 

You can repeat the same test with the increased timeout commented out and observe that the connection closes after 5 seconds.


stanleynguyen:
@prestonlimlianjie let's release this as well since it's a small fix

stanleynguyen (approved):

PR #1242, opened 2021-06-21
Email template (without logos)

### LINKED ISSUE #1216: HTML Email Templates

mjgui:
## Problem

Based on #1217 but without the db schema changes to support logos (as the exact mechanism for logo attachments has yet to be confirmed)

Closes #1216 (but **not** #1203, unlike my other PR, as no db schema changes are made)

## Solution

- Implemented a HTML template for future emails
- Updated existing email preview to use same styles 
- Conditionally show .gov.sg masthead depending on user

## Before & After Screenshots
_Note: Logos are not in this PR_

**Desktop View**

**Mobile View**

**Rich text Edtior**

## Tests

- Send out any email - email should be sent using the new theme
- Draft a new email campaign - rich text editor should be using Arial font (in black), email preview (in step 2 and 4) should show the new theme

## Deploy Notes

**Migration**
```
npm run db:migrate
```
(Specifically, `backend/src/database/migrations/20210621043226-create-get-messages-to-send-email-with-sender-function.js` needs to be run)

**Upgraded dependencies**:
- Bumped version of `xss` to 1.0.9 so typescript types are fixed

**New dev dependencies**:

- Added `copyfiles` package to `/shared` to copy static files (email template) to build folder (have also manually triggered this in Dockerfile for worker and backend)


shared/src/theme/index.ts
```diff
@@ -0,0 +1,69 @@
...
+   */
+  static generatePreheader(body: string): string {
+    return xss
+      .filterXSS(body, {
+        whiteList: {},
+        stripIgnoreTagBody: ['script'],
+        onIgnoreTag: (_tag, _html, options) => {
+          // Append whitespace after each closing tag
+          // @ts-ignore types are missing, upgrade to js-xss 1.0.9 which fixes missing types
+          if (options.isClosing) return ' '
```
lamkeewei, line 40: Might want to upgrade the library rather than using `ts-ignore`

frontend/src/components/common/rich-text-editor/RichTextEditor.module.scss
```diff
@@ -308,4 +308,49 @@
...
+  // These styles are repeated in EmailPreviewBlock.module.scss due to differences in DOM structure
+  color: #000000;
+  font-size: 14px;
+
+  * {
+    font-family: Arial, Helvetica, sans-serif;
+  }
+
+  h1 {
+    font-size: 28px !important;
```
lamkeewei, line 323: Generally, we should avoid `!important` unless we really do not have a way to specifically target the component you want to style. Overuse of `!important` would make it hard to reason about what styles will apply since the usual CSS order of precedence are no longer respected. 

In this case, I think `!important` can be avoided by moving the CSS blocks into `.editor`.

In addition, the styles can be extracted into a separate file and shared using `@extend` and placeholder selectors (similar to our typography.scss). This ensure that there's only one place where the styles need to be modified

worker/src/core/loaders/message-worker/email.class.ts
```diff
@@ -51,27 +53,30 @@ class Email {
...
-      subject: string
-      replyTo: string | null
-      from: string
-      campaignId: number
-    }[]
-  > {
-    return this.connection
-      .query('SELECT get_messages_to_send_email(:job_id, :rate);', {
+  async getMessages(jobId: number, rate: number): Promise<Message[]> {
+    interface MessageWithSenderEmail extends Message {
```
lamkeewei, line 57: Might want to move the interface definition somewhere else. Maybe `src/email/interfaces/index.ts`?

lamkeewei (changes requested):
Some other feedback:
1. Should update the tests for email campaign routes to take into account the new `themedBody` field.
2. Tests for `ThemeClient` would be useful in the long run. Particularly having tests for `generatePreheader` and `generateThemedBody` would help better define the expected behaviour and clarify their purpose.

frontend/src/components/common/email-preview-block/EmailPreviewBlock.tsx
```diff
@@ -34,19 +36,32 @@ const EmailPreviewBlock: FC<EmailPreviewBlockProps> = ({
...
-      <h5>Subject</h5>
-      <p>{subject}</p>
+        <h5>Subject</h5>
+        <p>{subject}</p>
 
-      <h5>Body</h5>
-      <RichTextEditor value={body} preview {...otherProps} />
+        {
+          // In case themedBody doesn't exist, render body as fallback instead
+          themedBody != undefined || (
```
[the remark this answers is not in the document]
mjgui: Oh I tried this too, but this doesn't work because react will render `themedBody` if it's not null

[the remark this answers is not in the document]
mjgui: Also, maybe I should just delete the whole block
I left it here to maintain compatibility with the old backend but maybe that's not important?

worker/src/core/loaders/message-worker/email.class.ts
mjgui: Currently `senderEmail` is only used here (to check for the masthead) so it feels odd to separate this specific type from its usage

For now I've inlined the interface declaration (using an intersection type) so it's more clear that it's a temporary thing

frontend/src/components/common/email-preview-block/EmailPreviewBlock.tsx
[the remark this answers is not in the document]
lamkeewei: Can removed. We will usually deploy the backend first then the frontend.

lamkeewei (approved):
lgtm! Tested the following on staging and works as expected:
- Database migration to create a new SQL function
- Email test and actual messages use email template
- Duplicated campaigns will use email template
- Rich text editor follows style of template
- Preview block shows template

@mjgui squash and merge whenever.

lamkeewei:
@mjgui were we also gonna remove the dark mode meta tags?

mjgui:
> @mjgui were we also gonna remove the dark mode meta tags?

Yepyep, just remembered this right after we ended the call

lamkeewei:
@mjgui tests seems to be broken because of the new preview component. Other changes looks ok.

PR #1250, opened 2021-06-24
feat: email open tracking

### LINKED ISSUE #1236: Email read receipts
- Add email read receipts as an option for each email campaign, design TBD
- Add read status to the delivery report: @miazima can we check what copy other tools use? e.g. "Opened" vs "Unopened" or "Unconfirmed" as should align as far as possible to the norm

miazima:
## Problem

Email read receipts enhance the experience of Postman and is a frequently requested feature.

- Provides user a email read metric to let users know the reach of their campaign and the effectiveness of using email as a channel of communication
- Provides users a way of knowing if the recipient has read the email, so that they can reach the recipient through other channels if needed

Closes #1236

## Solution

**Features**:

- Setup AWS SES configuration set to receive a SNS event callback when the recipient opens the email
- Handle incoming open events and update the status of the message in the db
- Introduce `READ` status to represent emails which have been opened 

## Tests

- Deploy branch to staging, ensure workers are deployed too
- Send out an email campaign to yourself
- Ensure successful delivery is recorded in staging db for campaign messages with status `SUCCESS`
- Open the email and then check staging db for campaign messages with status `READ`
- Ensure that delivery and bounce still works as expected

## Deploy Notes
- ~Create configuration set and SNS topic and subscription for prod~
- SES configuration set (`postman-email-open`) created for both production SES region

- `WORKER_SES_CONFIGURATION_SET` : Name of the AWS SES configuration set. Defaults to `postman-email-open`


backend/src/core/constants.ts
```diff
@@ -17,6 +17,7 @@ export enum MessageStatus {
   Sending = 'SENDING',
   Error = 'ERROR',
   Success = 'SUCCESS',
+  Read = 'READ',
```
lamkeewei, line 20: Was there a need to manually add `READ` as a possible value in `enum_email_messages_status`? If so, would be good to add a database migration to do this.

miazima: I added it here as the `email_messages` table `status` field references `MessageStatus` for the allowed enum values. Is that what you mean?
https://github.com/opengovsg/postmangovsg/blob/54d46e6e0b7f80b34a5cd5fcacdec0e9c554c876/backend/src/email/models/email-message.ts#L43-L47

lamkeewei: Since we don't use `sequelize.sync` anymore, I'm not sure if this will automatically update the Postgresql enum named `enum_email_messages_status`. We might have to manually modify the enum by running:

```sql
ALTER TYPE enum_email_messages_status ADD VALUE 'READ';
```

This should be added as a sequelize database migration under `backend/src/database`.

lamkeewei:
@miazima could I get your help to resolve the merge conflicts? Thanks!

lamkeewei (approved):

PR #1253, opened 2021-06-25
fix(verify-backup): increase timeout to 15m

lamkeewei:
## Problem

As backup size increase, more time is required to run the backup verification.

## Solution

**Improvements**:

- Increase timeout for Cloud Run.


lamkeewei:
Closing as increasing timeout alone wasn't enough to solve the issue.

PR #1261, opened 2021-06-29
feat: update footer copy for emails

mjgui:
## Changes

Updated footer copy for emails

> Alternatively, you may also contact the sender agency or organisation directly through their public contact channels ~~for updates to contact details or email preferences~~


lamkeewei (approved):
Lgtm!

PR #1262, opened 2021-06-30
fix(shared): clean build and reference build config in other modules

lamkeewei:
## Problem
- Other modules were not correctly referring to the build tsconfig under project references. 
- `tsc --build --clean` command was not referring to the right build tsconfig

## Solution

**Improvements**
- Reference build tsconfig in other modules
- Reference build tsconfig in clean command

## Test
- Run `npm run dev` consecutively, error should not occur as the clean command now clears the right file

### DIFF
--- a/shared/package.json
+++ b/shared/package.json
@@ -5,7 +5,7 @@
   "main": "build/index.js",
   "scripts": {
     "dev": "npm run postbuild && tsc --build tsconfig.build.json -w",
-    "build": "rimraf build && tsc -b --clean && tsc --build tsconfig.build.json",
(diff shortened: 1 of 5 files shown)


shared/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "build/index.js",
   "scripts": {
     "dev": "npm run postbuild && tsc --build tsconfig.build.json -w",
-    "build": "rimraf build && tsc -b --clean && tsc --build tsconfig.build.json",
+    "build": "rimraf build && tsc --build tsconfig.build.json --clean && tsc -p tsconfig.build.json",
```
mjgui, line 8: not very important, but any reason why we're using `--build` over `-b`?

mjgui (approved):
lgtm!

PR #1263, opened 2021-06-30
Email template logos

mjgui:
## Problem

Continuation of #1242 

Closes [#1203, #1216]

## Solution

- Add agency logo to HTML template
- Updated `agencies` table to include `name` and `logo_uri` columns, add `domains` table to store mappings from domains to agencies
- Added backend logic to lookup agency details
- Added backend logic to populate domain and agency for new users (NOTE: does not backfill values for existing users)

## Before & After Screenshots
_Note: Copy in screenshots is outdated_

**Desktop View**

**Mobile View**

**Show/Hide logo**
image

**Show/Hide logo if logo is not uploaded**
image

## Tests

- Build `backend` code
- Run ```npx sequelize db:migrate``` from `backend`
- Email creation form should show 'Your agency’s logo has not been uploaded. To upload a logo, please fill up this form.' for agency without logo (e.g. before populating DB)

---

- Populate `domains` and `agencies` database with agency/logo details
- Email template preview in steps 2 and 4 - should show/hide logo based on option in step 1
- Send test email (step 3) - should show/hide logo based on option in step 1
- Send email campaign (step 4) - should show/hide logo based on option in step 1 

## Deploy Notes

- Motivation behind new schema is [here](https://github.com/opengovsg/postmangovsg/issues/1203#issuecomment-852788863)
```
cd backend && npx sequelize db:migrate
```
- [ ] Populate list of agencies and domains
- [ ] After populating domains, backpatch existing users table with domains (if valid)
```sql
UPDATE users 
  SET email_domain = d.domain
  FROM domains d
  WHERE SUBSTR(users.email, POSITION( '@' in users.email ) )  = d.domain;
```

## Rollback
DB migrations are backwards compatible, no rollback is required (though undoing the migrations will do no harm)

**NOTE:** Frontend MUST be rolled back before backend.
New frontend is not compatible with old backend due to the new field `showLogo` (it will fail the validator)


mjgui:
The commit history is kinda messy becaused I based this branch off #1217 
Will fix when squashing and merging...

backend/src/database/migrations/20210630070237-add-columns-to-agencies.js
```diff
@@ -0,0 +1,47 @@
...
+    })
+    await queryInterface.addColumn('agencies', 'logo_uri', {
+      type: Sequelize.DataTypes.TEXT,
+    })
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    // Drop the whole table and re-create it, as NULL values for domain are
+    // not valid for the previous schema anyway
+    await queryInterface.dropTable('agencies')
```
lamkeewei, line 30: Just a comment, I guess this is ok since agency table is empty now.

lamkeewei (changes requested):
Let's add some tests for the changes before merging this in.

jeantanzj:
```
Jean  12 minutes ago
my concern is that we might create the domain nillywilly then validate the domain (edited) 

Jean  11 minutes ago
is it possible to put the domain .gov.sg in the domains table for validation, instead of using the env var? (edited) 

Gui Ming Jiang  11 minutes ago
I believe both hooks need to pass for the user to be created (edited) 

Jean  10 minutes ago
if populateDomain runs before validateEmail, there’s no check on whether the email ends with .gov.sg . we could put them in the same function for strict ordering? btw the hooks dont currently run in a transaction (edited) 

Gui Ming Jiang  9 minutes ago
Hm.. I suppose adding .gov.sg to the DB works, but it might not be clear to whoever's reading the code that that's the intention

Gui Ming Jiang  7 minutes ago
Oh I thought by passing in options.transaction into findOrCreate it should tie it up with the user transaction

Gui Ming Jiang  7 minutes ago
Then if validateEmail throws an error, it will stop the user creation action

Jean  4 minutes ago
prior to the addition of the agencies table, there was no need for a transaction because we weren’t looking up the db, just the env var.
right now validateDomain(email) should have options.transaction passed in as well.
I’m not sure if I am getting my point across, or if I’ve misinterpreted the code:
If validateEmail runs before populateDomain, when someone enters blah@abc.com , we would reject abc.com
If populateDomain runs before validateEmail, we would create abc.com in domains, then validate the email, and consequently accept abc.com (edited) 

Jean  2 minutes ago
the part about adding .gov.sg to the domains table is a separate issue - it’s so that we don’t have too many entrypoints to keep track of. Or at least remove specific domains from the env var?
```

jeantanzj:
@mingjiang  https://www.postgresql.org/docs/9.5/transaction-iso.html
>13.2.1. Read Committed Isolation Level
Read Committed is the default isolation level in PostgreSQL. When a transaction uses this isolation level, a SELECT query (without a FOR UPDATE/SHARE clause) sees only data committed before the query began; it never sees either uncommitted data or changes committed during query execution by concurrent transactions. In effect, a SELECT query sees a snapshot of the database as of the instant the query begins to run. **However, SELECT does see the effects of previous updates executed within its own transaction, even though they are not yet committed.** Also note that two successive SELECT commands can see different data, even though they are within a single transaction, if other transactions commit changes after the first SELECT starts and before the second SELECT starts.

jeantanzj:
@lamkeewei 
that's true, we should change https://github.com/opengovsg/postmangovsg/blob/dde8fe82edddf8533b6ad587fdb17833bc9dc635/backend/src/core/services/auth.service.ts#L265-L269

backend/src/core/models/domain.ts
```diff
@@ -0,0 +1,55 @@
...
+  @Column(DataType.INTEGER)
+  agencyId?: number
+
+  @BelongsTo(() => Agency)
+  agency?: Agency
+
+  // When a domain is created (i.e. a new user's domain is not recognised)
+  // By default, find (or create) agency with default name and populate foreign key with that value
+  @BeforeCreate
+  static async populateAgency(
```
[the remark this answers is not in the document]
lamkeewei: Let's combine this with the `validateEmail`. We shouldn't rely on the implicit ordering of the execution of the hooks. In addition, the same transaction should also be passed to `validateDomain`. My bad for not seeing there was an additional hook above.

backend/src/core/models/user/tests/user.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+afterEach(async () => {
+  await User.destroy({ where: {} })
+  await Domain.destroy({ where: {} })
+  await Agency.destroy({ where: {} })
+})
+
+afterAll(async () => {
+  await sequelize.close()
+  await RedisService.shutdown()
```
mjgui, line 27: Btw, the Redis clients aren't shutting down for me locally (though Travis still passes it)...
Not sure how to fix this

mjgui: Also, not sure if I've covered all the cases for this test

backend/src/core/utils/validate-domain.ts
```diff
@@ -45,7 +62,7 @@ const validateDomain = async (
     )
   } else {
     logger.info({
-      message: 'Domains whitelisted',
+      message: 'Domains whitelisted in env var',
       domainsToWhitelist,
       action: 'validateDomain',
```
mjgui, line 67: In the end I didn't remove this log since presumably the env var should be quite short

frontend/src/components/dashboard/create/email/EmailTemplate.tsx
```diff
@@ -280,9 +280,13 @@ const EmailTemplate = ({
...
               <p>
                 <label htmlFor="logo">
-                  Your agency’s logo will appear in the header of the email
+                  Your agency’s logo will appear in the header of the email.
                 </label>
               </p>
+              <p>
+                <b>Note:</b> Logos may not be visible in government email
+                inboxes.
+              </p>
```
mjgui: Actually not sure if it's a concern that the `<label>` tag is only in the first para

backend/src/core/utils/validate-domain.ts
```diff
@@ -21,13 +22,37 @@ const isValidDomain = (domain: string): boolean => {
...
+): Promise<boolean> => {
+  if (!validator.isEmail(email)) return false
+
+  // First, check if there exists an exact match in the domains table
+  const emailDomain = email.substring(email.lastIndexOf('@'))
+  const dbDomain = await Domain.findOne({
+    where: { domain: emailDomain },
+    transaction,
+  })
+  if (dbDomain != null) {
```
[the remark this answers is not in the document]
mjgui: Tbh I don't know LOL, but if `sequelize-typescript` is trusted I guess it doesn't matter

lamkeewei (approved):
lgtm! Tested the following scenarios:

- User with `email_domain`, `agency` and `domain` configured should be shown their agency logo and can send mail including it
- Toggling show logo to “off” should not include logo in sent email
- User with `email_domain = NULL` should not have access to option to include logo option. Sending should work as expected.
- User with `email_domain` set to a domain linking to an agency with an empty `logo_uri` should not have access to option to include logo
- User whose domain does not exist yet, should have it’s domain automatically created on login

Issue #1289, opened 2021-07-28
Links aren't rendered correctly in email preview


mjgui:
**Describe the bug**
In the email theme preview on the frontend, links aren't rendered with the correct background colour
Observed by @nauynix

**To Reproduce**

1. Create new campaign
2. Insert link in the content of the email body, observe that it is blue + underlined
2. Click on next. Observe that links are black and not underlined

**Expected behavior**
Links should be blue + underlined

**Screenshots**
<img width="523" alt="Screenshot 2021-07-28 at 4 06 51 PM" src="https://user-images.githubusercontent.com/9080115/127287242-47483ac9-9d70-47c6-8e68-09d083204043.png">

<img width="670" alt="Screenshot 2021-07-28 at 4 07 47 PM" src="https://user-images.githubusercontent.com/9080115/127287373-37bc6788-5e28-4f11-8c5d-50ea79f921de.png">

**Desktop (please complete the following information):**

- OS: MacOS
- Browser: Chrome

**Additional context**
I need to reset the styles for <a> tags properly for frontend

PR #1290, opened 2021-07-28
fix: reset link styles in html email preview

mjgui:
## Problem

Fix link previews not having the correct styles

Closes [#1289]

## Solution

- Used `color: revert` to reset color for a tags
- Added manual overriding `color: blue; text-decoration: underline;` to ensure backwards compat

## Screenshots
image

## Tests

- [x] Add link to email template, ensure that it is blue and underlined

## Deploy Notes
NA


lamkeewei (approved):
lgtm! Learned something new today!

PR #1291, opened 2021-08-04
fix(frontend): text wrapping for messages sent on tablet

lamkeewei:
## Problem

Text wrapping for "messages sent" is off for certain screen sizes (tablet).

## Solution

**Bug Fixes**:
- Apply margin on the container instead for the message count.

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

## Tests
- Replace the count to a large number (chrome inspector) and observe that the text wrapping looks normal when in tablet sized screen width


lamkeewei:
@syan-syan this looks ok right?

syan-syan:
> @syan-syan this looks ok right?

looks good! thanks

PR #1340, opened 2021-11-12
fix: deployment

jeantanzj:
## Problem

Dockerfile was not building due to
1) docker hub rate limit
```
An error occurred during execution of command [self-startup] - [Docker Specific Build Application]. Stop running the command. Error: failed to pull docker image: Command /bin/sh -c docker pull node:12-alpine failed with error exit status 1. Stderr:Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```
2) missing package (I believe alpine was updated to remove python2. pip does not come with python3. Previously, even though we installed python3, we were using pip2 to install packages)

database-backup was failing 
```Cannot open: https://download.postgresql.org/pub/repos/yum/11/redhat/rhel-7-x86_64/postgresql11-libs-11.5-1PGDG.rhel7.x86_64.rpm. Skipping.
Error: Nothing to do
Loaded plugins: ovl, priorities
Cannot open: https://download.postgresql.org/pub/repos/yum/11/redhat/rhel-7-x86_64/postgresql11-11.5-1PGDG.rhel7.x86_64.rpm. Skipping.
Error: Nothing to do
which: no pg_dump in (/var/lang/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/bin)
cp: missing destination file operand after '/var/task/bin'"
```  

## Solution

- Pushed node:12-alpine to our public ECR. Caveat: public ECR has a rate limit of 1 image pull/s. But we spin up very few servers so I think this is still workable. 
- Installed pip in Dockerfile
- Used amazon-linux-extras in place of the yum repo to install postgresql


lamkeewei (approved):
1. Triggering backup lambda with test event
2. Launch new instances of backend on Elastic Beanstalk (no more rate limit)

Had a minor issue with build for `backend` and `frontend` failing to build on Travis with a `Z_BUF_ERROR`. Resolved it by clearing Travis cache.

PR #1358, opened 2022-01-31
chore: pin node version to 14

prestonlimlianjie:
### Pin node version to 14

This PR pins the node version of the codebase for a better developer experience. The _node-sass_ library fails to install with node 16 and above. We will revisit this issue later (since we should be using the lts version of node) - but for now, we will pin the node version to 14.

### Update `.secrets.baseline`
Additionally, the pre-commit hook updated the `.secrets.baseline` file.


stanleynguyen:
We actually only need to pin this for the frontend sub-repo (the other 2 work fine with LTS). Do we want to pin for all subrepos?

prestonlimlianjie:
> We actually only need to pin this for the frontend sub-repo (the other 2 work fine with LTS). Do we want to pin for all subrepos?

I'm thinking that the developer workflow will be
```
nvm use # uses the .nvmrc file in root
npm install
npm run dev
```

My thought behind setting the .nvmrc file in the root folder is that we currently perform `npm run build` or `npm run dev` from that root package.json file (which then runs `npm run build/dev` in the frontend/backend/worker folders). I don't think we can get npm to run nvm since nvm is a binary that isn't managed by npm.

stanleynguyen (dismissed):

stanleynguyen:
[sass](https://www.npmjs.com/package/sass) is a pure JS implementation with no native bindings, and from what I've seen it's a drop-in replacement

PR #1360, opened 2022-02-04
build(eslint): update eslint configs to latest

stanleynguyen:
Update this to resolve all pre-commit check issue before fixing backend test

PR #1361, opened 2022-02-04
test(backend): fix open handles caused by redis

stanleynguyen:
**Note**: pending #1360 to be merged first

- problem: currently we're initializing redis clients in the main
program of redis.service, which hoist the variables in the global
scope through importing. As a result, the file handle opened by
`createClient` function call doesn't get closed when the application
scope exits

- solution: have a dedicated init function inside redis.service so
we will initialize redis clients in the application scope specifically

- some misc changes:
  - update of sequelize version to latest
  - deprecate body-parser in favor of native express middlewares

### DIFF
--- a/backend/src/email/middlewares/tests/email.middleware.test.ts
+++ b/backend/src/email/middlewares/tests/email.middleware.test.ts
@@ -2,9 +2,9 @@ import { NextFunction, Request, Response } from 'express'
 import { Sequelize } from 'sequelize-typescript'
 import { Campaign, User } from '@core/models'
 import sequelizeLoader from '@test-utils/sequelize-loader'
-import { RedisService } from '@core/services'
 import { ChannelType } from '@core/constants'
 import { EmailMiddleware } from '@email/middlewares/email.middleware'
+import { CreationAttributes } from 'sequelize/dist'
 
 let sequelize: Sequelize
 let campaignId: number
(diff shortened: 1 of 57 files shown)


backend/src/email/middlewares/tests/email.middleware.test.ts
```diff
@@ -21,22 +21,25 @@ beforeEach(() => {
...
+  } as CreationAttributes<Campaign>)
   campaignId = campaign.id
 })
 
 afterAll(async () => {
   await Campaign.destroy({ where: {} })
   await User.destroy({ where: {} })
   await sequelize.close()
-  await RedisService.shutdown()
+  // await RedisService.shutdown()
```
prestonlimlianjie, line 42: As you pointed out in our offline discussion, we can remove the commented out lines

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -36,7 +37,7 @@ afterAll(async () => {
   await User.destroy({ where: {} })
   await sequelize.close()
 
-  await new Promise((resolve) => RedisService.rateLimitClient.flushdb(resolve))
+  await new Promise((resolve) => RedisService.rateLimitClient?.flushdb(resolve))
   await RedisService.shutdown()
```
prestonlimlianjie, line 41: Should be deleted as well

prestonlimlianjie (approved):
lgtm apart from the 2 nits

stanleynguyen:
Circling back on this @prestonlimlianjie, it seems that this error (attached below) seems to crash the program out of the open handles, hence "fixing" the open handles. But once I fixed the attached error, the open handles from redis seems to come back. I'm investigating further

Screenshot 2022-02-08 at 5 15 01 PM

stanleynguyen:
Open handle when redis is completely removed

Screenshot 2022-02-08 at 5 33 02 PM

PR #1366, opened 2022-02-07
build(sass): replace node-sass with sass

stanleynguyen:
sass-loader starts supporting sass automatically since v7.2.0
https://github.com/webpack-contrib/sass-loader/releases/tag/v7.2.0
and we are alreading using v10. Hence, this is more or less a drop-in
replacement.

closes #1358


prestonlimlianjie (approved):
lgtm

PR #1388, opened 2022-02-15
test(backend): fix redis client open handles

### LINKED ISSUE #1387: Research into how to fix open handles issues for Redis

stanleynguyen:
- root cause: Previously, we create redis client right in the main
body of redis service file. This means that every import of redis
service will create a new set of redis clients. As a result, any
attempt to import RedisService to call RedisService.shutdown() will
only close the clients created by that immediate import instead of
closing all clients that might be imported from another file.

- solution: switch RedisService (and all related services that utilise
it) to an on-demand instance model that will be dependency-injected
into other services/middlewares/routes that utilise them. This way we
will only create exactly one set of redis clients and can safely add and
export a cleanup function from where they're initialised

resolves #1387 

Possible pending refactor as a result of this PR: switching the whole project to a DI model (even for those less obvious ones unlike redis service) and consolidate initialisation to a single places (Will create ticket once discussed)

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,11 +1,11 @@
 {
-  "generated_at": "2021-06-25T08:45:30Z",
+  "version": "1.0.3",
   "plugins_used": [
     {
-      "name": "AWSKeyDetector"
+      "name": "ArtifactoryDetector"
     },
     {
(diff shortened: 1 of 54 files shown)


prestonlimlianjie (approved):
lgtm - tested locally

PR #1408, opened 2022-02-21
hotfix(frontend): fix buffer polyfill missing in react-scripts v5

stanleynguyen:
Tested in staging


prestonlimlianjie (approved):
lgtm

stanleynguyen:
@prestonlimlianjie I was thinking whether for hotfix direct to master we should actually bump the patch version and make a new release instead of updating existing one?

PR #1413, opened 2022-03-01
feat(backend): interruptible recipients list uploads

stanleynguyen:
Replacing #1202 with rebased commits (created a new branch so we don't accidentally lose changes)


stanleynguyen:
Further context from @lamkeewei 

- there's a time limit before beanstalk get force killed. This time limit was too short to ensure that we can handle all uploads before the process is force killed.
Given this, this means that we need to store state of in progress upload processing and have another process pick it back up. That’s kind of how I landed on the queue and restart method.
- Also, seems like some part of beanstalk does not forward SIGINT or SIGKILL - https://stackoverflow.com/questions/34726888/process-signals-to-do-a-graceful-exit-with-a-nodejs-app-on-elastic-beanstalk

stanleynguyen:
@prestonlimlianjie this feature is complete and ready to be deployed. Shall we include it in the next release? I've tried it in my local and it seems to work fine

prestonlimlianjie (approved):
Originally written by Kee Wei, rebased and tested by Stanley

stanleynguyen:
Deployment checklist:
- [ ] Add `REDIS_UPLOAD_URI` to env vars

Issue #1438, opened 2022-03-22
Automatically run database migrations and seeds when running local dev


prestonlimlianjie:
Add sequelize db migration and seed step to `npm run dev`

PR #1439, opened 2022-03-22
build(backend): auto run db migrate and seed in npm run dev

prestonlimlianjie:
Resolves https://github.com/opengovsg/postmangovsg/issues/1438


backend/package.json
```diff
@@ -5,15 +5,17 @@
   "main": "build/server.js",
   "scripts": {
     "build": "rimraf build && tsc --build tsconfig.build.json",
-    "dev": "npm run postbuild && tsc-watch --onSuccess \"node ./build/server.js\"",
+    "dev": "npm run database:migrate && npm run database:seed && npm run postbuild && tsc-watch --onSuccess \"node ./build/server.js\"",
```
stanleynguyen, line 8: Do we actually want to run seed on every dev server bootup though? Because the seeds might clash with existing prior seeds and cause error. (Not sure how your local setup is but I have a persistent volume for my local postgres docker and keep reusing it)

prestonlimlianjie: Ah that's fair. Maybe we don't need this PR then.

prestonlimlianjie:
Reopened this PR because I realized that we still didn't have good instructions on how to run migrations when we were onboarding @yingyee0111 onto the project.

tan-zx (approved):

PR #1530, opened 2022-05-12
refactor(backend): remove cloudwatch log streaming

stanleynguyen:


stanleynguyen:
Waiting til end of Q2 for team to get used to the new system before switching over completely

prestonlimlianjie:
We would like to store logs in Cloudwatch because we can set the log retention to 10 years, and so this PR isn't necessary. This is helpful for audits.

PR #1532, opened 2022-05-18
fix(email-template): force table style

### LINKED ISSUE #1486: Issues with tables on outlook
User can't see the outline of tables on Outlook

stanleynguyen:
Resolves #1486 

the `alpha` property in CSS `rgba` is not supported by the underlying browser engine of Outlook => switching to use a hex code without alpha


prestonlimlianjie (approved):
lgtm

Issue #1428, opened 2022-03-14
Cypress Setup


stanleynguyen:
Context: Smoke tests are currently manually carried out by the engineer who deploy new releases

DoD:
- [ ] Cypress test setup with a "hello world" test case
- [ ] Automated Github Action to run the test post-deployment on:
  - [ ] staging
  - [ ] production

prestonlimlianjie:
Started working on this - ran into problems because my machine uses the M1 Mac chip. :(

Will attempt to hack my way around this.

> Currently, the way to run Cypress on Apple M1 ARM chip is by using Intel emulation with Rosetta 2, Apple's translation layer.

https://www.cypress.io/blog/2021/01/20/running-cypress-on-the-apple-m1-silicon-arm-architecture-using-rosetta-2/

stanleynguyen:
Resolved in #1538

PR #1538, opened 2022-05-19
feat: added cypress basic test

yingyee0111:
## Problem

Manual smoke test is required at deployment
Part of #1428 

## Solution

Use cypress to automate smoke test on staging

**Features**:

- added dev dependency and script to cypress
- cypress/integration/email.spec.js: runs through email campaign
- cypress/integration/email_encrypted.spec.js: runs through encrypted email campaign
- cypress/integration/sms.spec.js: runs through sms campaign
- created github action to run cypress (gmail api uses credentials uploaded to github secret)

## Deploy Notes
**Issue**:
- yet to integrate GitHub actions

**New scripts**:

- "cy:run": "npx cypress run" (Use: npm run cy:run -- -s <script>)

**New dev dependencies**:

- "cypress": "^9.6.1"
- "cypress-file-upload": "^5.0.8"
- "gmail-tester": "^1.3.4"


cypress/integration/email.spec.js
```diff
@@ -54,7 +54,7 @@ describe('Email Test', () => {
 
         //step 2 : upload csv file
         cy.get('input[type="file"]').attachFile(CSV_FILENAME)
-        cy.contains('Message preview', {timeout: 10000})
+        cy.contains('Message preview', {timeout: 50000})
```
prestonlimlianjie, line 58: Would be great to have these magic numbers as constants

package.json
```diff
@@ -53,6 +54,10 @@
   "homepage": "https://github.com/datagovsg/postmangovsg#readme",
   "devDependencies": {
     "concurrently": "^7.0.0",
+    "cypress": "^9.6.1",
```
stanleynguyen, line 57: I'm thinking whether we should have cypress tests in its own module (something like what we did with `backend`, `frontend`, etc) since it's a standalone test suite

yingyee0111: rearranged

stanleynguyen (approved):

PR #1541, opened 2022-05-24
feat(logging): add logger for incoming http requests

stanleynguyen:
Related to [this change](https://github.com/opengovsg/postmangovsg/blob/aa1ee3d95bf05e2a82ac144df293c592e810195e/backend/src/core/logger.ts#L21), I've opened #1550


prestonlimlianjie (approved):
lgtm

PR #1567, opened 2022-06-04
feat(recipient-upload): prevent duplicate jobs by setting id

stanleynguyen:


prestonlimlianjie (approved):
lgtm! could we add a comment before line 267 to explain why we add a `setId()` call

tan-zx:
the docs said to avoid a numeric job id — maybe

image

Issue #1572, opened 2022-06-06
add .nvmrc to standardise on Node version


tan-zx:
Think we should standardise on Node 16 instead of 14?

PR #1576, opened 2022-06-09
build: add nvmrc file for node 16.15.1

tan-zx:
Closes https://github.com/opengovsg/postmangovsg/issues/1572

Is this the right Node version?


stanleynguyen:
Should we pin to latest as of now `16.15.1`?

tan-zx:
OK fixed!

stanleynguyen (approved):

PR #1578, opened 2022-06-10
build: bump node version in ci and pin node mod versions

stanleynguyen:


tan-zx (approved):

tan-zx:
also opportunistically upgraded dependencies flagged up by Snyk and modify ci for frontend because of pending changes to Lottie (https://github.com/Gamote/lottie-react/issues/61) → to revert once upgrade is in

Gamote:
> thinking behind freezing dependencies:
> 
> * automatically updating dependencies based on SemVer has resulted in broken peer dependencies → as such, we have frozen all the dependencies to avoid this issue
> * instead of relying on `npm audit`, which is very noisy, we have decided to rely on Snyk for fixing security vulnerabilities instead. as such, have opportunistically upgraded those dependencies
> 
> to make this work, we modified the ci for frontend to run with `--legacy-peer-deps` option because there is a broken dependency to Lottie ([Gamote/lottie-react#61](https://github.com/Gamote/lottie-react/issues/61)) → to revert once this upgrade is in

@tan-zi the #61 issue was fixed in the `v2.3.0` [release](https://github.com/Gamote/lottie-react/releases/tag/v2.3.0).

tan-zx:
Thanks @Gamote!

Just tried to run `npm i` and it seems like there are other unresolved dependencies that prevent us from removing `--legacy-peer-deps` just yet.

image

tan-zx:
To update pending: 
- https://github.com/jpuri/react-draft-wysiwyg/pull/1245

Issue #1579, opened 2022-06-10
update cost per SMS on SMS message editor


lisatjide:
**Describe the bug**
Current pricing shown on editor is SGD 0.0395 but the pricing stated on Twilio's pricing page is in USD
<img width="896" alt="Screenshot 2022-06-10 at 3 29 28 PM" src="https://user-images.githubusercontent.com/23736580/173013638-b120cb16-b979-4cb5-9ca0-49386134ef2d.png">
<img width="661" alt="Screenshot 2022-06-10 at 3 25 03 PM" src="https://user-images.githubusercontent.com/23736580/173013863-51aa5310-f8bd-482d-803e-91071ad612e0.png">

However on our SMS message editor, it is currently SGD 0.0395.

**How to fix**
The accurate pricing on Postman editor should be 1.4*0.0395=0.0553. 
We can set the default exchange rate of 1USD=1.4SGD for now.

tan-zx:
will work on this! instead of hard-coding the price ($0.0395), thinking of fetching it from Twilio's API instead (found [this](https://www.twilio.com/docs/sms/api/pricing#messagingcountry-pricing-properties)) 

will use the default exchange rate and change the text to let user know the approximate exchange rate used! (haha thought about fetching the exchange rate via API but I think that's unnecessary complexity)

PR #1582, opened 2022-06-11
fix: fix displayed SGD pricing in Campaign SMS

tan-zx:
closes https://github.com/opengovsg/postmangovsg/issues/1579

Have used the backend to fetch Twilio pricing via API, so if they change prices, this would be automatically updated. ~~This requires loading env variables with Twilio credentials — not sure if the way I implemented it is correct; in theory, we should be able to use wtv credentials that was used to send the Demo SMS campaign?~~

Updated to use demo credentials. Exchange rate is hard-coded via an env variable — for future, can consider changing it to a cached API call to account for currency fluctuations.

***

Deployment checklist:

- [ ] Add `USD_TO_SGD_RATE` env variable → only if there is significant change in exchange rate; default of 1.4 provided


tan-zx:
not quite sure how to fix the backend test! @stanleynguyen

backend/src/core/config.ts
```diff
@@ -83,6 +83,39 @@ const config = convict({
       sensitive: true,
     },
   },
+  twilio: {
```
stanleynguyen, line 86: Instead of adding new env vars, we can use the demo creds. An example of how to use that [here](https://github.com/opengovsg/postmangovsg/blob/a55db4e1a60ddb5ee76eaf29db83d0440da15802/backend/src/sms/middlewares/sms.middleware.ts#L163-L170)

frontend/src/components/dashboard/create/common/BodyTemplate/BodyTemplate.tsx
```diff
@@ -9,26 +8,37 @@ import styles from './BodyTemplate.module.scss'
...
 } from 'components/common'
 import SaveDraftModal from 'components/dashboard/create/save-draft-modal'
 import { CampaignContext } from 'contexts/campaign.context'
 import { FinishLaterModalContext } from 'contexts/finish-later.modal.context'
 
-const SmsMessageBodyInfo = ({ body }: { body: string }) => {
-  const COST_PER_TWILIO_SMS_SEGMENT_IN_SGD = 0.0395 // correct as at 5 Feb 2022
+// correct as at 12 Jun 2022
+export const COST_PER_SMS = 0.0395
+export const USD_SGD_EXCHANGE_RATE = 1.4
```
stanleynguyen, line 23: Can we return the result in SGD? the conversion should be done in the backend either:
- Using a configurable backend env var to set the exchange rate
- Making an API call to somewhere to retrieve live exchange rate

stanleynguyen (changes requested):

backend/src/core/config.ts
tan-zx: ohh sick, let me do that

frontend/src/components/dashboard/create/common/BodyTemplate/BodyTemplate.tsx
tan-zx: that makes sense, will change!

backend/src/sms/routes/tests/sms-campaign.routes.test.ts
```diff
@@ -88,8 +100,6 @@ describe('POST /campaign/{campaignId}/sms/credentials', () => {
     expect(res.body).toEqual({
       message: `Campaign cannot use demo credentials. ${DefaultCredentialName.SMS} is not allowed.`,
     })
-
-    expect(mockSecretsManager.getSecretValue).not.toHaveBeenCalled()
```
tan-zx, line 92: demo credentials is now used for Twilio API call

tan-zx:
have rebased on develop — hope for a review :) @stanleynguyen @prestonlimlianjie

tan-zx:
tested on staging — works

backend/.env-example
```diff
@@ -59,7 +58,7 @@ FRONTEND_URL="http://localhost:3000"
 # TWILIO_API_KEY=""
 # TWILIO_ACCOUNT_SID=""
 
-DD_ENV="local"
-DD_SERVICE="postman"
-DD_LOGS_INJECTION="true"
-DD_TRACE_ENABLED="false"
\ No newline at end of file
+# DD_ENV="local"
```
prestonlimlianjie, line 61: Are these intentionally commented out?

tan-zx: Yup I was cleaning up the env variables in `backend/.env-example` such that optional env vars are commented out but compulsory env vars are not (so someone setting it up from scratch will know which is which)

which reminds me, I forgot to add the optional `USD_TO_SGD_RATE` env variable

tan-zx: added! https://github.com/opengovsg/postmangovsg/pull/1582/commits/3ed55112b30ef9a39e9a1ce2618f8965de5acc83

backend/src/sms/routes/tests/sms-campaign.routes.test.ts
```diff
@@ -88,8 +102,8 @@ describe('POST /campaign/{campaignId}/sms/credentials', () => {
...
+    // deleted the following test because demo credentials used to call Twilio API for SMS price
```
prestonlimlianjie, line 105: Could remove the comment + the expect line

prestonlimlianjie (approved):
lgtm - had a nit on removing the comment

PR #1583, opened 2022-06-11
build: slight change to frontend package-lock.json after npm ci

tan-zx:
@stanleynguyen changes seem pretty minor, probably sth to do with my Node version? (16.11.1)


stanleynguyen:
This should be because of the npm version. Not sure what happened there but let's try with the npm version that accompanies latest node `16.15.1` first?

tan-zx:
I think you're right — package-lock doesn't change with `16.15.1` I think

PR #1585, opened 2022-06-11
chore: turn off useUnknownInCatchVariables

tan-zx:
## Problem
Getting these type errors in my IDE
SCR-20220611-vys

## Solution

According to [this](https://stackoverflow.com/questions/68240884/error-object-inside-catch-is-of-type-unkown), one solution is to set `"useUnknownInCatchVariables": false`.

Another solution is to set the `instanceof Error` like this:
image

Have gone for the former in this PR


tan-zx:
oops realized it's caused by the wrong TS version

PR #1586, opened 2022-06-13
Docs/cypress local setup

yingyee0111:
Added .env-example and README.md for cypress


tan-zx (changes requested):
lgtm — need to make sure the `.md` file follows Prettier before merging though

PR #1597, opened 2022-06-14
refactor: optimize callback authentication

### LINKED ISSUE #1437: Reduce CPU utilisation when verifying authenticity of webhook requests

**Twilio callbacks**
Currently, when sending out SMSes using Twilio, we use `bcrypt` to generate the password for our callback URLs.

https://github.com/opengovsg/postmangovsg/blob/648c3da5e700fc6e96d580f18db28f8ad3872e37/worker/src/sms/services/twilio-client.class.ts#L79

This means that for each webhook callback that we receive we will need to carry out a relatively CPU intensive comparison to verify that the provided password is correct. 

https://github.com/opengovsg/postmangovsg/blob/648c3da5e700fc6e96d580f18db28f8ad3872e37/backend/src/sms/services/sms-callback.service.ts#L26

This is the main bottleneck for our callback servers when large SMS campaigns are sent out. 

**SES callbacks**
For email campaigns, the bottleneck is at the signature validation for each delivery status SNS message from SES. This involves a quite expensive asymmetric crypto operation. We've profiled this previously and identified this as the bottleneck. 

https://github.com/opengovsg/postmangovsg/blob/648c3da5e700fc6e96d580f18db28f8ad3872e37/backend/src/email/utils/callback/parsers/ses.ts#L103

For Twilio, I think there are two ways to solve this:
1. Use a cheaper hash function (SHA1/2) to hash the password
2. Get rid of the password authentication completely and just rely on [Twilio's `X-Twilio-Signature` header](https://www.twilio.com/docs/usage/webhooks/webhooks-security#validating-signatures-from-twilio)

For SES:
- We might want to consider if we can use a simpler method of verifying the authenticity of the webhook request. Perhaps we can also rely on a HMAC in a custom SMTP header.

stanleynguyen:
Resolves #1437 

Follow up with fix to #1598 

Deployment checklist:
- [x] Add `EMAIL_CALLBACK_HASH_SECRET` to env vars
  - [x] backend
  - [x] worker
  - [x] redaction-digest
  - [x] unsubscribe-digest


backend/src/sms/services/sms-callback.service.ts
```diff
@@ -23,7 +24,14 @@ const isAuthenticated = (
   const [username, password] = credentials.split(':')
   const plainTextPassword =
     username + messageId + campaignId + config.get('smsCallback.callbackSecret')
-  return bcrypt.compareSync(plainTextPassword, password)
+
+  let isAuthenticated = getSha256Hash(plainTextPassword) === password
+  // if not passed with the new hash, retry with the old way
+  // TODO: remove this after all campaigns sent with the old way have completed
+  if (!isAuthenticated) {
+    isAuthenticated = bcrypt.compareSync(plainTextPassword, password)
```
tan-zx, line 32: just curious: was under the impression that we should always use `bcrypt`'s method asynchronously, so as to prevent holding up the event loop

stanleynguyen: This is inside the context of a request, which is already handled async

backend/src/email/utils/callback/parsers/ses.ts
```diff
@@ -183,10 +181,37 @@ const parseEvent = async (
...
+  if (!id) {
+    logger.info({
+      message: 'No reference message id found',
+      id,
+      timestamp: record.Timestamp,
+    })
+    throw new Error('No reference message id')
+  }
+
+  const password = id + config.get('emailCallback.hashSecret')
```
lamkeewei, line 197: Rather than using the secret this way, why not generate a proper HMAC.

```ts
const secret = config.get('emailCallback.hashSecret')
const hash = crypto.createHmac('sha256', secret).update(`${id}`).digest('hex')
```

backend/src/email/utils/callback/parsers/ses.ts
```diff
@@ -183,10 +181,37 @@ const parseEvent = async (
 }
 
 const parseRecord = async (record: SesRecord): Promise<void> => {
-  if (!(record.SignatureVersion === '1' && (await validateSignature(record)))) {
-    throw new Error(`Invalid record`)
-  }
   const message = JSON.parse(record.Message)
+  const {
```
lamkeewei, line 185: Minor nit; this validation chunk here is quite lengthy. For readability, might want to extract this into a separate function. Something also to consider is whether we can introduce stronger types here too.

backend/src/sms/services/sms-callback.service.ts
```diff
@@ -23,7 +24,21 @@ const isAuthenticated = (
   const [username, password] = credentials.split(':')
   const plainTextPassword =
     username + messageId + campaignId + config.get('smsCallback.callbackSecret')
-  return bcrypt.compareSync(plainTextPassword, password)
+
+  let isAuthenticated = getSha256Hash(plainTextPassword) === password
```
lamkeewei, line 28:
```suggestion
  let isAuthenticated = crypto.timingSafeEqual(
      getSha256Hash(plainTextPassword),
      password
  )
```

Though very unlikely, when doing string equality comparison for crypto, should use [constant time comparison functions](https://nodejs.org/docs/latest-v16.x/api/crypto.html#cryptotimingsafeequala-b) to prevent [timing attacks](https://paragonie.com/blog/2015/11/preventing-timing-attacks-on-string-comparison-with-double-hmac-strategy).

backend/src/email/utils/callback/parsers/ses.ts
```diff
@@ -183,10 +181,37 @@ const parseEvent = async (
...
+    logger.info({
+      message: 'No reference message id found',
+      id,
+      timestamp: record.Timestamp,
+    })
+    throw new Error('No reference message id')
+  }
+
+  const password = id + config.get('emailCallback.hashSecret')
+  if (hash !== getSha256Hash(password)) {
```
lamkeewei, line 198: Use `crypto.timingSafeEqual` here too.

lamkeewei (commented):
Left a couple of comment. Caveat is that I did not test/run the code!

backend/src/sms/services/sms-callback.service.ts
```diff
@@ -21,9 +22,25 @@ const isAuthenticated = (
 
   const credentials = Buffer.from(secret, 'base64').toString('utf8')
   const [username, password] = credentials.split(':')
-  const plainTextPassword =
-    username + messageId + campaignId + config.get('smsCallback.callbackSecret')
-  return bcrypt.compareSync(plainTextPassword, password)
+  const plainTextPassword = username + messageId + campaignId
```
lamkeewei, line 25: Could be useful to put this format into `shared` for easier modification in future.

lamkeewei (approved):
lgtm! Trust you will test in staging before deploying this.

Minor note is that if there's a bug here in this PR, rollback is not an option if you don't want to drop any callback requests. If safety is preferred, I guess you could have both mechanism running in parallel, with the new HMAC based verification being optional. Only when you are confident then remove the old method.

But tbh, I think not worth the time.

backend/src/sms/services/sms-callback.service.ts
stanleynguyen: This is only applicable for sms (duplicated once only) so I prefer keeping it here

stanleynguyen:
> lgtm! Trust you will test in staging before deploying this.
> 
> Minor note is that if there's a bug here in this PR, rollback is not an option if you don't want to drop any callback requests. If safety is preferred, I guess you could have both mechanism running in parallel, with the new HMAC based verification being optional. Only when you are confident then remove the old method.
> 
> But tbh, I think not worth the time.

I'm actually logging on failure of verification using the new method (with fallback on the old one) so we can monitor when there's no further callbacks with the old signature. Then I will deprecate the old one 1 week after. Should be safe enough imo

prestonlimlianjie (approved):
skimmed it and it looks good - thanks for testing it on staging

PR #1599, opened 2022-06-14
docs: add comment to clariy manual override of halting

tan-zx:
small change to clarify what @lamkeewei mentioned today

### DIFF
--- a/backend/src/core/models/campaign.ts
+++ b/backend/src/core/models/campaign.ts
@@ -83,6 +83,7 @@ export class Campaign extends Model<Campaign> {
   @HasOne(() => Statistic)
   statistic?: Statistic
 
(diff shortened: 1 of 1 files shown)


backend/src/email/utils/callback/query.ts
```diff
@@ -185,6 +185,7 @@ export const haltCampaignIfThresholdExceeded = async (
         await EmailBlacklist?.sequelize?.transaction(async (transaction) => {
           const [numUpdated] = await Campaign.update(
             { halted: true },
+            // this allows for manual override of halting: if halted is set to null, campaign will not be halted
```
stanleynguyen, line 188: A better place to put this comment is in the model rather than here

tan-zx: ok have put it there

stanleynguyen (approved):

PR #1601, opened 2022-06-15
docs: instructions to run cypress locally

yingyee0111:
added instructions to run cypress locally


tan-zx (commented):
lgtm — think this is pending changes to the env variables? or will that be a separate PR

tan-zx (approved):
lgtm

PR #1625, opened 2022-06-22
chore: update lottie-react version and note --legacy-peer-deps

stanleynguyen:
- update lottie-react version
- made clear that devs should refer to https://github.com/opengovsg/postmangovsg/pull/1578 for further pending upgrades
- modified the default npm i behavior at the root to install frontend with --legacy-peer-deps option (think @yingyee0111 was - running into this issue when she's experimenting with different libraries on the frontend)
- have added comment in package.json following [this convention](https://stackoverflow.com/a/56098420) (probably not ideal, but better than not having any comments?)


stanleynguyen:
Replacing https://github.com/opengovsg/postmangovsg/pull/1611 for the new deployment workflow

tan-zx (approved):

PR #1633, opened 2022-06-23
ci(deploy): restrict deployment concurrency to max 1 at a time

stanleynguyen:
## Problem

Currently multiple deployments can be started at the same time and renders latest deployments failed because infrastructures might be in an `update` state

## Solution

Set max concurrency for deploy workflows to 1

## Deployment Checklist

N/A

### DIFF
--- a/.github/workflows/deploy-eb.yml
+++ b/.github/workflows/deploy-eb.yml
@@ -5,6 +5,8 @@ on:
       - staging
       - master
 
+# Making sure that the current deployment is completed before the next one
+concurrency: deploy-eb-${{ github.ref }}
 env:
   # Update this common config
   DIRECTORY: backend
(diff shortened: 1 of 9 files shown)


.github/workflows/deploy-eb.yml
```diff
@@ -5,6 +5,8 @@ on:
       - staging
       - master
 
+# Making sure that the current deployment is completed before the next one
+concurrency: deploy_eb
```
lamkeewei, line 9: Might want to prefix this with the branch name `<staging | master>_deploy_eb`. So that staging deploys don't block production deploys.

lamkeewei (approved):
Minor comment, lgtm otherwise!

.github/workflows/deploy-eb.yml
stanleynguyen: Oh yea lemme fix that

PR #1638, opened 2022-06-28
fix: actions/checkout version

tan-zx:
version 3.4.2 doesn't exist


tan-zx:
passed! https://github.com/opengovsg/postmangovsg/actions/runs/2573429624

stanleynguyen (approved):

Issue #1645, opened 2022-06-29
Unsub link not clickable for yahoo mail users


yingyee0111:
**Describe the bug**
The unsub link that is supposed to be embedded on the word "here" is greyed out and not clickable for yahoo mail recipients.

**Expected behavior**
Clicking on the word "here" should lead to the unsub page.

**Screenshots**
<img width="712" alt="Screenshot 2022-06-29 at 9 41 45 PM" src="https://user-images.githubusercontent.com/33883582/176451145-a34cb859-63c7-4a57-9144-502ae0f0a28f.png">

tan-zx:
@yingyee0111 seems like it won't auto-close if you put a bracket around the issue haha
<img width="121" alt="image" src="https://user-images.githubusercontent.com/67887489/176586050-3e059fe1-0831-481a-a796-b9d73f351be0.png">

PR #1646, opened 2022-06-29
fix: do not encode unsub link

yingyee0111:
## Problem

Unsub link was not clickable for yahoo mail recipients.
This is because `Mustache.render()` encodes the unsub link when filling in the footer template:

> `please click <a href="{{unsubLink}}">here</a> `

Yahoo mail can't decode the link properly and hence has it greyed out.

Closes [#1645]

## Solution

Surround the variable with triple brackets so that Mustache does not encode the link.

> `please click <a href="{{{unsubLink}}}">here</a> `

Note: this does not break anything.
Encoding of the link is unnecessary, the embedded Postman link in the footer was not encoded.

> `Powered by <a href="https://postman.gov.sg">Postman</a>`

## Deployment Checklist

N/A

### DIFF
--- a/shared/src/theme/email-theme.mustache
+++ b/shared/src/theme/email-theme.mustache
@@ -203,7 +203,7 @@
           agency or organisation that sent this to you.
           <br />
           <br />          
-          If you wish to unsubscribe from similar emails, please click <a href="{{unsubLink}}">here</a> to inform the sender of your email preference.
+          If you wish to unsubscribe from similar emails, please click <a href="{{{unsubLink}}}">here</a> to inform the sender of your email preference.
           Alternatively, you may also contact the sender agency or organisation directly through their public contact channels.
           <br />
           <br />


tan-zx:
I haven't seen this part of the codebase before but is the difference between `{{var}}` and `{{{var}}}` to do with unescaping HTML? Curious why this problem only occurs with Yahoo Mail hmm.

also noticed that in line 184 there is `{{{body}}}`. wonder if this poses any security risks 🤔 (see e.g. [here](https://stackoverflow.com/questions/1912501/unescape-html-entities-in-javascript) and [here](https://security.stackexchange.com/questions/141666/is-it-dangerous-to-serve-unescaped-user-input-as-html-response-to-the-same-user))

@stanleynguyen @prestonlimlianjie

yingyee0111:
Yup, it unescapes html. 
Correct me if I'm wrong, my understanding is that our text editor is supposed to act as a filter of the user input so that malicious code doesn't get saved into `body`. We take in the user input, convert it into draftjs (our library) data type and generate safe html to set in body. @tan-zx

tan-zx:
ah I see, lgtm!

stanleynguyen:
> Yup, it unescapes html. [Yahoo! Mail Does Not Render Anchor Tags With Encoded HREF Attributes](https://www.bennadel.com/blog/4115-yahoo-mail-does-not-render-anchor-tags-with-encoded-href-attributes.htm) Correct me if I'm wrong, my understanding is that our text editor is supposed to act as a filter of the user input so that malicious code doesn't get saved into `body`. We take in the user input, convert it into draftjs (our library) data type and generate safe html to set in body. There's also a XSS sanitisation that happens after this. @tan-zx

To further clarify, the frontend santization is not what we're relying on, we're relying on a sanitization in our backend. So the `body` portion should be safe to not use mustache to escape HTML

stanleynguyen (approved):

PR #1647, opened 2022-06-30
feat(frontend): use new standardised ogp footer

### LINKED ISSUE #1507: use ogp footer
postman's footer is not the standard ogp footer. once ogp footer is updated, we should use it

stanleynguyen:
## Problem

Closes #1507 

## Solution

Updated for both landing page and login page

Screenshot 2022-06-30 at 1 28 32 PM

Screenshot 2022-06-30 at 1 28 43 PM

## Deployment Checklist

N/A


sarahsalim:
@stanleynguyen landing page footer background should be #F7F9FE
for login page, please do not proceed with new footer yet. let's implement new design + new footer when ready.

Issue #1644, opened 2022-06-29
remove custom domain page on postman


lisatjide:
**Describe the bug**
Not so much of a bug but a gap that we need to fix for consistency. 

With the release of custom sender name, postman has been encouraging users to use custom sender instead of custom domain. hence we should remove the custom domain page on settings page for now, while we figure out how we want to make this feature more accessible and easy to setup. 

Keeping this page will mean that we still welcome custom domain requests and it is confusing to the users since ops often try to discourage them to going this route. 

<img width="1169" alt="Screenshot 2022-06-29 at 3 29 36 PM" src="https://user-images.githubusercontent.com/23736580/176378024-8003349d-47b0-460a-b224-3596a7ef352a.png">

lisatjide:
Also remove button on the settings empty state
<img width="915" alt="image (4)" src="https://user-images.githubusercontent.com/23736580/176610295-11c50c07-e523-4537-8b73-107bc1904051.png">

PR #1648, opened 2022-06-30
feat: remove button to link to custom domain page

tan-zx:
Closes https://github.com/opengovsg/postmangovsg/issues/1644

Only removes for users who have yet to set it up; users who've set it up should not be affected


tan-zx:
before & after for users looking at `renderEmptySettings`:
image

image

before & after for users looking at `renderSettings`

image

image

yingyee0111 (approved):
lgtm!

Issue #1649, opened 2022-06-30
glitch on settings page


lisatjide:
**Describe the bug**
There is a (very) quick flash ⚡ of empty settings page. See video below

https://user-images.githubusercontent.com/23736580/176619537-8d2e9f0b-c448-482c-a628-e4338ea2910f.mov

**To Reproduce**
You must make sure that your account has either on one of the credentials set up (API, SMS Twilio or custom sender). Click on settings on the main nav. 

**Expected behavior**
The quick flash of empty state should not be there.

PR #1655, opened 2022-07-03
fix: remove split second lag

tan-zx:
Closes https://github.com/opengovsg/postmangovsg/issues/1649

Can @lisatjide help to test on [staging](https://staging.postman.gov.sg/login) that the glitch is gone? Have pushed to staging.

@stanleynguyen Haha I think the way I fixed it is pretty hacky (I went to look into whether to use React Query etc. but it's all a bit too much effort...)

### DIFF
--- a/frontend/src/components/dashboard/settings/Settings.tsx
+++ b/frontend/src/components/dashboard/settings/Settings.tsx
@@ -51,23 +51,17 @@ const SETTINGS_LINKS_WITH_EMAIL = [
 const Settings = () => {
   const modalContext = useContext(ModalContext)
   const [isLoading, setIsLoading] = useState(true)
-  const [fetchedUserSettings, setFetchedUserSettings] = useState(false)
-  const [fetchedCustomFromAddresses, setFetchedCustomFromAddresses] = useState(
-    false
-  )
   const [hasApiKey, setHasApiKey] = useState(false)
   const [hasCustomFromAddresses, setHasCustomFromAddresses] = useState(false)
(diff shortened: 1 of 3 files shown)


frontend/src/components/dashboard/settings/Settings.tsx
```diff
@@ -50,6 +50,11 @@ const SETTINGS_LINKS_WITH_EMAIL = [
 
 const Settings = () => {
   const modalContext = useContext(ModalContext)
+  const [isLoading, setIsLoading] = useState(true)
+  const [fetchedUserSettings, setFetchedUserSettings] = useState(false)
+  const [fetchedCustomFromAddresses, setFetchedCustomFromAddresses] = useState(
+    false
+  )
```
stanleynguyen, line 57: We don't need 3 different states to manage the loading for this one, just `isLoading` is enough and less confusing. We can do that by changing the `useEffect` on line 63-66 to
```ts
useEffect(() => {
  setIsLoading(true)
  Promise.all([
    fetchUserSettings(),
    fetchCustomFromAddresses(),
  ]).finally(() => setIsLoading(false))
}, [])
```

stanleynguyen (changes requested):

frontend/src/components/dashboard/settings/Settings.tsx
tan-zx: oh shit that makes sense, let me fix that

stanleynguyen (approved):

PR #1656, opened 2022-07-03
style: upgrade and apply prettier

tan-zx:
noticed frontend prettier is quite outdated; have updated it to brought it to versions used elsewhere:
image

UPDATE: have updated the backend and worker prettier, so that it's `2.6.2` everywhere


stanleynguyen (approved):
You might want to check if there're other big PRs waiting to be merged and get those in first before this one to avoid complicated conflicts

tan-zx:
I think should be fine right — just needa run `npm run lint` to apply the changes, if any (nothing substantive anyway?)

PR #1665, opened 2022-07-05
feat(sending): allow 2 email campaigns to share the same credential

stanleynguyen:
## Problem

- Sometimes very large email campaigns will hog the worker and stop other email campaigns from being sent out.
- We want to solve the above problem while still not letting a single type of campaigns (i.e. email) hog the whole sending pipeline

## Solution

Option 3 from
https://docs.google.com/document/d/1c-ZjILpnl8BDQFm1wsUGbVT_xhRKlqu85IkBIz0BtDA/edit

## Deployment Checklist

- [x] (Consider) raising email default send rate to 225 (i.e. our max cap 450 divided by max 2 concurrent workers working on emails)
- [ ] Run latest migration


backend/src/database/migrations/20210429045222-create-is-credential-used-function.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
       'boolean',
       'plpgsql',
       `
-    SELECT TRUE INTO result 
+    SELECT TRUE INTO result_non_email
```
tan-zx, line 11: since the original `is_credential_used` is a db function created via a migration, we would need to create a new db migration to modify it?

backend/src/database/migrations/20210429045222-create-is-credential-used-function.js
```diff
@@ -17,13 +17,32 @@ module.exports = {
...
+    AND c1.type <> 'EMAIL'
     LIMIT 1;
+
+    SELECT count(*) INTO count_email_campaign_using
+    FROM campaigns c1 WHERE 
+    c1.id <> campaignId
+    AND c1.cred_name = (SELECT c.cred_name FROM campaigns c WHERE c.id = campaignId LIMIT 1)
+    AND c1.type = 'EMAIL';
+
+    SELECT
```
tan-zx, line 29: to summarise, `result` returns `TRUE` if:
- This credential is being used for 2 or more email campaigns; OR (is the >2 necessary? hmm)
- This credential is being used for another non-email campaign

(would it be helpful to include the above as comments within the SQL?)

backend/src/database/migrations/tests/is-credential-used.test.ts
```diff
@@ -0,0 +1,198 @@
+import { Sequelize } from 'sequelize-typescript'
```
tan-zx, line 1: wow this looks cool! this is like the first db function test in our codebase.

wonder if there is a better place to put it rather than the `tests` folder

backend/src/database/migrations/20210429045222-create-is-credential-used-function.js
tan-zx: I wonder what would be a good way to manage this though → ideally, we should have a single source of truths for the db functions (I have been using database browsers to look at the db function code, but I must say the current usage of db functions in the code makes tracing very difficult)

stanleynguyen: Oh yea u're right. Nice catch

stanleynguyen: > I wonder what would be a good way to manage this though → ideally, we should have a single source of truths for the db functions (I have been using database browsers to look at the db function code, but I must say the current usage of db functions in the code makes tracing very difficult)

Actually none of our db functions has been updated so far so I just refer to the original migration. It's not very difficult as of now as the migrations can be the single source of truth. We're planning to migrate the functions away from sql functions so let's not worry too much about this

backend/src/database/migrations/20210429045222-create-is-credential-used-function.js
stanleynguyen: > is the >2 necessary?

What do you mean?

> would it be helpful to include the above as comments within the SQL?

There're tests so that's better than comments

tan-zx: > is the >2 necessary?

Oh I mean will there be instances where a credential is used for 3 email campaigns? Probably not if the code is working properly haha. Just thinking out loud.

tan-zx (approved):
lgtm!

backend/src/database/migrations/20220711025216-update-is-credential-used-function.js
```diff
@@ -0,0 +1,76 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, _) => {
+    await queryInterface.createFunction(
```
tan-zx, line 5: I presume creating a new function with the same name will override the previous function?

stanleynguyen: yup

backend/src/database/migrations/20210429045222-create-is-credential-used-function.js
stanleynguyen: > Oh I mean will there be instances where a credential is used for 3 email campaigns? Probably not if the code is working properly haha. Just thinking out loud.

Under the right condition (i.e. 2 checks are done at the same time with 1 existing job), yes. But this is super unlikely to happen

backend/src/database/migrations/20210429045222-create-is-credential-used-function.js
tan-zx: btw dk if you know this, but there are some db functions documented here, including the one we're modifying here: `worker/src/core/resources/sql/is-credential-used.sql`

I suspect it's dead code that we should remove?

backend/src/database/migrations/20220711025216-update-is-credential-used-function.js
tan-zx: > oh, might be a good idea to include a comment in the previous database migration to say that it's outdated.

haha think it's still good to this, just to avoid future confusion

PR #1669, opened 2022-07-06
build(scripts): add script for developing on local against staging db and workers

tan-zx:
added scripts to allow development locally against db and workers
- `dev:connectstagingdb` creates writable connection to db
- `dev:connectstagingdbreadonly`creates readonly connection to db → avoids accidental changes to staging db


tan-zx:
added in distinction between `readonly` and normal connection to avoid weird bug from Thursday 😅

stanleynguyen (approved):

PR #1670, opened 2022-07-06
feat(backend): stream report instead of querying everything at once

stanleynguyen:
## Problem

Downloading reports of too large campaigns will crash our server as we're querying for too much data to be processed

## Solution

Space out the query and stream results to response

## Deployment Checklist

N/A


tan-zx (approved):
tested, it works!

backend/src/sms/services/sms-stats.service.ts
```diff
@@ -35,13 +36,15 @@ const refreshStats = async (campaignId: number): Promise<void> => {
 
 /**
  * Gets delivered recipients for sms campaign
+ * Stream to deal with large campaign size
  * @param campaignId
```
tan-zx, line 40: Nit: to remove the JSDoc thing (or to elaborate on it o.o)

backend/src/telegram/services/telegram-stats.service.ts
```diff
@@ -40,10 +41,15 @@ const refreshStats = async (campaignId: number): Promise<void> => {
  * @param campaignId
```
tan-zx, line 41: Nit: to remove the JSDoc thing (or to elaborate on it o.o)

backend/src/email/services/email-stats.service.ts
```diff
@@ -38,13 +39,15 @@ const refreshStats = async (campaignId: number): Promise<void> => {
 
 /**
  * Gets delivered recipients for email campaign
+ * Stream to deal with large campaign size
```
tan-zx, line 42: Nit: to remove the JSDoc thing (or to elaborate on it o.o)

PR #1677, opened 2022-07-07
feat(email): add unsub header to emails

stanleynguyen:
## Problem

Our emails don't contain an unsubscribe header - an easy way to increase our credibility score with email clients

## Solution

Add `List-Unsubscribe` header to our mail payload if there's an `unsubLink` provided

## Deployment Checklist

N/A


tan-zx:
Looks good to me, just a couple of points:
- Was trying out on my local machine → I then realised this doesn't actually add the `unsubLink` to our emails. Have done so here in `worker/src/core/loaders/message-worker/email.class.ts`, pushed to staging (since I don't have local workers) and will test it out against staging later (will take some time for staging to be fully deployed)
image
- Do you want to use the `comment` field too?
image
- Haha I know this is a legacy thing but wonder if there's some way to get rid of all the `any`s (realised we do have `@types/nodemailer` installed in shared)

Otherwise looks good to me!

tan-zx:
Just tested: seems to work! 
image

stanleynguyen:
> Was trying out on my local machine → I then realised this doesn't actually add the unsubLink to our emails. Have done so here in worker/src/core/loaders/message-worker/email.class.ts (it should just be unsubLink instead of unsubLink: unsubLink)

@tan-zx could you elaborate on the above point? I couldn't get what's being said

tan-zx:
ohh, as in with this PR, the email sent by workers won't have the unsubscribe header → need to add `unsubLink` in line 145 of `worker/src/core/loaders/message-worker/email.class.ts`.
image

after adding that, I tested it on staging and [it worked](https://github.com/opengovsg/postmangovsg/pull/1677#issuecomment-1195020147)

that's all loool

stanleynguyen:
> ohh, as in with this PR, the email sent by workers won't have the unsubscribe header → need to add `unsubLink` in line 145 of `worker/src/core/loaders/message-worker/email.class.ts`. image
> 
> after adding that, I tested it on staging and [it worked](https://github.com/opengovsg/postmangovsg/pull/1677#issuecomment-1195020147)
> 
> that's all loool

Nice catch. Thanks @tan-zx 

Next time if the file is not available in the PR for commenting. Try using permalink like [this one](https://github.com/opengovsg/postmangovsg/blob/e986d2e8efb0f93de36be25b7096dc34c356d6a2/worker/src/core/loaders/message-worker/email.class.ts#L141-L148) to make the reference => less tedious and more accurate than taking ss

tan-zx (approved):
lgtm

PR #1679, opened 2022-07-08
hotfix(sms-callback-auth): catch errors from timingSafeEuqual and eval as not equal

stanleynguyen:
We expected the [crypto.timingSafeEqual(a,b)](https://nodejs.org/api/crypto.html#cryptotimingsafeequala-b) method to only return a boolean - however, we discovered post-hoc that it also throws an error if the two params are not the same length.
> An error is thrown if a and b have different byte lengths.

This hotfix accounts for that behavior by wrapping the method in a try/catch handler.


tan-zx (approved):
lgtm

tan-zx:
Turns out there's a good reason for throwing an error rather than returning false 🤔

From https://www.youtube.com/watch?v=RijGNytjbOI

image

PR #1688, opened 2022-07-12
fix: increased wait time for page to load

### LINKED ISSUE #1627: Cypress E2E test throws an error on twilio
**Describe the bug**
An error of [14101 - Invalid "To" attribute] is thrown on Twilio console 

**To Reproduce**
Run the cypress test

**Screenshots**

Screenshot 2022-06-22 at 3 25 41 PM

yingyee0111:
## Problem

Cypress test was failing on sms sometimes.
Cypress couldn't find the textbox because the page loads too slowly.
(This is an assumed cause as Cypress works fine locally. Would have to verify again after pushing to github actions.)
Cypress also couldn't find the element sometimes even when it is loaded. This error is not consistent.

## Solution

Increased wait time on querying for the element and changed the identifier to a more stable one.

## Problem 2
Twilio was throwing some error because it was auto replying for every message received.
Closes #1627

## Solution 2
Configure Twilio to do nothing when messages are received at our SG number.
(no code change, configuration is done on Twilio console)


cypress/e2e/sms.cy.js
```diff
@@ -67,6 +67,7 @@ describe('SMS Test', () => {
     cy.contains(':button', 'Create').click();
 
     //step 1 : enter message template
+    cy.get('textarea[id="message"]', { timeout: TIMEOUT });
```
stanleynguyen, line 70: Can we have a default value for TIMEOUT?

yingyee0111: oh like increased all timeout?

stanleynguyen (approved):

PR #1695, opened 2022-07-13
fix: sequelize issue

tan-zx:
First 3 commits are the commits from previous PRs that have been rolled back (have rewritten git history accordingly). These rolled-back PRs are:
- https://github.com/opengovsg/postmangovsg/pull/1684
- https://github.com/opengovsg/postmangovsg/pull/1683
- https://github.com/opengovsg/postmangovsg/pull/1685

The fourth commit (upgrading sequelize to v 6.21.2) contains the fix to the original error.

What went wrong:
- When upgrading to Sequelize v6.21.2, this includes v6.19 which included tighter TypeScript hinting for `WhereOptions` (see [here](https://github.com/sequelize/sequelize/releases/tag/v6.19.0) for more info)
- In order to fix the TypeScript errors, I had replaced `null`s with `undefined`s as the latter was how the data model was explicitly defined within Sequelize. This was **wrong**.
- In Sequelize, the [default behavior](https://sequelize.org/docs/v6/core-concepts/validations-and-constraints/#:~:text=Allowing/disallowing%20null,our%20code%20example%3A) is to allow `null` for every column of a model. (This is something of an open issue; see [here](https://github.com/sequelize/sequelize/issues/14260).) As such, the previous data models of `Campaign`, `SmsMessage` and `EmailMessage`, even without explicitly allowing `null`, actually allowed `null` values, thus giving rise to the TypeScript error.

Solution:
- As such, I had restored the original Sequelize methods to check for `null` instead of `undefined`.
- To resolve the TypeScript error, I extended the data models to explicitly include `null` type (as opposed to rely on Sequelize's default behavior). 

I have tested the code manually and [using Cypress](https://github.com/opengovsg/postmangovsg/actions/runs/2662987726), and it now works as expected.

To align TypeScript with Sequelize, there is a need to rule out `undefined`. 
- At first, we tried using `variable: string | null =  null` → but this causes unintended behavior.
- As such, we used the `!` operator to tell TypeScript that `variable` cannot be undefined. This works as intended. 

Have spoken to @lamkeewei and he says it's a good idea to go through the codebase and tighten the typing used accordingly. See https://github.com/opengovsg/postmangovsg/issues/1696


tan-zx:
Talked to @karrui and @dextertanyj; the use of `!` non-null assertion operator, while in line with existing practice in the codebase, might be less than ideal. 

Another possibility is to follow https://github.com/datagovsg/phas/pull/133/files and set ` "strictPropertyInitialization": false,` is `tsconfig.json`.

prestonlimlianjie:
> Talked to @karrui and @dextertanyj; the use of `!` non-null assertion operator, while in line with existing practice in the codebase, might be less than ideal.
> 
> Another possibility is to follow https://github.com/datagovsg/phas/pull/133/files and set ` "strictPropertyInitialization": false,` is `tsconfig.json`; however, for `postmangovsg`, this has been previously set to true.

We could tackle this in a separate PR

tan-zx:
Subsequent PR:
- Go through every column to explicitly set whether `allowNull` is true or false and amend the typing accordingly. (imo better to make it extra explicit rather than rely on Sequelize's default behavior)
- OK to turn this off `"strictPropertyInitialization": false,`

Maybe investigate:
- Frontend did not show any error, but backend threw an error (I think?)

prestonlimlianjie (approved):
lgtm

frontend/src/services/unsubscribe.service.ts
```diff
@@ -25,10 +23,15 @@ export async function unsubscribeRequest({
...
+  if (
+    axios.isAxiosError(e) &&
+    e.response &&
+    e.response.data &&
+    e.response.data.message
+  ) {
     throw new Error(e.response.data.message)
   }
-  throw new Error(defaultMsg || e.response?.statusText)
+  throw new Error(defaultMsg)
```
stanleynguyen, line 36: Could you elaborate a bit more on this change? I'm a bit concerned about the fact that `defaultMsg` is optional and this is considered a breaking change

tan-zx: I'll fix in a separate PR; found this one too.

Thanks for flagging!

image

PR #1699, opened 2022-07-15
fix: email verification can succeed if domain is verified

tan-zx:
## Problem

Context: we are trying to add an agency whose domain has been previously verified (but not the exact email address), so that they can send out using a custom domain using this email address

However, the current email verification process verifies the exact email with AWS SES, leading the verification process to fail. This is overly narrow — if the domain has been verified, the verification process should succeed.

## Solution

Modify the `verifyEmailWithAWS` method to check the domain of the email, and not just the email itself. 

## Deployment Checklist

N/A


prestonlimlianjie (approved):
lgtm

PR #1722, opened 2022-07-26
fix(backend): add missing awaits to floating promises

stanleynguyen:
Reopening #1211

## Problem

Sequelize `commit` and `rollback` [return Promises](https://sequelize.org/master/class/lib/transaction.js~Transaction.html#instance-method-commit) that need to be awaited on. Otherwise, this might create situations where the Promise returned by various async service have resolved but the underlying database transactions have not. 

In addition, there are also other instances in our code where we forgot to handle Promises returned by async function calls. The returned Promises could end up resolving unexpectedly, causing potential hard to spot errors.

## Solution

**Improvement**:
- Add [no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md) eslint rule to help highlight places where we did not explicitly handle Promises

**Bug Fixes**:
- Add missing `await` throughout codebase

## Tests
- Functionality remains the same. Check for any regressions.


stanleynguyen:
Superseded by #1729

PR #1726, opened 2022-07-27
fix: update swagger for transactional/email/send

yingyee0111:
## Problem

document api /transactional/email/send 

## Note

Not sure which error codes from cloudflare are relevant to our clients. Things like SSL errors are on us to fix but idk if we should let them know?


gweiying:
Would love to sync up with you on this before this is reviewed!

stanleynguyen (approved):

stanleynguyen:
> Would love to sync up with you on this before this is reviewed!

@gweiying let's update this doc again in the attachment PR

gweiying (approved):
lgtm

gweiying:
> > Would love to sync up with you on this before this is reviewed!
> 
> @gweiying let's update this doc again in the attachment PR

Can also merge into my PR if that makes more sense

PR #1727, opened 2022-07-27
feat: new api for filter sort

yingyee0111:
## Problem

campaign org

## Doc
https://www.notion.so/opengov/Campaign-Org-dc087cf2561341358bf5132da372ae3e#aacf60a5d3be4068a617886c6ba15fad


backend/src/core/middlewares/campaign.middleware.ts
```diff
@@ -103,13 +103,19 @@ const listCampaigns = async (
   res: Response,
   next: NextFunction
 ): Promise<Response | void> => {
-  const { offset, limit } = req.query
+  const { offset, limit, type, status, name, sort_by, order_by } = req.query
```
tan-zx, line 106: I wonder if it's possible to type this somewhere (since the shape described here is supposed to be consistent with that described in `getCampaigns` at `frontend/src/services/campaign.service.ts`)

that way you get type hinting if you modify one part and forgot to change the other part

(for context, Postman began as a JavaScript codebase and only added TypeScript afterwards, so don't think the typing practice here are ideal)

backend/src/core/services/campaign.service.ts
```diff
@@ -144,25 +150,95 @@ const listCampaigns = ({
   userId,
   offset,
   limit,
+  type,
```
tan-zx, line 153: same point re: the typing here https://github.com/opengovsg/postmangovsg/pull/1727/files#r944067275

probably should be in the shared folder imo

backend/src/core/routes/tests/campaign.routes.test.ts
```diff
@@ -74,6 +75,168 @@ describe('GET /campaigns', () => {
       ]),
     })
   })
+
+  test('List campaigns order by created at', async () => {
```
tan-zx, line 79: just for learning, I've seen `it` used instead of `test` and was wondering the difference and found this: https://stackoverflow.com/questions/45778192/what-is-the-difference-between-it-and-test-in-jest

TLDR: functionally identical, but use the one that's best for readability

backend/src/core/routes/tests/campaign.routes.test.ts
```diff
@@ -74,6 +75,168 @@ describe('GET /campaigns', () => {
...
+    for (let i = 1; i <= 3; i++) {
```
tan-zx, line 80: haha I think it's more idiomatic to use functional instead of imperative programming with TypeScript (i.e. to use `map` instead of for loops here)

(totally a nit / opinionated take, feel free to ignore)

backend/src/core/routes/tests/campaign.routes.test.ts
```diff
@@ -15,6 +15,7 @@ beforeAll(async () => {
 })
 
 afterEach(async () => {
+  await JobQueue.destroy({ where: {} })
   await Campaign.destroy({ where: {}, force: true })
```
tan-zx, line 19: for my own learning → why is `force: true` needed for `Campaign` and not `JobQueue`?

backend/src/core/routes/tests/campaign.routes.test.ts
```diff
@@ -74,6 +75,168 @@ describe('GET /campaigns', () => {
...
+      protect: false,
+    } as Campaign)
+    await JobQueue.create({
+      campaignId: campaign.id,
+      status: JobStatus.Logged,
+    } as JobQueue)
+
+    const resDraft = await request(app)
+      .get('/campaigns')
+      .query({ status: 'Draft' })
```
tan-zx, line 195: omw out now, but at first glance, `Draft` should be an enum instead of a string?

tan-zx: I think the test is relying on the implicit behavior that the default status is Draft. Would be good to note that in a comment for readability!

My own thinking is there's no need for this if it's an obvious property like `created_at` (I'd say even `sent_at` should be thus annotated), but good to have for something more business logic specific!

backend/src/core/services/campaign.service.ts
```diff
@@ -144,25 +150,95 @@ const listCampaigns = ({
...
+      }
+      default: {
+        break
+      }
+    }
+    const statusFilter = { '$job_queue.status$': operation }
+    whereFilter = { ...whereFilter, ...statusFilter }
+  }
+
+  let orderOpt: any = [['created_at', 'DESC']]
```
tan-zx, line 220: think we spoke about this earlier; wonder if we can give it a more specific type hm

frontend/src/classes/Campaign.ts
```diff
@@ -13,10 +13,25 @@ export enum Status {
   Halted = 'Halted',
 }
 
+export enum StatusFilter {
```
tan-zx, line 16: hmm I see you're extending the existing style, but these are the same enums repeated in `backend/src/core/constants.ts` and should really be extracted out to the shared folder too hmm (maybe we can leave a comment here for future refactoring, instead of dealing with them in the same PR)

backend/src/core/routes/tests/campaign.routes.test.ts
tan-zx: answer: `Campaign` has soft delete, whereas `JobQueue` does not

backend/src/core/routes/campaign.routes.ts
```diff
@@ -9,6 +9,11 @@ const listCampaignsValidator = {
   [Segments.QUERY]: Joi.object({
     limit: Joi.number().integer().min(1).max(100).default(10),
     offset: Joi.number().integer().min(0).default(0),
+    type: Joi.string().valid(...Object.values(ChannelType)),
```
tan-zx, line 12: NB called "campaign type"

tan-zx: on the frontend, this same field is called "Mode" 🙄️
 
image

backend/src/core/services/campaign.service.ts
```diff
@@ -144,25 +150,95 @@ const listCampaigns = ({
...
+      }
+      case Status.Sent: {
+        operation = { [Op.eq]: JobStatus.Logged }
+        break
+      }
+      default: {
+        break
+      }
+    }
+    const statusFilter = { '$job_queue.status$': operation }
```
tan-zx, line 216: add comment that this is a db join (since it could be expensive)

wonder if there's a better way to do it @stanleynguyen

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -84,15 +114,39 @@ const Campaigns = () => {
       .join(' ')
   }
 
-  async function fetchCampaigns(selectedPage: number) {
-    setLoading(true)
-    const { campaigns, totalCount } = await getCampaigns({
+  function CamelCaseWord(word: string): string {
```
tan-zx, line 117: camelCasea

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -395,22 +672,17 @@ const Campaigns = () => {
           Create new campaign
         </PrimaryButton>
       </TitleBar>
-      {campaignCount > 0 && (
-        <DemoBar
-          numDemosSms={numDemosSms}
-          numDemosTelegram={numDemosTelegram}
-          isDisplayed={isDemoDisplayed}
-        />
+      {campaignCount < 0 ? (
```
tan-zx, line 675: refactor

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -191,6 +253,59 @@ const Campaigns = () => {
...
+      sortBy,
+      orderBy,
+      statusFilter,
+      modeFilter,
+      newFilter
+    )
+  }
+
   /* eslint-disable react/display-name */
   const headers = [
```
tan-zx, line 310: rename to column?

backend/src/core/services/campaign.service.ts
```diff
@@ -144,25 +150,95 @@ const listCampaigns = ({
...
   offset,
   limit,
+  type,
+  status,
+  name,
+  sortBy,
+  orderBy,
 }: {
   userId: number
   offset: number
```
tan-zx, line 160: whether offset and limit should be optional arguments too

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -1,6 +1,6 @@
 import cx from 'classnames'
 
-import { capitalize } from 'lodash'
+import { capitalize, debounce } from 'lodash'
```
tan-zx, line 3: try to make this file smaller — GitHub is not displaying this file by default because it's too huge

frontend/src/components/dashboard/campaigns/Campaigns.tsx
```diff
@@ -210,7 +325,22 @@ const Campaigns = () => {
...
+            onSelect={handleModeFilter}
+            options={[filterInit, ...Object.values(ChannelType)].map((opt) => ({
+              label: CamelCaseWord(opt),
+              value: opt,
+            }))}
+            defaultLabel={
+              modeFilter === filterInit ? name : CamelCaseWord(modeFilter)
+            }
+            aria-label={name}
+          ></DropdownFilter>
```
tan-zx, line 341: `<DropdownFilter />`

backend/src/core/routes/campaign.routes.ts
```diff
@@ -63,6 +68,40 @@ const updateCampaignValidator = {
...
+ *          required: false
+ *          schema:
+ *            type: string
+ *        - in: query
+ *          name: sort_by
+ *          description: field used to sort campaigns
+ *          required: false
+ *          schema:
+ *            type: string
+ *            enum: [name, created_at, sent_at]
```
stanleynguyen, line 97: What if I want to sort by more than one field? Can we use a `+created_at` `-created_at` system instead of having 2 diff parameters `sort_by` and `order_by`? (i.e. combining the order information right inside the values for `sort_by`)

backend/src/core/services/campaign.service.ts
```diff
@@ -144,25 +150,95 @@ const listCampaigns = ({
...
+  } = { user_id: userId }
+
+  if (type) {
+    whereFilter.type = type
+  }
+
+  if (name) {
+    const nameMatch = {
+      name: {
+        [Op.iLike]: '%' + name + '%',
```
stanleynguyen, line 183: Just wanna remind here to double check for potential sql injections

backend/src/core/services/campaign.service.ts
stanleynguyen: Joins are not expensive, comparatively to the alternative which is making multiple queries , so this is ok

frontend/src/classes/Campaign.ts
yingyee0111: Future refactoring

backend/src/core/routes/campaign.routes.ts
yingyee0111: Standardizaton in separate PR

frontend/src/components/dashboard/campaigns/Campaigns.tsx
yingyee0111: refactor PR

tan-zx (approved):

backend/src/core/services/campaign.service.ts
tan-zx: Rae talked to Preston; apparently Sequelize parametrize the query, so there shouldn't be injection points

backend/src/core/routes/campaign.routes.ts
tan-zx: to deal with in future PR?

stanleynguyen: This is an irreversible decision, we can't go back and revise it later after people have started integration using this API interface. It's gotta be decided and resolved right in this PR

tan-zx: ohh I see your point — I don't think Rae will be working on this further (she's flying off tmr), think we should resolve this between us.

stanleynguyen: As per discussed, we probably don't ever need sort by multiple fields with campaigns so just go with this design

stanleynguyen (approved):

stanleynguyen (approved):

PR #1728, opened 2022-07-28
feat(report): increase retention of delivery reports to 30 days

stanleynguyen:
## Problem

Increase the expiry period for delivery report to 30 days and include that in the redaction reminder

- [Ticket](https://www.notion.so/opengov/Increase-expiry-period-of-both-reports-delivery-unsub-to-30-days-1a909a8153154f41ac5ac2a15c076953)
- [Ticket](https://www.notion.so/opengov/update-reminder-to-campaign-expiry-copy-65bb41839f0a49168afcb3db2616e802)

## Solution

Update the default config to 30 days

## Deployment Checklist

N/A

### DIFF
--- a/backend/src/core/config.ts
+++ b/backend/src/core/config.ts
@@ -459,7 +459,7 @@ const config = convict({
   redaction: {
     maxAge: {
       doc: 'Maximum age before campaign is redacted',
-      default: 14,
+      default: 30,
       env: 'REDACTION_MAXIMUM_AGE',
       format: Number,
     },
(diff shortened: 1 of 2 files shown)


tan-zx:
If I'm not mistaken, on the AWS S3 bucket, the .csv files uploaded by the users are deleted after every 14 days.

Depending on how the delivery reports are generated (haven't looked into this in detail), we might need to modify that setting to 30 days 🤔

stanleynguyen:
> If I'm not mistaken, on the AWS S3 bucket, the .csv files uploaded by the users are deleted after every 14 days.
> 
> Depending on how the delivery reports are generated (haven't looked into this in detail), we might need to modify that setting to 30 days 🤔

They're generated from our db records so no we don't need to modify that settings

PR #1729, opened 2022-07-28
fix: add no floating promises linting rule and awaits

tan-zx:
Does what https://github.com/opengovsg/postmangovsg/pull/1722 is meant to do

- Add new linting rules
- Fixed all the files flagged by the linting rules


tan-zx:
@stanleynguyen should we add this linting rule to the frontend too? (don't see why not)

when I use WebStorm, it actually gives prompts for floating promises, which I've seen on frontend code for other products

stanleynguyen:
> @stanleynguyen should we add this linting rule to the frontend too? (don't see why not)
> 
> when I use WebStorm, it actually gives prompts for floating promises, which I've seen on frontend code for other products

Yea we should

stanleynguyen (approved):
For the frontend portion, it's ok if u wanna do in another PR or right in this one

tan-zx:
Have done it within this PR; hope to get your review

Most of the changes are on the frontend. What I've done is:
- Add await if the function itself is alr `async`
- Use `void` if it is synchronous (so as to minimise changes to the callsites)

tan-zx:
also this is the PR with some changes made after your approval pending review @stanleynguyen (though I'm pretty sure the changes I made are fine haha)

stanleynguyen (approved):
lgtm

PR #1730, opened 2022-07-29
feat(unsubscriber): add reason to unsubscription request

stanleynguyen:
## Problem

Add reason to unsubscribe flow

- [Ticket](https://www.notion.so/opengov/Update-unsub-flow-copy-48f85ed8622142668697703f11a2241c)

## Solution

Update unsub flow to the new one

## Deployment Checklist

- [x] Migrate database to add the new columns of `unsubscribers` table

### DIFF
--- a/backend/src/core/services/unsubscriber.service.ts
+++ b/backend/src/core/services/unsubscriber.service.ts
@@ -36,19 +36,34 @@ const validateHash = ({
   }
 }
 
-/**
- * Create a new unsubscriber
- * @param campaignId
- * @param recipient
- */
 const findOrCreateUnsubscriber = ({
+  campaignId,
+  recipient,
+  reason,
+}: {
+  campaignId: number
+  recipient: string
+  reason: string
+}): Promise<[Unsubscriber, boolean]> => {
+  return Unsubscriber.findOrCreate({
+    defaults: {
+      reason,
+    } as Unsubscriber,
+    where: {
+      campaignId,
+      recipient,
+    },
+  })
+}
+
+const deleteUnsubscriber = async ({
   campaignId,
   recipient,
 }: {
   campaignId: number
   recipient: string
-}): Promise<[Unsubscriber, boolean]> => {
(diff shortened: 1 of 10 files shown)


tan-zx:
image
hmm testing on staging and got this

tan-zx:
Some thoughts on the design (thinking out loud here)
- Currently, if the user unsubscribes, changes his mind, and changes his mind again, they will get an "Invalid unsubscribe request"
  - Is this expected behavior or is this user flow sth we might want to support?
  - Why this might happen: user changes their mind and wishes to change the provided reason

image

Seems like there is safeguard against malicious unsubscribing by adding the hash in the submission body

backend/src/core/services/unsubscriber.service.ts
```diff
@@ -44,11 +44,31 @@ const validateHash = ({
 const findOrCreateUnsubscriber = ({
   campaignId,
   recipient,
+  reason,
```
tan-zx, line 47: need to add this to the JSDoc above (or just delete the useless JSDoc grrr)

tan-zx (approved):
left some comments on the design; otherwise gltm

frontend/src/components/unsubscribe/tests/Unsubscribe.test.tsx
```diff
@@ -62,14 +70,16 @@ test('successfully unsubscribes user from a campaign', async () => {
...
   server.use(...mockApis())
   renderUnsubscribe()
 
+  await unsubscribe()
+
   // Wait for the component to fully load
   const stayButton = await screen.findByRole('button', {
-    name: /stay/i,
+    name: /subscribe me back/i,
   })
```
tan-zx, line 83: Do we want to mock the error cases too? E.g. when the hash is wrong, or when the "Other" reason is too long (is there any limit on this?)

stanleynguyen: Nope no restriction on length of Other reason

PR #1733, opened 2022-08-01
fix: cypress to check report

yingyee0111:
## Problem 1

Cypress test was flaky, can't find DOM elements sometimes

## Solution 1

Allow cypress to retry the tests and only fail if all 3 attempts fail.
Note: assertions don't help as methods are alr retrying by default

## Problem 2
Cypress test does not check the report generated

## Solution 2
load the tracking pixel in gmail to mark email as read
download the report and check for READ status
for sms test, check for SUCCESS status 

## Note
This increases the time taken for cypress test to run as we need to wait for the report to be generated.

## Deployment Checklist

NIL

### DIFF
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -1,6 +1,7 @@
 const { defineConfig } = require('cypress');
 const path = require('path');
 const gmail = require('gmail-tester');
+const fs = require('fs')
 
 require('dotenv').config();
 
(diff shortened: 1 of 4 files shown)


stanleynguyen (approved):

yingyee0111:
error when run on github action https://github.com/cypress-io/cypress/discussions/22128

PR #1738, opened 2022-08-10
feat(unsubscribe): add unsub report download to campaign dashboard

stanleynguyen:
## Problem

Supersedes #1728 

- [Allow CSV download of unsubscribers](https://www.notion.so/opengov/allow-csv-download-of-unsubscribers-4f3349f8f3594faba8fc07659e6eacdf)
- [Unsub CSV download on campaign summary page](https://www.notion.so/opengov/allow-unsub-csv-report-download-on-campaign-summary-page-099c896eeb554076b58bbe0a73d418dd)

## Solution

- Add endpoint to retrieve unsubscribers of a campaign
- Add component on frontend to process and transform that into a csv file

## Deployment Checklist

N/A


backend/src/core/interfaces/campaign.interface.ts
```diff
@@ -123,4 +124,6 @@ export interface DuplicateCampaignDetails {
  *          updated_at:
  *            type: string
  *            format: date-time
+ *          error_code:
```
tan-zx, line 127: Couple things:
- `CampaignInterface` is dead code; created an issue here
- Noticed we're not including optional properties here; from [Swagger docs](https://swagger.io/docs/specification/data-models/data-types/#:~:text=By%20default%2C%20all%20object%20properties%20are%20optional.%20You%20can%20specify%20the%20required%20properties%20in%20the%20required%20list%3A): By default, all object properties are optional. You can specify the required properties in the required list:

Probably not much point to fixing, but just flagging up

tan-zx:
some frontend tests are failing

frontend/src/components/common/export-recipients/ExportRecipients.tsx
```diff
@@ -203,11 +201,9 @@ const ExportRecipients = ({
     <>
       {isButton ? (
         <div className={styles.actionButton}>
-          <InfoBlock className={styles.notice}>
```
tan-zx, line 206: is this design change intentional?

BEFORE:
image

AFTER:
image

tan-zx: Oh it is intentional nvm
image

frontend/src/components/common/progress-details/ProgressDetails.tsx
```diff
@@ -107,15 +109,34 @@ const ProgressDetails = ({
...
+    const list = await exportEmailUnsubscribers(id)
+
+    const headers = ['Recipient', 'Reason', 'Unsubscribed At']
+    const rows = list.map((r) => [r.recipient, r.reason, r.unsubscribedAt])
+    const content = [headers, ...rows].map(toCsvRow).join('\n')
+    const exportedAt = new Date()
+
+    download(
+      new Blob([content]),
+      `${name}_unsubscribers_${exportedAt.toLocaleDateString()}_${exportedAt.toLocaleTimeString()}.csv`,
```
tan-zx, line 132: maybe a bit extra, but might we want to convert the name to snake_case for consistency (think ppl won't usually name their campaigns that way)

e.g. of my file name `test unsub reason_unsubscribers_23_08_2022_19_17_55`

tan-zx: also prefer date format of YYYY_MM_DD etc. but that's just me

frontend/src/components/common/progress-details/ProgressDetails.tsx
```diff
@@ -161,11 +182,9 @@ const ProgressDetails = ({
           isButton
         />
       ) : (
-        <InfoBlock className={styles.notice}>
-          <Trans>
-            Delivery report has expired and is no longer available for download.
-          </Trans>
-        </InfoBlock>
+        <strong>
+          Delivery report has expired and is no longer available for download.\
```
tan-zx, line 186: typo here

image

also is the grayed out text by design (I actually think no need, since it's no longer a button)

frontend/src/components/common/progress-details/ProgressDetails.tsx
```diff
@@ -217,8 +236,47 @@ const ProgressDetails = ({
...
+                </td>
+              )}
+              <td className={'sm'}>{unsubscribed}</td>
+            </tr>
+          )}
         </tbody>
       </table>
+      <InfoBlock className={styles.notice}>
+        We recommend that you remove unsubscribers from your campaign to avoid
+        being marked as spam.{' '}
```
tan-zx, line 271: image

I would prefer stronger language here and maybe some bolding.

> You **should remove** unsubscribers from future campaigns to avoid having them marked as spam and impacting your agency's reputation.

tan-zx (commented):
Left some comments; mostly lgtm

tan-zx:
I also wonder whether this user experience could be improved. Since we are aligning the availability of the delivery reports and the unsubscribers list, for the user who _should be maintaining_ the cleanliness of their recipient list, it makes sense to include unsubscriber list in the rest of the delivery report. (Since unsubscribers are a subset of valid recipients; they should be removing invalid recipients too.)

That way, they would have a "single source of truth" with which to manage their recipient list. 

I'm OK with allowing them to download the unsubscriber list separately (since, compared to the other recipient status, this one might take a longer time to update). What I'm saying is the unsubscriber list should also be included in the delivery report. 🤔 

One possibility is users just just download the delivery report in order to show their bosses that the campaigns have been sent and are not aware of the unsubscribe feature at all; one way the delivery report could be designed is to add an "Unsubscribed" column, which would be NA for invalid recipients, and null for valid recipients and Yes for unsubscribed recipient.

I also wonder if the UI can be improved to shape intended user behavior. For example, currently there's the "refresh stats" button, but I'm not 100% sure how that works. Is the user expectation that the stats displayed on the frontend could be stale/outdated and they need to hit the refresh stats button to get the latest? Or is it normally already updated, but we give them a button to hit _in case_ it's not updated? 

Tagging @lisatjide and @khaleedah since this is more product/design-related.

frontend/src/components/common/progress-details/ProgressDetails.tsx
stanleynguyen: This is following the existing convention from our delivery report. It's been working so far (and has implied training effect on our users) so I don't think we need to stray from

frontend/src/components/common/progress-details/ProgressDetails.tsx
stanleynguyen: This is the design

tan-zx: haha I know can we change the design @khaleedah

frontend/src/components/common/progress-details/ProgressDetails.tsx
tan-zx: ohh ok makes sense

frontend/src/components/common/progress-details/ProgressDetails.tsx
khaleedah: Ok –  I understand the need for the phrasing to be stronger but I don't think it needs to sound unnecessarily authoritative and aggressive - could be more turn off to the agencies. 

Could we rephrase it as the one below (action word upfront, added more context to the consequences so that they would be intrinsically motivated)  

Screenshot 2022-08-31 at 11 14 37 AM

> **Remove unsubscribers from your recipient list**, to avoid future campaigns being marked as spam and affecting the reputation of your agency. Learn more

tan-zx: sounds good!

khaleedah:
> I also wonder whether this user experience could be improved. Since we are aligning the availability of the delivery reports and the unsubscribers list, for the user who _should be maintaining_ the cleanliness of their recipient list, it makes sense to include unsubscriber list in the rest of the delivery report. (Since unsubscribers are a subset of valid recipients; they should be removing invalid recipients too.)
> 
> That way, they would have a "single source of truth" with which to manage their recipient list.
> 
> I'm OK with allowing them to download the unsubscriber list separately (since, compared to the other recipient status, this one might take a longer time to update). What I'm saying is the unsubscriber list should also be included in the delivery report. 🤔
> 
> One possibility is users just just download the delivery report in order to show their bosses that the campaigns have been sent and are not aware of the unsubscribe feature at all; one way the delivery report could be designed is to add an "Unsubscribed" column, which would be NA for invalid recipients, and null for valid recipients and Yes for unsubscribed recipient.
> 
> I also wonder if the UI can be improved to shape intended user behavior. For example, currently there's the "refresh stats" button, but I'm not 100% sure how that works. Is the user expectation that the stats displayed on the frontend could be stale/outdated and they need to hit the refresh stats button to get the latest? Or is it normally already updated, but we give them a button to hit _in case_ it's not updated?
> 
> Tagging @lisatjide and @khaleedah since this is more product/design-related.



No issues with the actual including unsub list in the delivery report, in fact yeap I agree we should include it in. I think the bigger issue is that users could be unsubbing past the delivery report date which expires, but if we already have the email push then that somewhat helps.

We can revisit the refresh stats button in a separate discussion, my suspicion on why we had the button is that user click that after they hit 'retry' but also they can just refresh the browser..?

stanleynguyen:
Eh forgot to check @khaleedah , these above designs are only applicable for email campaigns, so for the others should we:
- hide the unsub statistics
- hide the unsub warning

khaleedah:
> Eh forgot to check @khaleedah , these above designs are only applicable for email campaigns, so for the others should we:
> 
> * hide the unsub statistics
> * hide the unsub warning

@stanleynguyen yes, hide for SMS / Tele, thank you!

Issue #1589, opened 2022-06-13
add interface to config schema


tan-zx:
for typescript

PR #1739, opened 2022-08-10
chore: add ConfigSchema interfaces to backend and worker 

tan-zx:
Closes https://github.com/opengovsg/postmangovsg/issues/1589

This helps to ensure that the config file is properly typed (don't know how necessary, but I've seen this in other "recent" products like sync and PHAS and annoyed me slightly that Postman did not have this).


stanleynguyen (approved):

backend/src/core/config.ts
```diff
@@ -15,6 +15,137 @@ const rdsCa = fs.readFileSync(path.join(__dirname, '../assets/db-ca.pem'))
...
+    secretManagerSalt: string
+  }
+  database: {
+    databaseUri: string
+    databaseReadReplicaUri: string
+    dialectOptions: {
+      ssl: {
+        require: boolean
+        rejectUnauthorized: {
+          valueOf: any
```
tan-zx, line 35: is this a problem? I just copied from the type inference but I hate `any` @stanleynguyen

stanleynguyen: Oh iinw `rejectUnauthorized` should just be a boolean flag (which we defaulted to `true`)

tan-zx: I tried that but somehow TypeScript doesn't agree hmm

nvm just merge first

PR #1742, opened 2022-08-11
feat(backend): use pg advisory lock for file upload processing

stanleynguyen:
## Problem

Our processing function somehow hog event loop resulting in the worker process not being able to send back heartbeat for the whole duration of the file upload processing. 

## Solution

We're solving it using an advisory lock on the transaction for bulk inserting messages. 
The reason we choose this lock over:
- in-mem lock: this won't work across multiple instances of the backend
- redis (or any external) lock: this won't be released for restarts upon server shutdowns/crashes, which is the whole point of using bee-queue with its external (redis) persistence

## Deployment Checklist

N/A


stanleynguyen:
Tested to work in staging, but we can still consider raising the `stallInterval` to `120s` as we practically can safely have 5-10 mins of reprocessing delay before customers complain

lamkeewei (approved):
lgtm! I assume you've tested on staging. 

One other non-obvious thing is that we currently have a [timeout on the `/status` endpoint](https://github.com/opengovsg/postmangovsg/blob/2c40f212a1491a8d5d04a6f34bd3849e0ca55583/backend/src/core/services/upload.service.ts#L163-L196). How this works is that after 10 mins, on a `/status` call, if more than 10 mins have passed, it will reset the upload. We might want to release any lock at this point too. Otherwise, subsequent uploads will still fail.

Can see this original [PR](https://github.com/opengovsg/postmangovsg/pull/388) for context also.

stanleynguyen:
> lgtm! I assume you've tested on staging.
> 
> One other non-obvious thing is that we currently have a [timeout on the `/status` endpoint](https://github.com/opengovsg/postmangovsg/blob/2c40f212a1491a8d5d04a6f34bd3849e0ca55583/backend/src/core/services/upload.service.ts#L163-L196). How this works is that after 10 mins, on a `/status` call, if more than 10 mins have passed, it will reset the upload. We might want to release any lock at this point too. Otherwise, subsequent uploads will still fail.
> 
> Can see this original [PR](https://github.com/opengovsg/postmangovsg/pull/388) for context also.

The reason we had this is `If processing job exceeds 10 mins, it will be considered as timed out. This is to prevent the campaign from being stuck as uneditable if server dies while processing.`. But right now we have bee-queue to restart the job automatically if server dies, we can safely remove this check.

PR #1743, opened 2022-08-15
feat(email-template): update footer for more prominent unsubscribe button

stanleynguyen:
## Problem

- [Ticket](https://www.notion.so/opengov/update-unsub-link-on-email-footer-f79d938905264820887898c0cf2f9c1e)

## Solution

Update it duh

## Deployment Checklist

N/A

### DIFF
--- a/shared/src/theme/email-theme.mustache
+++ b/shared/src/theme/email-theme.mustache
@@ -202,14 +202,17 @@
           not the intended recipient, please delete this email, refrain from disclosing the contents to any person, and notify the
           agency or organisation that sent this to you.
           <br />
-          <br />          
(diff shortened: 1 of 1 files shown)


tan-zx:
Observing this weird behavior; the test email has the updated footer:
image

Whereas the actual email does not:
image

Will look into this further

tan-zx:
hmm gonna push to staging to see if sth wrong with my local

tan-zx:
OK works on staging — probably sth wrong with the local dev setup (probably need to clean the previously built stuff?)

tan-zx (approved):

PR #1744, opened 2022-08-16
fix(email-template): add missing masthead lion symbol

stanleynguyen:
## Problem

- [Ticket](https://www.notion.so/opengov/Add-lionhead-on-email-template-0967382828b14ae7930b893520d7be0a)

## Solution

When I was working on it, I realised the symbol wasn't added previously probably because it's super hard to vertically align image with text in an email. I had to use table with some pixel counting but it iz gud now

Screenshot 2022-08-16 at 11 57 42 AM

## Deployment Checklist

N/A


tan-zx:
Nice job, tested and it works. 

Not related to your PR, but noticed that there's a hyperlink around `gov.sg` which, while it is a legit gov website, doesn't seem like intended behavior on our end 🤔.

(think Gmail auto-inserts the hyperlink, so prob nothing we can do about that)

tan-zx (approved):

stanleynguyen:
> Nice job, tested and it works.
> 
> Not related to your PR, but noticed that there's a hyperlink around `gov.sg` which, while it is a legit gov website, doesn't seem like intended behavior on our end 🤔.
> 
> (think Gmail auto-inserts the hyperlink, so prob nothing we can do about that)
> 
> for contrast, this is the preview we are showing (no hyperlink):
> 
> image

Yup, I noticed that problem too, but don't think we can do anything about it

tan-zx:
haha found this https://stackoverflow.com/questions/11988534/prevent-gmail-from-creating-links-for-urls-and-email-addresses but don't know if it's worth the effort

stanleynguyen:
> https://stackoverflow.com/questions/11988534/prevent-gmail-from-creating-links-for-urls-and-email-addresses

It has never been a problem so I'd rather not tryna solve it

PR #1748, opened 2022-08-22
build(actions): wait for EB deployment status

stanleynguyen:
## Problem

Right now our CD only triggers EB deployments but doesn't wait nor report deployment status

## Solution

Switch to an action that wait for the deployment status

## Deployment Checklist

N/A


tan-zx (approved):
lgtm! will try it out later

PR #1749, opened 2022-08-22
fix(security): resolve dependabot alerts 20220822

stanleynguyen:


serverless/database-backup/package.json
```diff
@@ -30,7 +30,7 @@
     "typescript": "4.0.2"
   },
   "dependencies": {
-    "@google-cloud/kms": "2.11.1",
+    "@google-cloud/kms": "3.0.1",
```
tan-zx, line 33: no breaking change? 

or is it that this is alr broken anyway and @prestonlimlianjie is alr working to move this to ECS?

tan-zx (approved):

serverless/database-backup/package.json
stanleynguyen: No breaking change on the function that we're using. I already checked

PR #1751, opened 2022-08-24
build(actions): automatically run cypress post-deployments

stanleynguyen:
## Problem

E2E test is not automatically run post-deployment so we had to rely on human (who are forgetful by nature)

## Solution

Link up deployments and run E2E after

## Deployment Checklist

N/A


tan-zx (approved):
lgtm

PR #1753, opened 2022-08-29
fix(unsub): fix db constraint error because of soft delete

stanleynguyen:
## Problem

Currently re-unsubscribe will result in unique constraint error in sql as `findAndCreate` will try to insert again.

## Solution

Switch to upsert

## Deployment Checklist

N/A

### DIFF
--- a/backend/src/core/models/unsubscriber.ts
+++ b/backend/src/core/models/unsubscriber.ts
@@ -38,4 +38,7 @@ export class Unsubscriber extends Model<Unsubscriber> {
     type: DataType.STRING,
   })
   reason?: string
+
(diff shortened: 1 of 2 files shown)


tan-zx (commented):
is this what caused the 500 error on Datadog today?

since there's a new column, we need a db migration right

stanleynguyen:
> is this what caused the 500 error on Datadog today?
> 
> since there's a new column, we need a db migration right

It's already migrated, it's a problem with sequelize not handling unique index for soft-deleted records

tan-zx (approved):
lgtm

PR #1758, opened 2022-09-07
chore: init antivirus scan

gweiying:
## Problem

Closes [insert issue #]

## Solution

## Deployment Checklist

- [ ] First task
  - [ ] Sub-task 1
  - [ ] Sub-task 2
- [ ] Second task


jhibberd-gov:
Deprioritising this work for now.

PR #1761, opened 2022-09-13
build: changes relating to package.json commands

tan-zx:
Three changes:
- Fix README error
- Add script that makes connecting to production db (to run migrations) easier
- Edit `backend/.env-example` file for additional clarity (and updated corresponding env vars in 1Password vault)


backend/.env-example
```diff
@@ -63,10 +69,14 @@ TELEGRAM_BOT_GUIDE_URL="https://go.gov.sg/postman-recipient-guide"
...
 
+# DOMAIN_WHITELIST=".gov.sg"
+# COOKIE_DOMAIN="localhost"
+
 # FRONTEND_URL="http://localhost:3000"
 # SENTRY_DSN="https://123.ingest.sentry.io/456"
 
-# set of twilio credentials used for testing locally only
+# Note 1: set of twilio credentials used for testing locally only
+# Note 2: demo Twilio credentials are fetched from AWS Secrets Manager; NOT from env vars
```
stanleynguyen, line 79: can we just remove these bunch altogether, they're not helpful at all in anyway

tan-zx: I suspect they were used to make test API calls to Twilio in other parts of the codebase 🤔

I'll remove it here but keep it in the 1Password Vault in case it becomes helpful in the future.

tan-zx:
updated and rebased @stanleynguyen

stanleynguyen (approved):
Note: We will remove the production-related stuffs once we have db migration through CI

PR #1765, opened 2022-09-14
refactor: fix issue with new migration ending up in build folder

tan-zx:
Tested commands; works as intended

Another benefit of this PR is, while testing migrations, there is no need to run `npm run build` in order to run the migration with the latest changes.

Actually the `npx sequelize-cli model:generate --name XXX` command is quite useless:
- it generates a `.js` file that is quite different from how the rest of the data models are written. 
- but at least it helps to create a reasonably named migration file (though still need to manually edit the migration file, which is more of a hassle compared to my experience with TypeORM)

One problem I encountered is that, sometimes when I run a migration, if the migration relies on a model that doesn't exist, I need to run `npm run build` before the migration can tun properly.

I feel like this PR is only putting band-aid — it works for now, but ideally the migration should be automatically generated based on the data model defined and should be written in TypeScript rather than JavaScript.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,8 @@ npm run db:migrate # run all pending migrations
 npm run db:seed # seed database with dummy data
 ```
 
+If you run into errors at the `db:migrate` step, this is likely because you have created a new model in TypeScript that has not been compiled to JavaScript. Run `npm run build` to fix this error.
+
 If you need to undo any database migrations:
 
 ```zsh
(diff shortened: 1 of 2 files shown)


tan-zx:
Not sure whether I should change the other paths referenced in `.sequelizerc` (think they're not as problematic) @stanleynguyen

prestonlimlianjie (approved):
lgtm!

Issue #1713, opened 2022-07-21
Clean up database functions


tan-zx:
Noticed earlier this week that `email_from_address` accepts emails of any case (unlike `users` table) → should probably align the two to prevent unforeseen problems. For consistency, we should add this same trigger to `email_from_address`

More generally, noticed that there are database functions on staging and production dbs that are not on local dbs. Would be good to align them, in case of unforeseen behaviors.

tan-zx:
Got back to this; cross-checked the database functions across local, staging, and production and noticed the following discrepancies:

Total number of database functions are different:
- Production: 30
- Staging: 31
- Local: 24

Specifically, the diffs are:

Missing on local:
- `add_user`
- `format_user_on_insert`
- `log_next_job`
- `retry_jobs_email`
- `retry_jobs_sms`
- `update_stats`

Extra on staging
- `set_job_to_sent`

My current thinking is to align everything to production, but will look into the specific functions to see what's up with the differences.

tan-zx:
Deleting the extra function on staging:

name: `set_job_to_sent`
code:
```
CREATE OR REPLACE FUNCTION public.set_job_to_sent(job_id integer)
 RETURNS void
 LANGUAGE plpgsql
AS $function$
BEGIN
-- Set status to sent so that the logger will finalize it
UPDATE job_queue SET status = 'SENT', updated_at = clock_timestamp() where id = job_id AND status = 'SENDING';
END $function$
```

tan-zx:
Diffs of the migrated functions (left is local) vs what was originally in production (should be all just formatting differences):

`add_user`
<img width="718" alt="image" src="https://user-images.githubusercontent.com/67887489/190152529-4c9ec895-a0fb-4d7b-8031-af1b7d1893fb.png">

`format_user_on_insert`
<img width="710" alt="image" src="https://user-images.githubusercontent.com/67887489/190152730-82ac7638-85fb-4b71-a113-2632e88dd01e.png">

`log_next_job`
![image](https://user-images.githubusercontent.com/67887489/190153032-c36cee82-fc75-475a-befc-5f59727091d4.png)

`retry_jobs_email`
<img width="708" alt="image" src="https://user-images.githubusercontent.com/67887489/190153395-05cbb54b-2f6c-43f4-adf8-b815a238ca0e.png">

`retry_jobs_sms`
[...]

PR #1766, opened 2022-09-14
feat: make database triggers explicit

tan-zx:
## Problem

Closes https://github.com/opengovsg/postmangovsg/issues/1713

## Solution

- Formally add migration for `format_user_on_insert` and `format_user_trigger`; these already exist on staging and production
- Add `format_email_from_address` and `format_email_from_address_trigger`; these do not currently exist and are nice-to-have
- Formally add `add_user` database function migration + comments on usage; these already exist on staging and production
- Remove all the extra db functions from staging and production that are not in use (see comment below)

## Deployment Checklist

- [x] To run against staging to make sure everything works as intended
- [ ] For staging and production, since they alr have `format_user_trigger`, manually add `20220914082001-add_format_user_on_insert.js` to the`sequelize_meta` table to avoid migration errors. (Not an issue for database functions like `add_user`, since those would be replaced by default.)
- [x] Tested on staging
- [ ] To do on production


tan-zx:
Diffs of the migrated functions (left is local) vs what was originally in production (should be all just formatting differences):

`add_user`
image

`format_user_on_insert`
image

`log_next_job`
image

`retry_jobs_email`
image

`retry_jobs_sms`
image

`update_stats`
image

tan-zx:
hmm did not work; found out that an extra step to create the trigger is needed.

SSH-ed into the staging db and found the current trigger. only 1 trigger exists. 
```
Triggers:
    format_user_trigger BEFORE INSERT OR UPDATE ON users FOR EACH ROW EXECUTE PROCEDURE
 format_user_on_insert()
```

image

tan-zx:
was tracing the code and it seems like, out of the functions that are missing on local:
- `add_user`: not found in codebase
- `format_user_on_insert`: used in database trigger
- `log_next_job`: not found in codebase (only specific variants like `log_next_job_[email|sms|telegram]` are found in `loaders/message-worker/index.ts`)
- `retry_jobs_email`: not found in codebase (only general `retry_jobs` is used in `job.service.ts`)
- `retry_jobs_sms`: not found in codebase (only general `retry_jobs` is used in `job.service.ts`)
- `update_stats`: not found in codebase (only specific `update_stats_[email|sms|telegram]` are found

should the right course of action be to **_delete_** all the functions (except for `format_user_on_insert`) from staging and production, instead of adding them to local via a database migration? @stanleynguyen

tan-zx:
Talked to Kee Wei:
- `add_users` was a utility method to add someone that was not of whitelisted email domain → we shall retain this in a database migration and annotate clearly that it is meant to be triggered manually (and hence not invoked directly in the codebase)
- previously, intern has made sure that all required database functions are already in database migrations

For the remaining functions, to remove on staging and test first, before doing the same on production.

prestonlimlianjie (approved):
lgtm

Issue #1696, opened 2022-07-13
Tighten TypeScript setting of data models 


tan-zx:
Flowing from https://github.com/opengovsg/postmangovsg/pull/1695#issue-1303088974, to tighten TypeScript settings of data models (Sequelize has much better TypeScript support compared to when postmangovsg was first written)

As discussed [here](https://github.com/opengovsg/postmangovsg/pull/1695#issuecomment-1184144050), can set set `"strictPropertyInitialization": false` in `tsconfig.json` too.

tan-zx:
TODO:
- Noticed that columns that should have unique constraint (e.g. `users.email`, `email_from_address.email`) do not have them → this is quite dangerous, especially in light of `add_user` function (see [this](https://github.com/opengovsg/postmangovsg/issues/1713#:~:text=add_user%20was%20a%20utility%20method%20to%20add%20someone%20that%20was%20not%20of%20whitelisted%20email%20domain%20%E2%86%92%20we%20shall%20retain%20this%20in%20a%20database%20migration%20and%20annotate%20clearly%20that%20it%20is%20meant%20to%20be%20triggered%20manually%20(and%20hence%20not%20invoked%20directly%20in%20the%20codebase)))
- Generally, would be good to go through models and tighten them up

jhibberd-gov:
Moved to Notion.

PR #1768, opened 2022-09-15
feat: save transactional emails in database

tan-zx:
# Problem

Ticket [here](https://www.notion.so/opengov/Save-transactional-emails-to-DB-cc93dbd88ea040048f855440baa52650) 

See RFC [here](https://www.notion.so/opengov/RFC-Sending-Verifiable-Messages-35976433961c481d97f15b7b2ee611e9) for in-depth discussion

Tested and it works on staging (caa 2022-09-30)

# Solution

- Create a new table to store transactional email messages (one table for one type of transactional messages)
  - To do that, I need to impose uniqueness constraint on `users.email`, in order to display users directly in the `email_messages_tx` table
  - Also deals with this [comment](https://github.com/opengovsg/postmangovsg/issues/1696#issuecomment-1248106100) 
- Support detailed error statuses (using the `error_code` column), so we can know precisely what goes wrong for users (collect as much info as possible)
  - Supported error statuses:
    - Invalid from address
    - Rate limit reached (not sure whether there might an issue where this leads to excess database calls? I don't think so, but just flagging up here)
    - Unsupported file type
    - Malicious file
    - Blacklisted recipient
  - Unsupported error statuses (if we need to add these, we can move db entry creation forward, but at this point I don't think it's necessary, since it's just users using our API wrongly):
    - Attachment too large
    - Too many attachments
    - Invalid parameters (caught by joi)

# Not in this PR

- Processing attachments metadata e.g. hash, filename, file size etc. (to do in subsequent PR)
- Updating statuses based on callbacks (to do in subsequent PR)

# Remaining To-Dos

- Intend to write tests after receiving approval on the general direction of my PR

# Deployment Notes
- [x] Need to run db migrations _before_ deploying backend
- [x] Need to grant permissions to `flamingo_iam` first (because new table is created). Specifically:
  - [x] SSH to jumphost, connect to RDS via `psql -h <XXX>.rds.amazonaws.com -p 5432 -U <user_name><db_name> -W`
  - [x] Run `GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO flamingo_iam;` (NB semi colon is important)
  - [x] Run `GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA public TO flamingo_iam;` (NB semi colon is important)


backend/src/core/middlewares/file-attachment.middleware.ts
```diff
@@ -37,14 +42,33 @@ function preprocessPotentialIncomingFile(
...
+// ) {
+//   if (req.files?.attachments) {
+//     const attachments = req.files.attachments as fileUpload.UploadedFile[]
+//     await FileAttachmentService.uploadAttachments(attachments)
+//   }
+//   next()
+// }
+
 export const FileAttachmentMiddleware = {
+  // not sure whether to amend to
```
tan-zx, line 70: delete stray comment

backend/src/core/middlewares/file-attachment.middleware.ts
```diff
@@ -1,6 +1,8 @@
 import { Request, Response, NextFunction } from 'express'
 import fileUpload from 'express-fileupload'
 import config from '@core/config'
+// import { EmailTransactionalService } from "@email/services";
```
tan-zx, line 4: changes to this file is strictly not necessary for this PR

backend/src/database/migrations/20220915123536-create-email-messages-tx.js
```diff
@@ -0,0 +1,87 @@
...
+        allowNull: false,
+        type: Sequelize.DataTypes.DATE,
+      },
+    })
+  },
+  down: async (queryInterface, _) => {
+    await Promise.all([
+      queryInterface.dropTable('email_messages_tx'),
+      queryInterface.sequelize.query(
+        'DROP TYPE "enum_email_messages_tx_status";'
```
tan-zx, line 83: needed for db migrations to be fully reversible (this suggests earlier migrations that create enums without this are not really reversible)

this is because if you drop the table, the enum still persists (weirdly)

backend/src/core/services/file-attachment.service.ts
```diff
@@ -35,19 +37,45 @@ const parseFiles = async (
 }
 
 const sanitizeFiles = async (
-  files: { data: Buffer; name: string }[]
+  files: { data: Buffer; name: string }[],
```
tan-zx, line 40: at some point, probably good to align the various variable names (`files` and `attachments`) and type (think there is a `UploadedFile` type we could use that's more informative)

backend/src/database/migrations/20220915123536-create-email-messages-tx.js
```diff
@@ -0,0 +1,87 @@
...
+        type: Sequelize.DataTypes.ENUM(
+          TransactionalEmailMessageStatus.Unsent,
+          TransactionalEmailMessageStatus.InvalidFromAddressError,
+          TransactionalEmailMessageStatus.RateLimitError,
+          TransactionalEmailMessageStatus.InvalidMessageError,
+          TransactionalEmailMessageStatus.UnsupportedFileTypeError,
+          TransactionalEmailMessageStatus.MaliciousFileError,
+          TransactionalEmailMessageStatus.BlacklistedRecipientError,
+          TransactionalEmailMessageStatus.Accepted,
+          TransactionalEmailMessageStatus.Sent
```
tan-zx, line 57: compared to previous db migrations, using TypeScript in a `.js` file here lowers the risk of errors

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -23,53 +26,120 @@ export const InitEmailTransactionalMiddleware = (
...
-      } = req.body
+        replyTo,
+      },
+      hasAttachment: !!attachments,
+      attachmentS3Object: null,
+      status: TransactionalEmailMessageStatus.Unsent,
+      sentAt: null,
+      errorCode: null,
+      // not sure why unknown is needed to silence TS (yet other parts of the code base can just use `as Model` directly hmm)
+    } as unknown as EmailMessageTx)
```
tan-zx, line 73: not sure why unknown is needed to silence TS (yet other parts of the code base can just use `as Model` directly hmm)

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -23,53 +26,120 @@ export const InitEmailTransactionalMiddleware = (
...
+    if (!emailMessageTx) {
+      // TODO: log error, etc.
+      res.status(500).json({
+        message: 'Unable to create entry in email_message_tx',
+      })
+      throw new Error('Unable to create entry in email_message_tx')
+    }
+    // attach to body for subsequent middlewares
```
tan-zx, line 81: modify comment to make clear:
1. save db calls (`userEmail`)
2. to tell whether the request is an API request (see `req.body.emailMessageTxId` used in `email.middleware.ts`

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -23,53 +26,120 @@ export const InitEmailTransactionalMiddleware = (
...
         message: 'Failed to send email',
-        action: ACTION,
-        error: err,
+        action,
+        error,
       })
 
       const BAD_REQUEST_ERRORS = [
-        TemplateError,
+        InvalidMessageError,
```
tan-zx, line 133: thought it's weird to use `TemplateError` in API call → have created a new error under `transactional.error.ts` instead

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -92,11 +162,22 @@ export const InitEmailTransactionalMiddleware = (
         message: 'Rate limited request to send transactional email',
         userId: req?.session?.user.id,
       })
+      // think we can use void here? not using return value, no need to wait
```
tan-zx, line 165: not v sure about this

backend/src/email/models/email-message-tx.ts
```diff
@@ -0,0 +1,112 @@
...
+  UnsupportedFileTypeError = 'UNSUPPORTED_FILE_TYPE_ERROR',
+  MaliciousFileError = 'MALICIOUS_FILE_ERROR',
+  BlacklistedRecipientError = 'BLACKLISTED_RECIPIENT_ERROR',
+  Accepted = 'ACCEPTED',
+  Sent = 'SENT',
+  // unsupported for now
+  // Bounced = 'BOUNCED',
+  // Delivered = 'DELIVERED',
+  // Opened = 'OPENED',
+  // Complaint = 'COMPLAINT',
```
tan-zx, line 37: to implement when we support callbacks

backend/src/email/models/email-message-tx.ts
```diff
@@ -0,0 +1,112 @@
...
+  filename?: string
+  temp_filename?: string
+  error?: string
+}
+
+export enum TransactionalEmailMessageStatus {
+  Unsent = 'UNSENT',
+  // AttachmentTooLargeError = 'ATTACHMENT_TOO_LARGE_ERROR', // unused
+  // TooManyAttachmentsError = 'TOO_MANY_ATTACHMENTS_ERROR', // unused
+  // InvalidParamsError = 'INVALID_PARAMS_ERROR', // unused; these are errors caught by Joi
```
tan-zx, line 24: listed status in chronological order of appearance

have erred on the side of having as many statuses as possible (realistically we probably don't need so many?)

backend/src/email/models/email-message-tx.ts
```diff
@@ -0,0 +1,112 @@
...
+  @Column({ type: DataType.JSON, allowNull: true })
+  attachmentS3Object!: AttachmentS3ObjectInterface | null
+  // @Column(DataType.STRING)
+  // messageId: string | null
+
+  // supported statuses:
+  // 1. null (upon first creation)
+  // 2. ERROR (if error out at any point along the way)
+  // 3. SENDING (if API is successfully called without error on backend)
+  // other statuses requires callback, which will be handled in a later PR
```
tan-zx, line 79: updated comment, to fix later

backend/src/email/models/email-message-tx.ts
```diff
@@ -0,0 +1,112 @@
...
+  @Column({
+    type: DataType.ENUM(...Object.values(MessageStatus)),
```
tan-zx, line 81: this should be `TransactionalEmailMessageStatus` — will fix later

backend/src/email/models/email-message-tx.ts
```diff
@@ -0,0 +1,112 @@
...
+    allowNull: false,
+  })
+  status!: TransactionalEmailMessageStatus
+
+  // will be null upon first creation
+  @Column({ type: DataType.DATE, allowNull: true })
+  sentAt!: Date | null
```
tan-zx, line 88: maybe should rename as `acceptedAt` to align with `TransactionalEmailMessageStatus`

at the same time, named it `sentAt` because that's what is used in `email_messages` table. I thought the distinction should be:
- Accepted: accepted by our system
- Sent: received a response from downstream system, say, AWS SES

Not sure if this is too fine a distinction

backend/src/email/routes/email-transactional.routes.ts
```diff
@@ -230,8 +230,13 @@ export const InitEmailTransactionalRoute = (
     FileAttachmentMiddleware.fileUploadHandler,
     FileAttachmentMiddleware.preprocessPotentialIncomingFile,
     celebrate(sendValidator),
+    // can move saveMessage to an earlier step if we want to capture errors that might arise earlier too
+    emailTransactionalMiddleware.saveMessage,
     emailMiddleware.isFromAddressAccepted,
     emailTransactionalMiddleware.rateLimit,
+    // TODO this later; ideally, would want to upload attachments to S3 directly via backend, rather than send pre-signed URL to user?
+    // also, uploading attachments to S3 should not stand in the way of sending the email and should be done asynchronously?
+    // FileAttachmentMiddleware.uploadAttachments,
```
tan-zx, line 239: see discussion in PR conversation

backend/src/email/models/email-message-tx.ts
tan-zx: I have been hand-writing my migrations — this is probably wrong? would love if someone could show me how to use Sequelize properly...

backend/src/database/migrations/20220915123536-create-email-messages-tx.js
tan-zx: have replaced with `...Object.values(TransactionalEmailMessageStatus)`, which is even pithier

tan-zx:
Tested on staging — does not work because `SequelizeDatabaseError: permission denied for table email_message_tx`; think it's probably because:
1. This is a newly created table on RDS
2. The user role assigned to the backend doesn't have access to this table?

prestonlimlianjie:
Let's do up a RFC/engineering design doc and book a 1-hour discussion with all engineers.

backend/src/core/middlewares/file-attachment.middleware.ts
```diff
@@ -11,6 +13,9 @@ const fileUploadHandler = fileUpload({
   },
   abortOnLimit: true,
   limitHandler: function (_: Request, res: Response) {
+    // not sure whether to amend to use dependency injection so as to
+    // 1. use logger to log these errors (currently these are unlogged?)
```
stanleynguyen, line 17: All request and responses are logged rn so it shouldn't be an issue. The only potential problem is that we're logging the whole request including the file field - rendering the log record too big and thus truncated => not valid JSON for DataDog to index => We might wanna look into excluding these file fields from the logs to prevent this

backend/src/core/middlewares/file-attachment.middleware.ts
```diff
@@ -37,14 +42,32 @@ function preprocessPotentialIncomingFile(
      * express-fileupload does not throw error if num files
      * exceeded, instead truncates array to specified num
      */
+    // same as comment above; unlogged + status
     if (req.body.attachments.length > FILE_ATTACHMENT_MAX_NUM) {
       res.status(413).json({ message: 'Number of attachments exceeds limit' })
     }
   }
   next()
```
stanleynguyen, line 50: @gweiying why are we `next`-ing here instead of returning early ah?

stanleynguyen: Just for my context if this was intentional

backend/src/core/middlewares/file-attachment.middleware.ts
```diff
@@ -37,14 +42,32 @@ function preprocessPotentialIncomingFile(
...
+//     const attachments = req.files.attachments as fileUpload.UploadedFile[]
+//     await FileAttachmentService.uploadAttachments(attachments)
+//   }
+//   next()
+// }
+
 export const FileAttachmentMiddleware = {
   fileUploadHandler,
   preprocessPotentialIncomingFile,
+  // uploadAttachments,
```
stanleynguyen, line 72: I dont think we need to save the actual file on to S3 but rather just keep the original file name in a column of our DB as an audit trail of the file being sent over.

backend/src/core/models/user/user.ts
```diff
@@ -32,6 +32,7 @@ export class User extends Model<User> {
       isEmail: true,
       isLowercase: true,
     },
+    unique: true,
```
stanleynguyen, line 35: We might not need this if we use `id` instead

backend/src/core/services/file-attachment.service.ts
stanleynguyen: We can possibly use `ReadableStream` type for these I think

backend/src/database/migrations/20220915024630-alter-user-email-unique.js
```diff
@@ -0,0 +1,15 @@
+'use strict'
```
stanleynguyen, line 1: we might not need this migration

backend/src/database/migrations/20220915123536-create-email-messages-tx.js
stanleynguyen: It's not weird, it's a feature of sql databases - enums, and these are in no way related to the table created. They might be used in table field type declaration, but there's no cascading effect of removing the table on the enum type.

stanleynguyen: On the same note, u should remove the table first before removing the type (rn u're trying to do them at the "same" time which might cause flaky behaviors), because the enum type can't be removed if it's still used in the table declaration.

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -23,53 +26,120 @@ export const InitEmailTransactionalMiddleware = (
...
         subject,
         body,
         from,
-        recipient,
-        reply_to: replyTo,
-        attachments,
-      } = req.body
+        replyTo,
+      },
+      hasAttachment: !!attachments,
```
stanleynguyen, line 67: why don't we save the list of file names instead of a flag?

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -23,53 +26,120 @@ export const InitEmailTransactionalMiddleware = (
...
+      sentAt: null,
+      errorCode: null,
+      // not sure why unknown is needed to silence TS (yet other parts of the code base can just use `as Model` directly hmm)
+    } as unknown as EmailMessageTx)
+    if (!emailMessageTx) {
+      // TODO: log error, etc.
+      res.status(500).json({
+        message: 'Unable to create entry in email_message_tx',
+      })
+      throw new Error('Unable to create entry in email_message_tx')
```
stanleynguyen, line 79: If u're throwing an error, it will be returned as a 500 error already, no need to `res.json`

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -23,53 +26,120 @@ export const InitEmailTransactionalMiddleware = (
...
+    }
+    // attach to body for subsequent middlewares
+    req.body.userEmail = userEmail
```
stanleynguyen, line 82: Why do we need this?

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -92,11 +162,22 @@ export const InitEmailTransactionalMiddleware = (
...
+      void EmailMessageTx.update(
+        {
+          status: TransactionalEmailMessageStatus.RateLimitError,
+          errorCode: '429',
+        },
+        {
+          where: { id: req.body.emailMessageTxId },
+        }
+      )
       res.sendStatus(429)
```
stanleynguyen, line 175: Can we update this to `.json({ message: 'Too many requests. Please try again later.' })`

backend/src/database/migrations/20220915123536-create-email-messages-tx.js
tan-zx: > On the same note, u should remove the table first before removing the type (rn u're trying to do them at the "same" time which might cause flaky behaviors), because the enum type can't be removed if it's still used in the table declaration. 

Ohh you're right

tan-zx: > It's not weird, it's a feature of sql databases - enums, and these are in no way related to the table created. They might be used in table field type declaration, but there's no cascading effect of removing the table on the enum type.

Yeah that's fair, I was referring to the fact that our earlier migrations actually don't take this into account (so the down migration doesn't really entirely reverse the up migration)

I was just thinking, is it safe to use the following in the migration (as opposed to enumerating all the enums)?
```
status: {
        type: Sequelize.DataTypes.ENUM(
          ...Object.values(TransactionalEmailMessageStatus)
        ),
        allowNull: false,
      },
```

My thinking is, the shape of `TransactionalEmailMessageStatus` is actually not fixed and if I create more enums in the future, they will actually be created on the first migration right? 🤔 But that makes the migration longer and uglier

backend/src/email/middlewares/email-transactional.middleware.ts
tan-zx: ok done

backend/src/email/models/email-message-tx.ts
```diff
@@ -0,0 +1,109 @@
...
+    type: DataType.BIGINT,
+    autoIncrement: true,
+    primaryKey: true,
+    allowNull: false,
+  })
+  id!: number
+
+  @ForeignKey(() => User)
+  @Column({ type: DataType.STRING, allowNull: false })
+  userEmail!: string
```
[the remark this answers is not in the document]
tan-zx: I was thinking:
- we should have the uniqueness requirement in `user.email` anyway
- by saving `userEmail` here, we can visually browse through the db using Tableplus to get a sense of who's sending what, rather than having to relate it back to the `users` table all the time.

Not sure if there are any downsides

backend/src/core/models/user/user.ts
tan-zx: isn't it good to have this additional constraint anyway, in case some part of the backend code is screwed up or someone manually modifies the db by mistake

backend/src/core/services/file-attachment.service.ts
tan-zx: TypeScript TypeScript TypeScript

backend/src/email/middlewares/email.middleware.ts
```diff
@@ -218,6 +219,18 @@ export const InitEmailMiddleware = (
         fromAddress,
         action: 'isFromAddressAccepted',
       })
+      // update status for transactional emails
+      if (req.body.emailMessageTxId) {
+        await EmailMessageTx.update(
```
stanleynguyen, line 224: I think we can void here also

backend/src/email/services/email-transactional.service.ts
```diff
@@ -22,32 +22,47 @@ async function sendMessage({
...
-      attachments
-    )
-  }
+  const sanitizedAttachments = attachments
+    ? await FileAttachmentService.sanitizeFiles(attachments, emailMessageTxId)
+    : undefined
 
   const blacklisted = await isBlacklisted(recipient)
   if (blacklisted) {
+    await EmailMessageTx.update(
```
stanleynguyen, line 57: Can void here also

backend/src/core/models/user/user.ts
stanleynguyen: haha yea still good to have

backend/src/email/middlewares/email-transactional.middleware.ts
tan-zx: Just saw this! Following today's discussion, thinking we should store this as a [JSON array element](https://www.postgresql.org/docs/11/functions-json.html#:~:text=Get-,JSON%20array%20element,-(indexed%20from%20zero) with the following shape:

```TS
type Attachments = AttachmentMetadata[]

interface AttachmentMetadata {
  filename: string
  author: string
  hash: string
  // not sure what other stuff to include; will play by ear
}
```

It's my first time using JSON array element, but I think should be fine?

backend/src/core/middlewares/file-attachment.middleware.ts
```diff
@@ -37,14 +42,32 @@ function preprocessPotentialIncomingFile(
      * express-fileupload does not throw error if num files
      * exceeded, instead truncates array to specified num
      */
+    // same as comment above; unlogged + status
     if (req.body.attachments.length > FILE_ATTACHMENT_MAX_NUM) {
       res.status(413).json({ message: 'Number of attachments exceeds limit' })
```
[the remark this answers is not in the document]
tan-zx: I suspect it's a bug right — have amended to return early

backend/src/core/middlewares/file-attachment.middleware.ts
tan-zx: ohh where is this logging being done? is this an issue to be tackled in a separate PR or part of this PR?

backend/src/core/middlewares/file-attachment.middleware.ts
tan-zx: resolved → as discussed, we'll save a hash of the file attachment as a reasonable compromise between accountability and privacy.

backend/src/email/middlewares/email-transactional.middleware.ts
stanleynguyen: Yup should be fine, they will be saved as a jsonb string. Checkout `params` from `email-message` model for similar implementation

stanleynguyen:
> Btw I'm not super familiar with the middleware pattern, but it seems to me there's no reason to use `Promise<Response | void>` as the return type for a middleware? Rather, we can always just use `Promise<void>`, call the `next()` directly, and return early if necessary?
> 
> image

Sorry I dont get this question. But (my guess) is that u're talking about 
```ts
return res.status.json
```
versus
```ts
res.status.json
return
```

Either way is fine

tan-zx:
> > Btw I'm not super familiar with the middleware pattern, but it seems to me there's no reason to use `Promise<Response | void>` as the return type for a middleware? Rather, we can always just use `Promise<void>`, call the `next()` directly, and return early if necessary?
> > image
> 
> Sorry I dont get this question. But (my guess) is that u're talking about
> 
> ```ts
> return res.status.json
> ```
> 
> versus
> 
> ```ts
> res.status.json
> return
> ```
> 
> Either way is fine

haha yes that's what I'm talking about. thanks!

backend/src/core/middlewares/file-attachment.middleware.ts
stanleynguyen: Separate PR ser

tan-zx: OK created [issue](https://github.com/opengovsg/postmangovsg/issues/1788) (and added to [Notion](https://www.notion.so/opengov/Exclude-unnecessary-fields-from-request-to-enable-Datadog-indexing-e1d44a0dd74c4a5388a80f8616765677))

I love creating chore tasks loool

backend/src/email/middlewares/email-transactional.middleware.ts
tan-zx: hmm I think it's a bit different, since it's a JSON array rather than a JSON? because there could be multiple attachments? 🤔

tan-zx: I'm thinking of sth like `{ type: Data.Type.Array(DataType.JSON) }` — I'll experiment and make sure it works before requesting for review :)

backend/src/database/migrations/20220915123536-create-email-messages-transactional.js
```diff
@@ -0,0 +1,79 @@
...
+      },
+      attachments_metadata: {
+        type: Sequelize.DataTypes.ARRAY(Sequelize.DataTypes.JSON),
+        allowNull: true,
+      },
+      status: {
+        type: Sequelize.DataTypes.ENUM(
+          TransactionalEmailMessageStatus.Unsent,
+          TransactionalEmailMessageStatus.Accepted,
+          TransactionalEmailMessageStatus.Sent
```
stanleynguyen, line 51: Can we have these bunch as well so we don't have to make another migration when we start handling the other statuses? As the
Screenshot 2022-10-02 at 12 26 50 PM
y're already confirmed from our RFC

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -23,53 +29,129 @@ export const InitEmailTransactionalMiddleware = (
...
+      void EmailMessageTransactional.update(
+        {
+          attachmentsMetadata: [],
+        },
+        {
+          where: { id: emailMessageTransactional.id },
+        }
+      )
+    }
+    // TODO: process attachments metadata
```
stanleynguyen, line 90: TODO

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -23,53 +29,129 @@ export const InitEmailTransactionalMiddleware = (
...
-        replyTo:
-          replyTo ?? (await authService.findUser(req.session?.user?.id))?.email,
+        replyTo: replyTo ?? userEmail,
         attachments,
+        emailMessageTransactionalId,
       })
+      await EmailMessageTransactional.update(
+        {
+          status: TransactionalEmailMessageStatus.Accepted,
+          sentAt: new Date(),
```
stanleynguyen, line 130: should be `acceptedAt` to not be confused with `Send` event from SES

stanleynguyen: TODO: fix for emails sent by workers as well

backend/src/email/models/email-message-transactional.ts
```diff
@@ -0,0 +1,97 @@
...
+  recipient: string
+
+  @Column({ type: DataType.JSON, allowNull: false })
+  params: Record<string, string>
+
+  @Column({ type: DataType.STRING, allowNull: true })
+  messageId: string | null
+
+  // null means this column has yet to be processed
+  // empty array means there are no attachments
```
stanleynguyen, line 62: This is unnecessary as it's very unlikely that this column will be null under this circumstance, and if it does - we have a bug to fix rather than an audit need. It's not worth it to require another db update as we have application logs for this kinda things

backend/src/database/migrations/20220915123536-create-email-messages-transactional.js
```diff
@@ -11,14 +11,13 @@ module.exports = {
         primaryKey: true,
         type: Sequelize.BIGINT,
       },
-      user_email: {
```
tan-zx: haha I put in `user_email` instead of `user_id` so that we can look at the table rows to see who's sending it immediately (without having to look up the `users` table), but it's probably no big deal either way!

stanleynguyen (approved):

PR #1774, opened 2022-09-21
feat: user-managed lists

stanleynguyen:
## Problem

Users might want to reuse the same csv for campaigns of different purposes but has no way of doing it directly on Postman right now.

## Solution

Implement managed CSVs 

## Deployment Checklist

- [ ] Migrate our database


prestonlimlianjie:
Closing this PR and reopening a new one so I can tag @stanleynguyen for review

PR #1775, opened 2022-09-21
feat: add UI with link to form for managed list

prestonlimlianjie:
This PR introduces a UI section to gather interest in our experimental managed list feature. The dropdown is currently disabled intentionally.

Screenshot 2022-09-21 at 7 14 10 PM

### DIFF
--- a/frontend/src/components/dashboard/create/email/EmailRecipients.tsx
+++ b/frontend/src/components/dashboard/create/email/EmailRecipients.tsx
@@ -29,6 +29,7 @@ import {
   WarningBlock,
 } from 'components/common'
 import useIsMounted from 'components/custom-hooks/use-is-mounted'
+import { ManagedListSection } from 'components/experimental'
 import { LINKS } from 'config'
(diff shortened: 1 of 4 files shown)


tan-zx:
LGTM 

Some comments/thoughts:
1. We're not committed to this particular design right? The goal is really to get agencies to express interest?
2. With the previous point in mind, I thought the currently UI doesn't make clear that the pre-existing "Upload recipient list in CSV format" and the new "Choose recipient list" are really in a "either-or" relationship? Have proposed a couple of alternative designs below

## Design 1: pre-face with "Options"
image

# Design 2: add link to form under CSV 
image

The NEW can be in a similar style as the DEMO below, which would certainly help in getting user's attention
image

tan-zx (approved):

prestonlimlianjie:
Thanks @tan-zx - these are great suggestions. Let's get this PR out first and monitor the number of form submissions; if the submission rate is 0/low, then let's bring these ideas to PM/design and run it by them

PR #1776, opened 2022-09-22
feat: whitelist for.sg as allowedImageSources

tan-zx:
## Problem

- `for.sg` file links are not showing up in preview

## Solution

- whitelist `file.for.sg` links 

Tested, works on staging


prestonlimlianjie (approved):
lgtm

PR #1777, opened 2022-09-26
feat(frontend): allow rich text in email protected content

stanleynguyen:
## Problem

Password-protected emails' protected content can only contain plain text atm

## Solution

Allow rich text 

## Deployment Checklist

- NA


prestonlimlianjie (approved):
took a quick glance, lgtm

Issue #759, opened 2020-10-01
[Custom Domain] Verification of custom from address across regions 


jeantanzj:
In #728 , Keewei pointed out that the workers and backend might use ses credentials in different regions. So we need to check that the custom from address is verified on both.

This problem can be resolved once we release #711 as workers and backend will share secrets.

lamkeewei:
If I remembered correctly, we previously set different SES credentials for OTPs (backend) and workers to ensure that users can still login even in the case our sending SES creds get blocked by high bounces. So just a note that this redundancy will be removed once we share SES creds on workers and backend.

jeantanzj:
Not going to remove the redudancy.  It's just that since they share secrets, backend can access worker's secrets and validate the custom from address using  workers' secrets

lamkeewei:
Re-opening this. After #910 is merged, we need to implement the DKIM checks for both AWS regions used in backend and worker's SES credentials. 

Currently, the checks are only for the regions used in backend's SES credentials.

tan-zx:
Picking up this issue:
- Right now the worker and backend call two different SES regions to send out emails but the env variables are set up such that each has access to the env variables for both regions
- Currently, only the backend verifies the custom domain with its own AWS SES region. This is not great because (1) the actual sending is done by the worker, which uses a different region; (2) while, currently, as a matter of practice, we add every custom domain to both regions, there is a small chance where we might screw up and add it to one and not the other.

As such, the solution is for the backend to validate across both regions.
- Currently, when workers send out emails, they do not verify the custom address with AWS SES. this is fine as long as the backend as already done the validation at an earlier step, which will surface any such error earlier
- We might switch between the two regions if our bounce rates get too high

To-Dos:
- [ ] Change config schema to make clear env variables for worker vs backend
- [ ] Use backend to validate email with both set of AWS regions
- [ ] Check to see if a similar verification process exists for API emails

PR #1778, opened 2022-09-26
feat: validate from address on both worker and backend SES regions

tan-zx:
## Problem

Closes https://github.com/opengovsg/postmangovsg/issues/759

See [here](https://github.com/opengovsg/postmangovsg/issues/759#issuecomment-1257414111) for context

### DIFF
--- a/backend/src/email/services/custom-domain.service.ts
+++ b/backend/src/email/services/custom-domain.service.ts
@@ -153,17 +153,10 @@ const storeFromAddress = async (
  * 3. Checks the domain's dns to ensure that the cnames are there
  */
 const verifyFromAddress = async (email: string): Promise<void> => {
-  const dkimTokensBackend = await verifyEmailWithAWS(
-    email,
-    backendSes,
-    backendSesRegion
-  )
-  const dkimTokensWorker = await verifyEmailWithAWS(
-    email,
-    workerSes,
-    workerSesRegion
-  )
-
+  const [dkimTokensBackend, dkimTokensWorker] = await Promise.all([
+    verifyEmailWithAWS(email, backendSes, backendSesRegion),
+    verifyEmailWithAWS(email, workerSes, workerSesRegion),
+  ])
   await verifyCnames([...dkimTokensBackend, ...dkimTokensWorker], email)
 }
 
--- a/backend/src/email/services/custom-domain.service.ts
+++ b/backend/src/email/services/custom-domain.service.ts
@@ -27,13 +27,15 @@ const verifyCnames = async (
   // get email domain
   const domain = email.slice(email.lastIndexOf('@') + 1)
   try {
-    for (const token of tokens) {
-      const cname = `${token}._domainkey.${domain}`
-      const [result] = await dns.promises.resolve(cname, 'CNAME')
-      if (result !== `${token}.dkim.amazonses.com`) {
-        throw new Error(`Dkim record doesn't match for ${email}`)
-      }
-    }
+    await Promise.all(
+      tokens.map(async (token) => {
+        const cname = `${token}._domainkey.${domain}`
+        const [result] = await dns.promises.resolve(cname, 'CNAME')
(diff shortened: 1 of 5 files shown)


tan-zx:
@stanleynguyen have made a related change to do the same checks for API emails (not sure if there is any reason against including these hmm might slow down speed of API calls? maybe don't need to check for every single call? wdyt hmm)

stanleynguyen:
> @stanleynguyen have made a related change to do the same checks for API emails (not sure if there is any reason against including these hmm might slow down speed of API calls? maybe don't need to check for every single call? wdyt hmm)

It's quite unnecessary to do that on every call because they will get an error immediately on every call anyway, please remove it. We should formulate the verification as a separate endpoint for the API users to check if they wanna

tan-zx:
> It's quite unnecessary to do that on every call because they will get an error immediately on every call anyway, please remove it. We should formulate the verification as a separate endpoint for the API users to check if they wanna

OK have removed. In the case where they get the error on each call, will the error message be informative though? 🤔

backend/src/email/services/custom-domain.service.ts
```diff
@@ -50,7 +52,11 @@ const verifyCnames = async (
  * @returns DKIM tokens generated for the email address
  * @throws Error if the domain's dns do not have the cname records
  */
-const verifyEmailWithAWS = async (email: string): Promise<Array<string>> => {
+const verifyEmailWithAWS = async (
+  email: string,
+  ses: AWS.SES,
```
stanleynguyen, line 57: Is this unused?

stanleynguyen: Oh moved from global to local

backend/src/email/services/custom-domain.service.ts
```diff
@@ -143,9 +153,18 @@ const storeFromAddress = async (
...
+  const dkimTokensBackend = await verifyEmailWithAWS(
+    email,
+    backendSes,
+    backendSesRegion
+  )
+  const dkimTokensWorker = await verifyEmailWithAWS(
+    email,
+    workerSes,
+    workerSesRegion
+  )
```
stanleynguyen, line 165: Rewrite this to 
```ts
await Promise.all([
  verifyEmailWithBackend,
  verifyEmailWithWorker,
])
```
to make more efficient use of the event loop cycle

tan-zx: Ohh good catch. Done.

stanleynguyen:
> > It's quite unnecessary to do that on every call because they will get an error immediately on every call anyway, please remove it. We should formulate the verification as a separate endpoint for the API users to check if they wanna
> 
> OK have removed. In the case where they get the error on each call, will the error message be informative though? 🤔

They will get an internal server error, but we have logs of what happened so should be fine. I don't think this will be a problem cos ppl will read our guide first before integrating

backend/src/email/services/custom-domain.service.ts
```diff
@@ -143,9 +153,11 @@ const storeFromAddress = async (
...
  */
 const verifyFromAddress = async (email: string): Promise<void> => {
-  const dkimTokens = await verifyEmailWithAWS(email)
-
-  await verifyCnames(dkimTokens, email)
+  const [dkimTokensBackend, dkimTokensWorker] = await Promise.all([
+    verifyEmailWithAWS(email, backendSes, backendSesRegion),
+    verifyEmailWithAWS(email, workerSes, workerSesRegion),
+  ])
+  await verifyCnames([...dkimTokensBackend, ...dkimTokensWorker], email)
```
stanleynguyen, line 160: We can refactor the `verifyCnames` function to use `Promise.all` as well. Do u wanna do it in this PR?

tan-zx: good catch, done (not 100% sure, but think it's right)

stanleynguyen (approved):

PR #1789, opened 2022-09-30
refactor: email messages data model

tan-zx:
Inspired by working on https://github.com/opengovsg/postmangovsg/pull/1768

Changes
- Tighten email messages data model (per https://github.com/opengovsg/postmangovsg/issues/1696) by making columns nullable rather than undefinable
- Set `strictPropertyInitialization` to false and removed unnecessary `!`


stanleynguyen:
> Unsure whether the following needs to be changed in the same manner:
> 
> ```ts
>   @BelongsTo(() => Unsubscriber)
>   unsubscriber?: Unsubscriber
> ```

No, as this is not a column

backend/src/core/models/user/user.ts
```diff
@@ -35,6 +35,8 @@ export class User extends Model<User> {
   })
   email!: string
 
+  // to be accurate, this should be api_key_ash
+  // should I rename this for clarity?
```
stanleynguyen, line 39: up to ya but u'll need a migration for this

tan-zx: OK think I can do it in a separate PR (with similar refactoring to get rid of all the `!` and stuff)

tan-zx:
PENDING: to test on staging first (will remove this comment when done)

stanleynguyen (approved):

Issue #1769, opened 2022-09-16
text on Swagger docs seem to be wrong?


tan-zx:
Hi @gweiying not 100% sure about this, but it seems to be that the text for authorisation on the Swagger docs seems to be wrong?
- the bearer auth is where the user should paste the user's API key
- the cookie auth is where the user should paste their cookie's value, i.e. not API key?

![SCR-20220916-tvg](https://user-images.githubusercontent.com/67887489/190677231-8f4d1698-dfcb-4674-80e2-bffda71ac6b5.png)

PR #1790, opened 2022-09-30
fix: session-based cookie auth is not supported on swagger docs

tan-zx:
## Problem

Closes https://github.com/opengovsg/postmangovsg/issues/1769

Currently, the "try it out" is misleading. 

When we do authentication via cookies, we do so by checking if the request is coming from an authenticated session , rather than looking at the cookie attached in the request. 

As such, have removed it.

Furthermore, have improved the documentation to make clear the format of the bearer token.


stanleynguyen:
This definition [is correct](https://swagger.io/docs/specification/authentication/cookie-authentication/) though, please dont remove it. Maybe add a description for this security schema to clarify instead

stanleynguyen (changes requested):
This definition [is correct](https://swagger.io/docs/specification/authentication/cookie-authentication/) though, please dont remove it. Maybe add a description for this security schema to clarify instead

backend/src/core/routes/index.ts
```diff
@@ -170,7 +170,9 @@ export const InitV1Route = (app: Application): Router => {
    *     bearerAuth:
    *       type: http
    *       scheme: bearer
+   *       bearerFormat: username_versionNumber_apiKey
    *     cookieAuth:
+   *       description: while this description (is correct)[https://swagger.io/docs/specification/authentication/cookie-authentication/], do note that cookie authentication is currently not supported for "try it out" requests due to browser security restrictions.
```
stanleynguyen, line 175: It is supported though, we're just using the wrong cookie name. Should be `postmangovsg-v2`
Screenshot 2022-10-02 at 12 17 55 PM

stanleynguyen: It was working when I "tried it out"

tan-zx: Ohh I copied that line from the page you linked and it didn't work when I tried. 

When I checked the cookie name from my browser, it didn't have the v2, but let me try again when I have the time!

tan-zx: where did you get the `COOKIE_NAME` in the screenshot from? mine is here:

image

and I tried it here (which didn't work):

image

tan-zx: oh I found it on production, that one has the `v2`

tan-zx: hmm but trying out production using that cookie still gives me a Internal Server Error 
image

tan-zx: from the website [you linked](https://swagger.io/docs/specification/authentication/cookie-authentication/) 
image

stanleynguyen: I guess we can remove it then. We're narrowing this swagger doc to just API later on anw so

tan-zx: by remove, you mean remove the `cookieAuth` right? will do

tan-zx: done

stanleynguyen (approved):

PR #1809, opened 2022-10-05
fix(backend): fix nonnullable from name error in txn email saving

stanleynguyen:
## Problem

https://sentry.io/organizations/open-government-products-re/issues/3643069534/?project=5272357&referrer=slack

## Solution

Make the field nullable as it should be

## Deployment Checklist

- [ ] Run db migration


stanleynguyen:
Bypassing and merging as a hotfix

PR #1815, opened 2022-10-05
feat(backend): handle status callbacks for transactional emails

stanleynguyen:
## Problem

Enable and handle callbacks for status update of transactional emails

## Solution

Implement transactional callbacks handling on the same endpoint as campaign-sent
- Add `mailConfigurationSet` as a backend env var. 
  - For context, we use different AWS SES regions for campaigns (prod: N. Virignia, staging: Frankfurt) and for transactional emails (prod: Sydney, staging: Frankfurt) but they are all named `postman-email-open`.
  - We are not distinguishing between the production and staging ones and are using a single env var on the backend (for transactional emails) and a single env var [on the worker](https://github.com/opengovsg/postmangovsg/blob/915f7152a58085d0ba82be61d250e0754897153d/worker/src/core/config.ts#L205) (for campaign emails) to refer to both config sets. This works because they're all named the same.
- Add extra SMTP header to indicate whether a given email is a transactional email one. This will be included in the callback AWS SES sends us, which we then parse to know to update `email_messages_transactional` table

Some notes:
- We decided to take the simple approach to email statuses by simply updating statuses based on what the AWS SES event is. 
  - We considered a more complex approach that would impose a certain order (e.g. preventing updating an email with a `Delivery` status to `Sent`, or an email with `Bounced` to `Delivery`), but decided against it. 
  - This is because AWS SES events typically arrive in the right order and, if we discover this to be an issue in the future, we can fix it then.
- We considered renaming `postman-email-open` to a more accurate name. 
  - For context, it was originally named `open` because the other events (`Sent`, `Bounced`, `Delivery`) were handled at the config set of the specific *verified identity* rather than at the AWS SES level. 
  - `Open`, however, could only be handled at the AWS SES level (because it required embedding a pixel to tell whether an email has been opened), which was why the config set was created and named thus. We decided to move all the config set to the AWS SES level so that there is a single source of truth. 
  - However, renaming the config set would be very troublesome if we wanted to ensure there is no downtime. Therefore we decided against it. 
- We considered whether to log the update status the way it is done in the [callbacks of campaign emails](https://github.com/opengovsg/postmangovsg/blob/31f023304ced17d0f79c178b7e15a8b519e68ee8/backend/src/email/utils/callback/query.ts), but we ultimately decided against as we didn't think it was useful for debugging. 

## Deployment Checklist

- [x] Migrate callbacks to use config set only
- [x] Migrate db

### DIFF
--- a/backend/src/core/config.ts
+++ b/backend/src/core/config.ts
@@ -81,6 +81,7 @@ interface ConfigSchema {
     workerHost: string
   }
   mailFrom: string
+  mailConfigurationSet: string
   mailVia: string
   mailDefaultRate: number
   transactionalEmail: {
@@ -432,6 +433,11 @@ const config: Config<ConfigSchema> = convict({
     doc: 'The email address that appears in the From field of an email',
     default: '',
     env: 'BACKEND_SES_FROM',
+  },
+  mailConfigurationSet: {
+    doc: 'The configuration set specified when sending an email',
+    default: 'postman-email-open',
+    env: 'BACKEND_SES_CONFIGURATION_SET',
   },
   mailVia: {
     doc: 'Text to appended to custom sender name',
--- a/backend/src/email/services/email-transactional.service.ts
+++ b/backend/src/email/services/email-transactional.service.ts
@@ -8,6 +8,7 @@ import {
   EmailMessageTransactional,
   TransactionalEmailMessageStatus,
 } from '@email/models'
+import { SesEventType } from '@email/interfaces/callback.interface'
 
 const logger = loggerWithLabel(module)
 
@@ -103,12 +104,12 @@ type CallbackMetaData = {
   }
 }
 async function handleStatusCallbacks(
-  type: string,
+  type: SesEventType,
   id: string,
   metadata: CallbackMetaData
 ): Promise<void> {
   switch (type) {
-    case 'Delivery':
+    case SesEventType.Delivery:
       await EmailMessageTransactional.update(
         {
           status: TransactionalEmailMessageStatus.Delivered,
@@ -119,7 +120,7 @@ async function handleStatusCallbacks(
         }
       )
       break
-    case 'Bounce':
+    case SesEventType.Bounce:
       await EmailMessageTransactional.update(
         {
           status: TransactionalEmailMessageStatus.Bounced,
@@ -134,7 +135,7 @@ async function handleStatusCallbacks(
         }
       )
       break
-    case 'Complaint':
+    case SesEventType.Complaint:
       await EmailMessageTransactional.update(
         {
           status: TransactionalEmailMessageStatus.Complaint,
@@ -146,7 +147,7 @@ async function handleStatusCallbacks(
         }
       )
       break
-    case 'Open':
+    case SesEventType.Open:
       await EmailMessageTransactional.update(
         {
           status: TransactionalEmailMessageStatus.Opened,
@@ -157,7 +158,7 @@ async function handleStatusCallbacks(
         }
       )
       break
-    case 'Send':
+    case SesEventType.Send:
       await EmailMessageTransactional.update(
         {
           status: TransactionalEmailMessageStatus.Sent,
--- a/backend/src/email/services/email-transactional.service.ts
+++ b/backend/src/email/services/email-transactional.service.ts
@@ -4,7 +4,10 @@ import { loggerWithLabel } from '@core/logger'
 import { isBlacklisted } from '@email/utils/query'
 import { InvalidMessageError, InvalidRecipientError } from '@core/errors'
 import { FileAttachmentService } from '@core/services'
-import { EmailMessageTransactional } from '@email/models'
+import {
+  EmailMessageTransactional,
+  TransactionalEmailMessageStatus,
+} from '@email/models'
 
 const logger = loggerWithLabel(module)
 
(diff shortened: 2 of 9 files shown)


stanleynguyen:
Pending answer from AWS for `Send` status callbacks

stanleynguyen:
Tested in staging

stanleynguyen:
@tan-zx plis review

backend/src/email/services/email-transactional.service.ts
```diff
@@ -74,17 +77,108 @@ async function sendMessage({
...
+    complaintFeedbackType: string
+    complaintSubType: string
+  }
+}
+async function handleStatusCallbacks(
+  type: string,
+  id: string,
+  metadata: CallbackMetaData
+): Promise<void> {
+  switch (type) {
```
tan-zx, line 110: ohh just found `backend/src/email/utils/callback/update-status.ts` and `backend/src/email/utils/callback/query.ts` — wonder if we can re-use the code there

it seems like what they've done there is to run a logger.info before actually making any db changes:
```
const { errorCode, errorSubType, timestamp, id } = opts

  logger.info({
    message: 'Updating email_messages table',
    opts,
    action: 'updateMessageWithError',
  })
```

maybe we can follow that pattern here? (in case sth goes wrong, so we can have logs to trace)

tan-zx: separately, maybe we can just use `void` instead of `await`, since we're not actually waiting for the function to return anything

shared/src/clients/mail-client.class/index.ts
```diff
@@ -8,6 +8,10 @@ const CONFIGURATION_SET_HEADER = 'X-SES-CONFIGURATION-SET' // Case sensitive
 
 export * from './interfaces'
 
+export type SendEmailOpts = {
+  extraSmtpHeaders: Record<string, any>
```
tan-zx, line 12: I don't really understand what's happening with this SMTP headers thing!

backend/src/email/utils/callback/parsers/ses.ts
```diff
@@ -12,6 +12,7 @@ import {
 import { addToBlacklist } from '@email/utils/callback/query'
 import config from '@core/config'
 import { compareSha256Hash } from '@shared/utils/crypto'
+import { EmailTransactionalService } from '@email/services/email-transactional.service'
```
tan-zx, line 15: truth be told, don't really understand what's going on in this file... seems to me that it's used to parse the callbacks from SES?

backend/src/email/utils/callback/parsers/ses.ts
```diff
@@ -222,6 +224,7 @@ const parseRecord = async (record: SesRecord): Promise<void> => {
   await blacklistIfNeeded(message)
 
   const id = smtpApiHeader?.unique_args?.message_id
+  const isTransactional = smtpApiHeader?.isTransactional
```
tan-zx, line 227:
```suggestion
  const isTransactional = !!smtpApiHeader?.isTransactional
```

tan-zx: just a nit, so it won't be `undefined` (type-safetyytyyzz)

tan-zx (changes requested):
looks good! thanks for sorting out the callback / config set stuff on AWS SES!

backend/src/email/services/email-transactional.service.ts
```diff
@@ -74,17 +77,108 @@ async function sendMessage({
...
+        {
+          status: TransactionalEmailMessageStatus.Sent,
+          sentAt: metadata.timestamp,
+        },
+        {
+          where: { id },
+        }
+      )
+      break
+    default:
```
tan-zx, line 171: Just thinking out loud (and aware that this might lead to over-engineering), but should we enforce some kind of order in the update operations?

For example, our code should disable updating a row whose status was previously set to `Opened` to `Sent`. (Can think of more examples.) This will prevent callbacks that come back in the wrong order (unlikely but possible?) from messing up our records?

backend/src/core/config.ts
```diff
@@ -433,6 +434,11 @@ const config: Config<ConfigSchema> = convict({
     default: '',
     env: 'BACKEND_SES_FROM',
   },
+  mailConfigurationSet: {
+    doc: 'The configuration set specified when sending an email',
+    default: 'postman-email-open',
```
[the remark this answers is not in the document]
stanleynguyen: re: Rename => ok sure, but we will have to do it like how we rename db column, mafan asf
re: different config set => nope, we're handling them using the same endpoint and will eventually move both to the same standards, why would I want diff config sets

backend/src/email/services/email-transactional.service.ts
stanleynguyen: > 

Nope, this was intentionally written without reusing that as we're having 2 separate standards of dealing with callbacks at the time of writing

shared/src/clients/mail-client.class/index.ts
stanleynguyen: Can elaborate more?

backend/src/email/utils/callback/parsers/ses.ts
stanleynguyen: undefined is evaluated to falsy in js, it's common to not do `!!`

backend/src/email/services/email-transactional.service.ts
stanleynguyen: I thought about that:
- First thing is, there's no inherent order to these apart from just `Send` happens before the rest
- To implement this, it's actually not to determine an order of the statuses but rather just taking whichever has the latest timestamp. However, right now I see this problem to be quite minor, most of the time SES sends these events in the order of their timestamp (earliest to latest). We can come back to this if it becomes a problem but for now I don't think it's worth the effort

backend/src/core/config.ts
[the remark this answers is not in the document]
tan-zx: > re: Rename => ok sure, but we will have to do it like how we rename db column, mafan asf

oh no, let's just write it in comments instead then 

> re: different config set => nope, we're handling them using the same endpoint and will eventually move both to the same standards, why would I want diff config sets

Ooh so does this mean that we have to make sure that the various config sets across all the regions we use AWS SES for are named `postman-email-open`? (Not sure if I'm misunderstanding how the config set thing works — I though this is something that's manually configured on AWS SES) 

My understanding is these three config sets on the various regions are distinct and not the same?

image
image
image

backend/src/email/services/email-transactional.service.ts
tan-zx: OK with keeping the code separate — any reason why we're not logging the update though?

shared/src/clients/mail-client.class/index.ts
tan-zx: Oh I just don't really understand this part of the PR haha — maybe can find some time to talk through for my understanding? 🙏

backend/src/email/services/email-transactional.service.ts
tan-zx: The other ones I can think of is `Complaint` and `Bounce` should happen after `Delivery` (if at all), i.e. we wouldn't want `Delivery` to overwrite the previous two. It does seem like quite a lot of additional complexity to enforce this ordering though.

Will defer to you! (For stuff like this, do we add in the sprint tasks so that we can keep track of this and come back to in the future? Should these discussions be centralised somewhere so it's easier to track in the future? Feels like PR comments are not gr8 documentation)

backend/src/email/services/email-transactional.service.ts
stanleynguyen: > OK with keeping the code separate — any reason why we're not logging the update though?

Cos we're saving it in database? Would be a waste of computing resource and storage

backend/src/email/services/email-transactional.service.ts
stanleynguyen: > The other ones I can think of is `Complaint` and `Bounce` should happen after `Delivery` (if at all), i.e. we wouldn't want `Delivery` to overwrite the previous two. It does seem like quite a lot of additional complexity to enforce this ordering though.
> 
> Will defer to you! (For stuff like this, do we add in the sprint tasks so that we can keep track of this and come back to in the future? Should these discussions be centralised somewhere so it's easier to track in the future? Feels like PR comments are not gr8 documentation)

Like I mentioned in the above comment, there's no inherent order to some of the events, if we're to implement we will need to do it based on the timestamps. I honestly don't think we need to tackle this, it's `medium effort, low-to-zero impact`

backend/src/core/config.ts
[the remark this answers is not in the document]
stanleynguyen: They are not the same, just happened to all be named the same (i.e. the default name)

shared/src/clients/mail-client.class/index.ts
stanleynguyen: Yea can u elaborate what it is that u dont understand?

stanleynguyen:
All resolved, could u help review again @tan-zx

tan-zx (approved):
LGTM. I also updated PR description based on our discussion. Think it'll be helpful for future documentation too!

Haha if you don't mind, could write sth like this in the future, so it'll be easier to review in the future! You'll probably (no guarantee) get less stupid questions from me too loool

tan-zx:
btw just to clarify, for the deployment, we should make these changes first (config set + db migration) before merging in and deploying the backend right? 

do you mind pairing on the making changes to the config set on prod? thanks!

PR #1816, opened 2022-10-05
feat: managed lists

prestonlimlianjie:
## Problem

Users might want to reuse the same csv for campaigns of different purposes but has no way of doing it directly on Postman right now.

### Implement managed CSVs 
- Users can set their newly uploaded CSV file as a managed list to be used for later via a toggle in the UI
- The setting for whether a newly uploaded CSV file will become a managed list will only be saved when the user actually sends the campaign
- Users can see a dropdown of previously uploaded managed lists and select one to be used for a new campaign (up to an expiry date, it is currently 14 days, in line with our S3 file expiry deadline)
- Managed lists are specific to campaign types
- Managed lists are currently only released for the SMS and email campaign types

### Creating a Lists table
Each managed list is stored as a record in a new Lists table. The List record contains similar information as the `s3_object` field in the campaigns table. This is because a managed list is actually a pre-existing CSV file in the S3 bucket that was previously uploaded by the user.

### Creating a UserLists table
The relationship between managed lists and users is a many-to-many one. That is, each user can have many managed lists, and each managed list can belong to many users. As such, we need to create a UserLists join table.

### Saving list mechanism
- This PR introduces a new `should_save_list` column in the Campaigns table
- We only update this column when the campaign has been sent (because we don't want unsent campaign CSVs to pollute the managed lists)
- At this same point, we create a new record in the Lists table, as well as a corresponding record in the UserLists table.

### Reusing the exact same CSV file in S3
As this feature is experimental, I did not want to introduce new pieces of infrastructure unnecessarily. The trade off that we get here is that each managed list file can only live for 14 days (before S3 deletes it, and we also set it to expire in our database by specifying the expiry timeline in our DB queries). The benefit of this trade off is that we could get this rather significant new feature of allowing users to set their own managed lists out within 1 day.

## Deployment Checklist

- [ ] Run migration - there are 3 migration files


backend/src/core/services/list.service.ts
```diff
@@ -0,0 +1,121 @@
...
+    const user = await User.findOne({
+      where: { id: userId },
+      include: [
+        {
+          model: List,
+          where: {
+            channel,
+            createdAt: {
+              // We currently only want to retrieve managed lists that were uploaded fewer than <DAYS_TO_LIST_EXPIRY> ago
+              [Op.gte]: currentDateTime.setDate(
```
richard-yak, line 30: wow nice, TIL you could offset time like this directly in TS!

backend/src/core/services/list.service.ts
```diff
@@ -0,0 +1,121 @@
...
+}): Promise<List> => {
+  return List.create({
+    s3key,
+    etag,
+    filename,
+    channel,
+  })
+}
+
+const addUserToList = async ({
```
stanleynguyen, line 103: can we rename this `grantListAccessToUser`

backend/src/email/middlewares/email-template.middleware.ts
```diff
@@ -174,6 +177,76 @@ export const InitEmailTemplateMiddleware = (
...
+      // Enqueue upload job to be processed
+      await EmailTemplateService.enqueueUpload({
+        campaignId: +campaignId,
+        template,
+        s3Key,
+        etag,
+        filename,
+      })
+
+      // Return early because bulk insert is slow
```
stanleynguyen, line 229: I think this comment is a bit misleading cos our processing happened elsewhere (as suggested from the `enqueueUpload` function call while this one might imply that it happens in this same place

backend/src/core/interfaces/campaign.interface.ts
```diff
@@ -21,9 +21,18 @@ export interface CampaignDetails {
   valid: boolean
   has_credential: boolean
   csv_filename: string
+  s3_object: {
```
stanleynguyen, line 24: can use `CampaignS3ObjectInterface`

stanleynguyen (approved):

backend/src/email/middlewares/email-template.middleware.ts
prestonlimlianjie: Resolved [9dad668](https://github.com/opengovsg/postmangovsg/pull/1816/commits/9dad668792a789f19da3dcf90e36e61e8644c843)

backend/src/core/services/list.service.ts
prestonlimlianjie: Resolved [13c39c2](https://github.com/opengovsg/postmangovsg/pull/1816/commits/13c39c2019568b6736dcbab909fcf9dc6f1a5460)

PR #1830, opened 2022-10-06
feat(backend): transactional email GET endpoints design

stanleynguyen:
## Problem

GET endpoints for:
- listing transactional emails
- get transactional email by ID

## Solution

Implement it duh

## Deployment Checklist

- [ ] NA

### DIFF
--- a/backend/src/email/routes/email-transactional.routes.ts
+++ b/backend/src/email/routes/email-transactional.routes.ts
@@ -234,6 +234,161 @@ export const InitEmailTransactionalRoute = (
     emailTransactionalMiddleware.rateLimit,
     emailTransactionalMiddleware.sendMessage
   )
+  /**
+   * @swagger
+   * paths:
+   *   /transactional/email:
+   *     get:
+   *       security:
+   *         - bearerAuth: []
+   *         - cookieAuth: []
+   *       tags:
+   *         - Email
+   *       summary: "List transactional emails"
+   *       parameters:
+   *         - in: query
+   *           name: limit
+   *           description: max number of messages returned
+   *           required: false
+   *           schema:
+   *             type: integer
+   *             minimum: 1
+   *             maximum: 100
+   *             default: 10
+   *         - in: query
+   *           name: offset
+   *           description: offset to begin returning messages from
+   *           required: false
+   *           schema:
+   *             type: integer
+   *             minimum: 0
+   *             default: 0
+   *         - in: query
+   *           name: status
+   *           description: status of messages to filter for
+   *           required: false
+   *           schema:
+   *             type: array
+   *             items:
+   *               type: string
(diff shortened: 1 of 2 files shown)


backend/src/email/routes/email-transactional.routes.ts
```diff
@@ -236,5 +236,243 @@ export const InitEmailTransactionalRoute = (
...
+   *
+   *       responses:
+   *         200:
+   *           description: Succcessfully retrieve a list of messages
+   *           content:
+   *             application/json:
+   *               schema:
+   *                 type: object
+   *                 required:
+   *                   - has_more
```
prestonlimlianjie, line 326:
```suggestion
   *                   - next_id
```

Perhaps provide a `next_id` instead, so that API users can rely on our API responses to paginate without having to do math on their end (i.e. `next_id` = `curr_id` + `offset`)

backend/src/email/routes/email-transactional.routes.ts
```diff
@@ -236,5 +236,243 @@ export const InitEmailTransactionalRoute = (
...
+   *                   - data
+   *                 properties:
+   *                   has_more:
+   *                     type: boolean
```
prestonlimlianjie, line 330:
```suggestion
   *                   next_id:
   *                     type: integer
```

See above comment

backend/src/email/routes/email-transactional.routes.ts
```diff
@@ -236,5 +236,243 @@ export const InitEmailTransactionalRoute = (
...
+   *       parameters:
+   *         - in: path
+   *           name: emailId
+   *           required: true
+   *           schema:
+   *             type: string
+   *             example: 69
+   *       responses:
+   *         200:
+   *            description: Succcessfully retrieve transactional email with corresponding ID
```
prestonlimlianjie, line 415:
```suggestion
   *         200:
   *           description: Succcessfully retrieve transactional email with corresponding ID
   *           content:
   *            application/json:
   *              schema:
   *                $ref: '#/components/schemas/EmailMessageTransactional'
```

backend/src/email/routes/email-transactional.routes.ts
stanleynguyen: We're doing `offset + limit` instead of a `start_id + limit` (or `end_id + limit`) though so giving `next_id` is not really helpful in this case

backend/src/email/routes/email-transactional.routes.ts
stanleynguyen: Oh shit forgot to commit this one 🤣

backend/src/email/routes/email-transactional.routes.ts
prestonlimlianjie: Ah I meant `next_id` as a `next_offset_value`. But this isn't a big deal. I'm ok with `has_more`.

prestonlimlianjie (approved):
lgtm

PR #1831, opened 2022-10-07
Fix/allow null managed list

prestonlimlianjie:
## Problem

Backend throws a 400 error if the user does not interact with the save managed list button.

## Solution

Make the API endpoint accept a null value for `should_save_list`


richard-yak (approved):
lgtm

PR #1832, opened 2022-10-09
refactor: user-related data models

tan-zx:
Tightened data models in `user.ts`, `user-demo.ts`, `user-credential.ts`, `user-feature.ts`, per https://github.com/opengovsg/postmangovsg/issues/1696

Previously (see discussion below and history of this PR description), renamed `api_key` to `api_key_hash`, but will now be dealt with in separate PRs (see https://github.com/opengovsg/postmangovsg/pull/1841 and https://github.com/opengovsg/postmangovsg/pull/1842)


backend/src/core/models/user/user-credential.ts
```diff
@@ -17,27 +17,25 @@ export class UserCredential extends Model<UserCredential> {
...
+  label: string
 
   @ForeignKey(() => Credential)
   @Column({
     type: DataType.STRING,
     allowNull: false,
   })
-  credName!: string
-
-  cred!: Credential
```
tan-zx, line 36: deleting this as it is dead code

tan-zx:
PENDING: to test on staging first (will remove this comment when done)

stanleynguyen (changes requested):
This can't be released this way, as there will be a gap after u have done the migration, but the new code haven't deployed yet - resulting in all API key related operations failing.
Please break this PR down to:
- a migration to creating a new column called `api_key_hash` with a trigger to insert this value to be the same as `api_key`
- backfill all existing `api_key` values to corresponding `api_key_hash` values
- deploy the new code
- another migration removing `api_key` field and the trigger from step (1)

tan-zx:
> This can't be released this way, as there will be a gap after u have done the migration, but the new code hasn't been deployed yet - resulting in all API key related operations failing. Please break this PR down to:
> 
> * a migration to creating a new column called `api_key_hash` with a trigger to insert this value to be the same as `api_key`
> * backfill all existing `api_key` values to corresponding `api_key_hash` values
> * deploy the new code
> * another migration removing `api_key` field and the trigger from step (1)

ohhh good point — will do!

stanleynguyen (approved):

PR #1835, opened 2022-10-09
feat: use jsonb instead of json throughout codebase

tan-zx:
### Background

 I was more exploring why the metadata is stored so weirdly, `{"{\"hash\": \"ebc2628a3fa93f210d783ce645edcaf1\", \"fileName\": \"zx.csv\", \"fileSize\": 31}"}`

But I have since concluded that it doesn't matter since both Sequelize and PostgreSQL recognise and can manipulate the data nicely. As such, this PR is not strictly necessary, but might be good-to-have nonetheless.

### Change

This PR converts `attachments_metadata` from `json[]` to `jsonb[]`. 

Some advantages of `jsonb` (basically better processing) from [here](https://stackoverflow.com/questions/22654170/explanation-of-jsonb-introduced-by-postgresql)
image

### SQL commands to run to check for length of array:

E.g. to find all emails with 2 attachments:
```SQL
SELECT *
FROM email_messages_transactional
WHERE array_length(attachments_metadata, 1) = 2;
```

## Deployment Checklist
- [ ] Run db migration (order doesn't matter)

Tested on staging (CI [here](https://github.com/opengovsg/postmangovsg/actions/runs/3239692198))

### DIFF
--- a/backend/src/core/models/campaign.ts
+++ b/backend/src/core/models/campaign.ts
@@ -64,7 +64,7 @@ export class Campaign extends Model<Campaign> {
   @BelongsTo(() => Credential)
   credential?: Credential
 
-  @Column(DataType.JSON)
+  @Column(DataType.JSONB)
   s3Object?: CampaignS3ObjectInterface
 
   @Column({
--- a/backend/src/database/migrations/20210422073013-create-campaigns.js
+++ b/backend/src/database/migrations/20210422073013-create-campaigns.js
@@ -35,6 +35,7 @@ module.exports = {
         },
         onUpdate: 'CASCADE',
       },
+      // superseded by 20221009141317-change-json-to-jsonb-everywhere.js
       s3_object: {
         type: Sequelize.DataTypes.JSON,
         allowNull: true,
--- a/backend/src/database/migrations/20210422073931-create-sms-messages.js
+++ b/backend/src/database/migrations/20210422073931-create-sms-messages.js
@@ -22,6 +22,7 @@ module.exports = {
         type: Sequelize.DataTypes.STRING(255),
         allowNull: true,
       },
+      // superseded by 20221009141317-change-json-to-jsonb-everywhere.js
       params: {
         type: Sequelize.DataTypes.JSON,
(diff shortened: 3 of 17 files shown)


backend/src/database/migrations/20221009141317-change-attachments-metadata-to-jsonb-array.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.changeColumn(
```
stanleynguyen, line 5: Are we sure we can just do this without taking care of the existing rows?

tan-zx: yup it works — tested on local and will test on staging

haha I wonder if this PR is necessary — now that I think about it, I don't really think we are running operations on this column now. it's just that in the process of researching this, I saw somewhere that says a `json[]` is almost always an error 🤔

(relatedly, wonder if a similar change should be made to the other columns like `params` in `email_messages` and `email_messages_transactional`)

stanleynguyen: hmm if that's the case, let's update everything if we're to update. I'd rather have bad but consistent code than inconsistently  good code

tan-zx: OK I will update everything then

stanleynguyen (changes requested):
We should update the datatype in `Column` decorators in sequelize models too

tan-zx:
> We should update the datatype in `Column` decorators in sequelize models too

Thanks good catch, have updated in latest commit (and rebased on master)

tan-zx:
> We should update the datatype in `Column` decorators in sequelize models too

Thanks good catch, have updated in latest commit (and rebased on master)

stanleynguyen (approved):

PR #1838, opened 2022-10-10
build: add serverless

prestonlimlianjie:
A draft PR with single serverless.yml and a dummy hello world lambda function - tested deployment on staging.


prestonlimlianjie:
Closing this PR since we are likely to use sst.dev moving forward.

PR #1839, opened 2022-10-11
feat: allow use of postman template in transactional email

lamkeewei:
## Problem

Plain text emails often look phishy. Postman has spent quite a bit of effort in developing a template that can help improve legitimacy of the emails sent. This feature should be available for transactional emails also.

## Solution
- Add `use_template` flag to allow for the user of template in transactional email.

### DIFF
--- a/backend/src/email/middlewares/email-transactional.middleware.ts
+++ b/backend/src/email/middlewares/email-transactional.middleware.ts
@@ -113,8 +113,10 @@ export const InitEmailTransactionalMiddleware = (
       reply_to: replyTo,
       attachments,
       emailMessageTransactionalId, // added by saveMessage
+      use_template: useTemplate,
     }: ReqBody & {
       emailMessageTransactionalId: number
+      use_template?: boolean
     } = req.body
 
     try {
@@ -127,6 +129,7 @@ export const InitEmailTransactionalMiddleware = (
           replyTo ?? (await authService.findUser(req.session?.user?.id))?.email,
         attachments,
         emailMessageTransactionalId,
+        useTemplate,
       })
       await EmailMessageTransactional.update(
         {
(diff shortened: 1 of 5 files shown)


lamkeewei:
Oops, didn't realise we're using `master` as the base branch now.

prestonlimlianjie:
Thanks for the PR @lamkeewei - we might just remove the param and make all transactional emails use the template. Thoughts @lisatjide @stanleynguyen @tan-zx?

lamkeewei:
> Thanks for the PR @lamkeewei - we might modify your PR to just remove the `use_template` param and make all transactional emails use the template (to keep it consistent with campaign mode). Thoughts @lisatjide @stanleynguyen @tan-zx?

Make sense to me. My primary concern was to make sure that we don't surprise existing users (if any).

tan-zx:
agree with Stanley — think we should give them the flexibility to preserve their own branding. conversely, using the template for simple OTP emails might also be too heavy.

making this change mandatory will also make rollout difficult (ideally, we would need to tell them that we're making this change ahead of time and give them enough time to make alternative plans)

lamkeewei:
> > Thanks for the PR @lamkeewei - we might modify your PR to just remove the `use_template` param and make all transactional emails use the template (to keep it consistent with campaign mode). Thoughts @lisatjide @stanleynguyen @tan-zx?
> 
> 
> 
> Some of our users might not want that cos they might have their own template for their own application though

Actually on the point of users having their own templates, it's actually quite hard to style the message now because of our quite restrictive xss rules. 

My original plan was to copy the raw HTML for postman and template and render it on my end before calling the API. But this was not possible because of the xss filtering rules.

stanleynguyen:
We're still managing okay with XSS relaxation so putting this in our backburner

PR #1841, opened 2022-10-12
refactor: api key hash

tan-zx:
Part 1 of 2 to rename `api_key` to `api_key_hash`

What this PR does:
- create new `api_key_hash` column
- create a trigger that ensures that any new values written to `api_key` will sync with `api_key_hash` column
- run a SQL command to backfill existing `api_key` values to `api_key_hash`
- create a trigger `format_user_email` that is meant to replace the existing (also poorly named) `format_user_on_insert`

See [here](https://github.com/opengovsg/postmangovsg/pull/1832#pullrequestreview-1136705832) for context

**Deployment notes**: safe to just deploy, order doesn't matter.


backend/src/database/migrations/20221012051800-add-api-key-hash-column-to-user.js
```diff
@@ -0,0 +1,36 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('users', 'api_key_hash', {
+      type: Sequelize.DataTypes.STRING(255),
+      allowNull: true,
+    })
+    await queryInterface.sequelize.query(
+      'UPDATE users SET api_key_hash = api_key;'
```
stanleynguyen, line 10: Trigger should be created before backfill, otherwise there's still a slim chance a new insert on users table happened before the trigger has been created resulting in a row without `api_key_hash`

backend/src/database/migrations/20221012051800-add-api-key-hash-column-to-user.js
```diff
@@ -0,0 +1,36 @@
...
+    )
+    await queryInterface.createFunction(
+      'format_user_email',
+      [],
+      'trigger',
+      'plpgsql',
+      `
+      NEW.email = LOWER(NEW.email);
```
stanleynguyen, line 18: I think u forgot to modify this trigger copied from another migration?

backend/src/database/migrations/20221012051800-add-api-key-hash-column-to-user.js
tan-zx: Hmm I see your point, but I don't think it's relevant here. 

This new trigger (`format_user_email`) is meant to replace `format_user_on_insert`, which (1) normalizes user emails by ensuring they're lower case; and (2) if the manual edit to the user row did not specify `api_key`, i.e. it is `''`, it'll be automatically set to `NULL`.

What `format_user_email` does is to only do (1) and not do (2). I don't think (2) is necessary and was in fact a hidden coupling that was a hindrance to this refactoring (I ran into an error and had to debug to find out that `api_key` was referred to in this db function.

When this trigger is created, both `format_user_email` and `format_user_on_insert` would exist concurrently (the latter is only removed in part 2), so there is no risk of any missed insertion.

Hope this makes sense?

backend/src/database/migrations/20221012051800-add-api-key-hash-column-to-user.js
tan-zx: I'm not sure what you mean, but the omission of any reference to `api_key` is intentional! See [here](https://github.com/opengovsg/postmangovsg/pull/1841/files#r994095300).

backend/src/database/migrations/20221012051800-add-api-key-hash-column-to-user.js
tan-zx: Thanks for the clarification!!

backend/src/database/migrations/20221012051800-add-api-key-hash-column-to-user.js
```diff
@@ -0,0 +1,58 @@
...
+    )
+    // this trigger is needed to ensure that any users that generate new api keys AFTER the backfill will be synced to api_key_hash
+    await queryInterface.sequelize.query(
+      'CREATE TRIGGER sync_api_key_with_api_key_hash_trigger BEFORE INSERT OR UPDATE ON users FOR EACH ROW EXECUTE PROCEDURE sync_api_key_with_api_key_hash();'
+    )
+    await queryInterface.sequelize.query(
+      'UPDATE users SET api_key_hash = api_key;'
+    )
+    await queryInterface.createFunction(
+      'format_user_email',
```
stanleynguyen, line 29: Can you remind me why we need this trigger again?

tan-zx: this is for when prod ops (or anyone with db access) adds a user directly into the database (typically from a nonwhitelisted domain/.gov.sg account), to ensure that the email address will be normalised (i.e. all lower case)

stanleynguyen: I thought we already have that trigger from another migration? Why are we adding it again here?

tan-zx: oh so that trigger (added by the previous team) does two things:
1. ensure the user's email is lower case
2. making the edit easier by allowing the DBA to just leave `api_key` blank and the trigger would automatically convert it to `NULL`

I am adding it here to only do 1 and to NOT to do 2. That trigger will be dropped in the `part 2` PR. Other options are:
- remove the trigger entirely (i.e. don't do 1)
- replicate the functionality of that trigger (so that the DBA can leave `api_key_hash` blank and the trigger automatically convert it to null.

Haha I chose to keep 1 but drop 2 because I think: 1 is quite important to have, in case the DBA makes an error, could really cause some downstream errors; 2 is an annoying hidden coupling that's not that useful anyway.

Hope this clarifies!

image

stanleynguyen: Hmm but I don't see any part of this code dropping the trigger created by the previous team though 🤔

tan-zx: Oh I dropped this trigger in the [next PR](https://github.com/opengovsg/postmangovsg/pull/1842/files) instead

In the unlikely even there are manual db inserts between these two PRs, the trigger should still run hahah

stanleynguyen: Why can't we just drop the trigger from the previous team in this PR?

tan-zx: Hmm yeah upon reflection ya we can drop it here. Please wait let me change the PR!

stanleynguyen (approved):
Don't forget to test in staging first

PR #1842, opened 2022-10-12
refactor: api key hash part 2

tan-zx:
What this PR does:
- Change backend code to write to `api_key_hash` instead of `api_key`
- Drop previously created trigger that syncs `api_key` and `api_key_hash`
- Drop `format_user_on_insert` function and trigger

**Deployment notes**:
1. Deploy backend code 
2. Ensure backend is fully deployed before running db migration

See [here](https://github.com/opengovsg/postmangovsg/pull/1832#pullrequestreview-1136705832) for context

### DIFF
--- a/backend/src/database/migrations/20221012054742-remove-api-key-column-from-user-table.js
+++ b/backend/src/database/migrations/20221012054742-remove-api-key-column-from-user-table.js
@@ -1,0 +1,36 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, _) => {
+    await queryInterface.removeColumn('users', 'api_key')
+    await queryInterface.sequelize.query(
+      'DROP TRIGGER sync_api_key_with_api_key_hash_trigger ON users;'
+    )
+    await queryInterface.dropFunction('sync_api_key_with_api_key_hash', [])
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.createFunction(
+      'sync_api_key_with_api_key_hash',
+      [],
+      'trigger',
+      'plpgsql',
+      `
+      NEW.api_key_hash = NEW.api_key;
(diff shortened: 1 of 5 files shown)


backend/src/database/migrations/20221012054742-remove-api-key-column-from-user-table.js
```diff
@@ -0,0 +1,37 @@
...
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.createFunction(
+      'format_user_on_insert', // actually a misnomer, since activates on UPDATE too, but stick to existing naming
+      [],
+      'trigger',
+      'plpgsql',
+      `
+      NEW.email = LOWER(NEW.email);
+      NEW.api_key = CASE WHEN NEW.api_key='' THEN NULL ELSE NEW.api_key END;
```
stanleynguyen, line 20: Incorrect trigger function body

stanleynguyen: Should set `api_key_hash = api_key`

tan-zx: ohh I think in this case your [previous comment](https://github.com/opengovsg/postmangovsg/pull/1841/files#r994058139) is correct, in the down migration, I should create the column and backfill first, before creating the trigger and function

(let me test again locally to make sure)

tan-zx: alternatively, lmk (via Slack) if we should hop on a call to discuss this (none of this is particularly urgent tho, so no hurry)

stanleynguyen (approved):

stanleynguyen (approved):

Issue #1837, opened 2022-10-10
Uncaught error if from address is not verified for transactional email endpoint


lamkeewei:
<img width="847" alt="image" src="https://user-images.githubusercontent.com/3666479/194825726-32e03a3b-3a9a-4eb1-955a-8642f42a4d96.png">

tan-zx:
We actually noticed and discussed this earlier (see [here](https://github.com/opengovsg/postmangovsg/pull/1778#issuecomment-1264544947)) and @stanleynguyen is in the process of building out an endpoint for verifying from address (see [here](https://opengovproducts.slack.com/archives/CT8D1FESY/p1665651696739409))

Actually, I was typing this, I realize a simple way would be to check out `email_from_address` table (rather than actually run the check with AWS SES 🤔)

PR #1846, opened 2022-10-14
tests: transactional email sending and saving

tan-zx:
tests for transactional email sending and saving
- include tests as promised in https://github.com/opengovsg/postmangovsg/pull/1768
- include tests for API transactional emails to cover edge cases that were previously omitted (e.g. malicious attachments etc.)

includes some light refactoring

I originally created these files too, but as discussed with @stanleynguyen, these integration tests should be able to cover everything we need at this point

image


backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -31,7 +52,10 @@ beforeAll(async () => {
   )
 })
 
-afterEach(() => jest.resetAllMocks())
```
tan-zx, line 34: not very sure why this was chosen, but I had to change to `restoreAllMocks` to get rid of side effects so that tests running in series would pass. (before that, they would pass individually but not as a group.)

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -55,85 +92,107 @@ describe('POST /transactional/email/send', () => {
...
+    expect(transactionalEmail?.params).toMatchObject({
+      subject: validApiCall.subject,
+      body: validApiCall.body,
+      from: 'Hello <donotreply@mail.postman.gov.sg>',
+      replyTo: user.email,
+    })
   })
 
+  // something I don't understand — is the verification of the from address mocked? or added to `email_from_address`?
+  // afaict, user.email is not added to `email_from_address` in the test setup, yet somehow the test passes
```
tan-zx, line 159: to remove before merging in (genuinely not sure how it's done and whether there's a lesson here for how I should write my tests differently if so)

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -142,92 +201,411 @@ describe('POST /transactional/email/send', () => {
...
-        })
+        .send(validApiCall)
     }
 
     // First request passes
     let res = await send()
     expect(res.status).toBe(202)
     expect(mockSendEmail).toBeCalledTimes(1)
     mockSendEmail.mockClear()
+    // should only return 1
```
tan-zx, line 492: oops to remove

backend/src/email/services/email-transactional.service.ts
```diff
@@ -35,16 +43,18 @@ async function sendMessage({
...
       {
-        errorCode: 'Error 400: Message contains invalid HTML tags',
+        errorCode: EMPTY_MESSAGE_ERROR_CODE,
       },
       {
         where: { id: emailMessageTransactionalId },
       }
     )
-    throw new InvalidMessageError()
+    throw new EmptyMessageError()
```
tan-zx, line 57: refactored this for greater precision

backend/src/email/utils/query.ts
```diff
@@ -1,13 +0,0 @@
-import { EmailBlacklist } from '@email/models'
```
tan-zx, line 1: deleted this and included in `EmailService`
- fits better there
- allows me to mock the return value

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -142,92 +201,411 @@ describe('POST /transactional/email/send', () => {
...
+      },
+      {
+        fileName: validAttachment2Name,
+        fileSize: validAttachment2Size,
+        hash: validAttachment2Hash,
+      },
+    ])
+  })
+
+  // test('Message with too many attachments should fail', async () => {
```
tan-zx, line 477: oops missed this, will do later

backend/src/core/services/file-attachment.service.ts
```diff
@@ -43,19 +48,19 @@ const sanitizeFiles = async (
...
+        errorCode: UNSUPPORTED_FILE_TYPE_ERROR_CODE,
       },
       {
         where: { id: emailMessageTransactionalId },
       }
     )
     throw new UnsupportedFileTypeError()
   }
-  const isSafe = await virusScan(files)
+  const isSafe = await FileAttachmentService.areFilesSafe(files)
```
stanleynguyen, line 59: This function is in this file, we don't have to `FileAttachmentService.`

stanleynguyen: This is creating unnecessary depending on the `FileAttachmentService` object

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -25,6 +25,14 @@ export interface EmailTransactionalMiddleware {
   rateLimit: Handler
 }
 
+export const RATE_LIMIT_ERROR_MESSAGE =
+  'Error 429: Too many requests, rate limit reached'
+
+export const getMd5HashFromAttachment = (content: Buffer): string => {
```
stanleynguyen, line 31: Why are we changing this function name? Adding `Md5` and `From` doesn't provide any extra context. Because for the reader, I don't need to care about the underlying algo and preposition

backend/src/core/errors/transactional.errors.ts
```diff
@@ -6,10 +6,11 @@ export class RateLimitError extends Error {
...
 }
 
-export class InvalidMessageError extends Error {
-  constructor(
-    message = 'Message template is invalid as it only contains invalid HTML tags'
-  ) {
+export const EMPTY_MESSAGE_ERROR =
+  'Message is empty after removing invalid HTML tags'
+
+export class EmptyMessageError extends Error {
```
stanleynguyen, line 12: This name change is misleading cos the original message is not `empty` it's `invalid`

backend/src/email/services/email-transactional.service.ts
stanleynguyen: See [above](https://github.com/opengovsg/postmangovsg/pull/1846/files#r1005439545)

backend/src/email/middlewares/email-transactional.middleware.ts
stanleynguyen: Also, please don't export this function, u can just mock the underlying `crypto` package. This function is intentionally kept private

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -20,7 +39,9 @@ beforeEach(async () => {
     email: `user_${userId}@agency.gov.sg`,
   } as User)
   apiKey = await user.regenerateAndSaveApiKey()
-  userId += 1
+  mockSendEmail = jest
```
stanleynguyen, line 42: Please don't move this here. It's over-generalisation. For tests, the default approach should be sharing as little as possible. Test cases should be easily readable and understandable in isolation

stanleynguyen: You need a very very very very good reason for generalisation in test cases

backend/src/email/routes/tests/email-transactional.routes.test.ts
stanleynguyen: Because we're mocking the email sending, ofc it passes

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -142,92 +201,435 @@ describe('POST /transactional/email/send', () => {
     expect(mockSendEmail).not.toBeCalled()
   })
 
-  test('Should send a message with a valid attachment', async () => {
-    const mockSendEmail = jest
-      .spyOn(EmailService, 'sendEmail')
-      .mockResolvedValue('message_id')
+  test('Should throw an error if email subject or body is empty after removing invalid HTML tags', async () => {
```
stanleynguyen, line 204: This test case is testing way more than this title suggests, it's testing for the email being persisted in the db with the corresponding error as well

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -142,92 +201,435 @@ describe('POST /transactional/email/send', () => {
...
+      status: TransactionalEmailMessageStatus.Unsent,
+    })
+    expect(transactionalEmail?.params).toMatchObject({
+      from: validApiCall.from,
+      replyTo: validApiCall.reply_to,
+    })
+    expect(transactionalEmail?.errorCode).toBe(UNSUPPORTED_FILE_TYPE_ERROR_CODE)
+  })
+
+  test('Should throw an error if attached file is malicious', async () => {
```
stanleynguyen, line 284: Can we be more specific about the `error` here?

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -142,92 +201,435 @@ describe('POST /transactional/email/send', () => {
...
+      status: TransactionalEmailMessageStatus.Unsent,
+    })
+    expect(transactionalEmail?.params).toMatchObject({
+      from: validApiCall.from,
+      replyTo: validApiCall.reply_to,
+    })
+    expect(transactionalEmail?.errorCode).toBe(MALICIOUS_FILE_ERROR_CODE)
+  })
+
+  test('Should throw an error if recipient is blacklisted', async () => {
```
stanleynguyen, line 324: Can we be more specific about the `error` here?

stanleynguyen (changes requested):
image

backend/src/core/errors/transactional.errors.ts
tan-zx: so this error is only thrown if the message becomes empty after removing the invalid HTML tags

normally, if there are invalid HTML tags, they will just be automatically removed and this error won't be thrown. that's why I renamed it

(you can see that's what the original message was getting at `message = 'Message template is invalid as it only contains invalid HTML tags'` → it becomes empty after the invalid HTML tags are removed)

backend/src/core/services/file-attachment.service.ts
tan-zx: Ohh so the reason I wrapped it in the `FileAttachmentService` object is so that I can mock the value here:
image

Is there some other way I can achieve this without creating this dependency 🤔

backend/src/email/middlewares/email-transactional.middleware.ts
tan-zx: > Also, please don't export this function, u can just mock the underlying crypto package. This function is intentionally kept private

Ohh that's a very good point, let me change it back.

Can you show me how to mock the underlying `crypto` package? I think I have seen you do it in a previous PR before, but I don't fully understand how that works.

backend/src/email/routes/tests/email-transactional.routes.test.ts
tan-zx: Fair enough! I was trying to generalise the code because I saw a lot of repetition, but I can totally see this particular `mockSendEmail` as going one step too far.

backend/src/email/routes/tests/email-transactional.routes.test.ts
tan-zx: OK will fix the title

backend/src/email/routes/tests/email-transactional.routes.test.ts
tan-zx: Ohh I was confused why this tests passes and why the next test fails.

```TS
test('Should thrown an error with invalid custom from address', async () => {
    const res = await request(app)
      .post(endpoint)
      .set('Authorization', `Bearer ${apiKey}`)
      .send({
        ...validApiCall,
        from: 'Hello <invalid@agency.gov.sg>',
        reply_to: user.email,
      })

    expect(res.status).toBe(400)
    expect(mockSendEmail).not.toBeCalled()
})
```

But looking back I think I figured out what's going on.

I was expecting a db call to `EmailFromAddress` to check whether the the custom from address is in our database. This is used e.g. in `EmailMiddleware.existsFromAddress`, which is called for email campaigns.

In `email-transactional.routes.ts`, we call `emailMiddleware.isFromAddressAccepted`, which simply checks whether `(fromAddress !== userEmail && fromAddress !== defaultFromAddress)`. 

Per our previous discussion, we said it's enough to have an API for users to check that their custom from email address is acceptable and just throw them a 500 error if they pass the check above but fail at the AWS SES (see this [issue](https://github.com/opengovsg/postmangovsg/issues/1837)).

backend/src/email/routes/tests/email-transactional.routes.test.ts
tan-zx: just checking, this and all the ones below are the same as [this comment](https://github.com/opengovsg/postmangovsg/pull/1846#discussion_r1005448762) right, which is to use a longer title to describe the test in full?

stanleynguyen: yesser

backend/src/core/errors/transactional.errors.ts
stanleynguyen: > so this error is only thrown if the message becomes empty after removing the invalid HTML tags
> 
> normally, if there are invalid HTML tags, they will just be automatically removed and this error won't be thrown. that's why I renamed it
> 
> (you can see that's what the original message was getting at `message = 'Message template is invalid as it only contains invalid HTML tags'` → it becomes empty after the invalid HTML tags are removed)
> 
> to be exact, it could be either the subject or the body that is empty, but if at least either one is empty, the error will be thrown

Hmm but I feel that we don't need that kinda granularity here. `Invalid` would be more future-proof in case we add more rules to this validation

backend/src/email/middlewares/email-transactional.middleware.ts
stanleynguyen: Use [this](https://jestjs.io/docs/manual-mocks)

backend/src/core/services/file-attachment.service.ts
stanleynguyen: > Ohh so the reason I wrapped it in the `FileAttachmentService` object is so that I can mock the value here: image
> 
> Is there some other way I can achieve this without creating this dependency 🤔

No u shouldn't be exporting a private function just so that u can mock it - this is a really really bad reason to export, mock the external dependency that it relies on

backend/src/email/routes/tests/email-transactional.routes.test.ts
tan-zx: Oh I guess it's not in the sense that some of the test case codepath would've terminated before hitting the `sendEmail` part? But otherwise the tests all assume that `sendEmail` would be successful?

backend/src/core/errors/transactional.errors.ts
tan-zx: Hmm but I must say that in the beginning, when I was trying to cause this error before I looked at the code in detail, I assumed that the error would be thrown if invalid HTML tags were used or if the entire message was wrapped around invalid HTML tags.

Then after poking further, I realise that our code removed the HTML tags directly and only throws the error in the edge case that this removal causes the field to be empty.

I would prefer to be a bit more specific, but it's no big deal either way. Will defer to your decision.

backend/src/email/routes/tests/email-transactional.routes.test.ts
stanleynguyen: > Oh I guess it's not in the sense that some of the test case codepath would've terminated before hitting the `sendEmail` part? But otherwise the tests all assume that `sendEmail` would be successful?

That's assumption is not true (or at least not for long). We dont have it here yet but we might need to handle errors from `sendEmail` and write tests for that

backend/src/core/errors/transactional.errors.ts
stanleynguyen: > Hmm but I must say that in the beginning, when I was trying to cause this error before I looked at the code in detail, I assumed that the error would be thrown if invalid HTML tags were used or if the entire message was wrapped around invalid HTML tags.
> 
> Then after poking further, I realise that our code removed the HTML tags directly and only throws the error in the edge case that this removal causes the field to be empty.
> 
> I would prefer to be a bit more specific, but it's no big deal either way. Will defer to your decision.

I think that `Empty` is actually more misleading, as the original message is **not** empty

backend/src/email/routes/tests/email-transactional.routes.test.ts
tan-zx: Do you think this is something we should add? (Could be a separate PR, would handle the issue Kee Wei flagged as well)

backend/src/core/errors/transactional.errors.ts
tan-zx: Have renamed it `MessageError` — feel like it's more consistent with the `TemplateError` [thrown elsewhere](https://github.com/opengovsg/postmangovsg/blob/08da008200a75721bf0a23283f65409ad2e89d2d/backend/src/email/services/email-template.service.ts#L180-L184). 

Have added the test cases too.

backend/src/email/routes/tests/email-transactional.routes.test.ts
tan-zx: As I'm adding this back into each test, I think I see your point about overgeneralisation.

Basically, for the tests that fail even before reaching the `sendEmail` stage, there is no need to mock their resolved value, so mocking resolved value for all tests is an overkill!

tan-zx:
I have fixed the stuff mentioned in the resolved convos (have made discrete commits so it's easier to track changes).

I'm not very sure how to use the manual mock you linked and how to mock the external dependency for the malicious file test → would be great if we could find some time to go through them? 🙏 (I have previously encountered this issue and have struggled to fix it... teach me sempai)

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -55,179 +89,626 @@ describe('POST /transactional/email/send', () => {
...
+      subject: invalidHtmlTagsSubjectAndBody.subject,
+      body: invalidHtmlTagsSubjectAndBody.body,
+      from: validApiCall.from,
+      replyTo: validApiCall.reply_to,
+    })
+    expect(transactionalEmail?.errorCode).toBe(null)
+
+    expect(mockSendEmail).toBeCalledWith(
+      {
+        subject: '<br />HELLO<br />',
```
tan-zx, line 287: realize that we actually insert `<br />` into the subject since we're using the same function `replaceNewLinesAndSanitize` for both subject and body. 

it doesn't affect the email sent, but just wanted to check if this is okay?

or if we should use two separate functions for sanitizing subject vs body

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -55,179 +89,626 @@ describe('POST /transactional/email/send', () => {
...
+    })
+    expect(transactionalEmail?.errorCode).toBe(EMPTY_MESSAGE_ERROR_CODE)
+  })
+
+  test('Should send email if subject and body are not empty after removing invalid HTML tags and metadata is saved correctly in db', async () => {
+    mockSendEmail = jest
       .spyOn(EmailService, 'sendEmail')
       .mockResolvedValue('message_id')
+    const invalidHtmlTagsSubjectAndBody = {
+      subject: 'HELLO',
```
tan-zx, line 251: as I was writing this test, I realize that we're actually replacing `\n` and `\r\n` in the subject header with `<br />` as part of the sanitization process. I thought it's weird because there shouldn't be line breaks in the subject line at all. 

conversely, it seems like we're not doing this for the body? I'm not quite sure what's going on there (the comments suggest that it's doing this so that things render properly on the frontend, but I still don't fully understand), but I don't think it's worth the effort to fix.

backend/src/core/services/file-attachment.service.ts
```diff
@@ -68,4 +73,5 @@ const sanitizeFiles = async (
 
 export const FileAttachmentService = {
   sanitizeFiles,
+  areFilesSafe,
```
stanleynguyen, line 76: This is still exported solely for mocking. Please fix

backend/src/email/middlewares/email-transactional.middleware.ts
stanleynguyen: Unresolved comment

tan-zx:
Fixed the tests per PR comments! Remaining unresolved convos are either just random comments or suggestions to be taken up in subsequent PRs.

backend/src/email/services/email-transactional.service.ts
```diff
@@ -58,11 +68,11 @@ async function sendMessage({
       )
     : undefined
 
-  const blacklisted = await isBlacklisted(recipient)
+  const blacklisted = await EmailService.isRecipientBlacklisted(recipient)
```
tan-zx, line 71: I moved this from the utils folder into `EmailService` — not 100% sure whether this is correct, hence flagging this.

stanleynguyen (approved):

backend/src/email/routes/tests/email-transactional.routes.test.ts
stanleynguyen: Yup we will need to handle that error, but not in this PR. Don't think it's a big issue (yet) since we're still manually onboarding our B2B users and not leaving then to self-service yet

Issue #1179, opened 2021-05-07
Sort imports using eslint-plugin-import/eslint-plugin-simple-import-sort


zwliew:
Discussion in #1129. It would be a good idea to sort to standardize the import order across files. Can refer to vaccine ops.

jhibberd-gov:
Moved to Notion.

PR #1850, opened 2022-10-20
style: eslint sort imports

tan-zx:
Closes https://github.com/opengovsg/postmangovsg/issues/1179

Changes:
- Modelled `eslint` rules after `ts-template`. See first commit. Compared to `ts-template` rules I commented out `import/no-extraneous-dependencies` because...we actually have a lot of extraneous dependencies (which require even more knock-on changes to fix). 
  - If we're keen to have the rule, am happy to fix in subsequent commits (or PR), before this one gets too big
- For ease of review, see the commits (just skip the 3rd one, where I apply the linting rules, which is where a lot of files were changed).

Some changes I had to make to make the tests pass again (see 4th commit):
- Cannot use alias `'@core/services'` or the test will fail (likely to do with how aliases were set up? some webpack config thing?)
- Had to use `/* eslint-disable import/first */` to disable entire file for `backend/src/core/services/tests/upload.service.test.ts` because of how the mocking is set up (I actually don't understand how the test in that file works! looks like some cool currying thing)

Tested on staging, works fine.


tan-zx:
for reasons I don't fully understand, keep failing to deploy on staging — will fix and try again later

stanleynguyen:
Hmm I think this is quite high effort low impact as this is just subject code style - there's no real/significant performance improvement whatsoever, I'd prefer we keep the current style and scrap this suggestion

tan-zx:
> Hmm I think this is quite high effort low impact as this is just subjective code style - there's no real/significant performance improvement whatsoever

ya agree that it doesn't impact performance, but I do think it's helpful to have sorted imports for readability though, especially when the import lines get very long. 

haha it's not that high effort la, like for the new stuff we're merging in, we just have to run `npm run lint` and it'll automatically fix. 

hope you'll reconsider!

stanleynguyen:
> > Hmm I think this is quite high effort low impact as this is just subjective code style - there's no real/significant performance improvement whatsoever
> 
> ya agree that it doesn't impact performance, but I do think it's helpful to have sorted imports for readability though, especially when the import lines get very long.

I don't think anyone in their right mind would read the import lines from top to bottom - the usual flow is you just use your IDE to locate the imported module. My problem with this (a big one) is that it stops the compilation during development if I'm importing something new in and the autocompletion doesn't put it in the right place (not sure if your newly added rule to the backend will cause that, but pretty sure it's already happening on the frontend and it's annoying asf) 

> 
> haha it's not that high effort la (or rather, I have incurred the effort alr haha), like for the new stuff we're merging in, we just have to run `npm run lint` and it'll automatically fix.
> 
> hope you'll reconsider!

The high effort is not in running the linting command, the high effort is trying to rebase this into on-going PRs. I got no issue with refactoring the code style, but we can only get this in during "down" period of open PRs

tan-zx:
> My problem with this (a big one) is that it stops the compilation during development if I'm importing something new in and the autocompletion doesn't put it in the right place (not sure if your newly added rule to the backend will cause that, but pretty sure it's already happening on the frontend and it's annoying asf)

Ah I see your point. Agree that it's quite annoying in that case. I think we can fix that — we can just change it all to warning instead of error (on the frontend as well) so that it won't stop compilation.

tan-zx:
Oh I was thinking more like warning while you're developing, but then you can get rid of all of them before merging in by running `npm run lint`. GitHub will show these warnings too, so we'll be reminded to fix them 🤔

stanleynguyen:
> Oh I was thinking more like warning while you're developing, but then you can get rid of all of them before merging in by running `npm run lint`. GitHub will show these warnings too, so we'll be reminded to fix them 🤔

Trust me no one is gonna do that unless they're not able to do other things that they were intending to

tan-zx:
> > Oh I was thinking more like warning while you're developing, but then you can get rid of all of them before merging in by running `npm run lint`. GitHub will show these warnings too, so we'll be reminded to fix them 🤔
> 
> Trust me no one is gonna do that unless they're not able to do other things that they were intending to

😂 

This is what I meant though, like warnings will show up during code review and we can just run `npm run lint` to fix it before merging in the PR! 

image

PR #1851, opened 2022-10-24
style: apply no extraneous deps

tan-zx:
turned on the `import/no-extraneous-dependencies` rule in ESLint in `shared`, `backend`, `frontend`, and `worker`.

Got this error for frontend, so while I fixed the errors flagged by installing the dependencies, I turned the rule off.
image

based on my research, the solutions are to:
- drop the `@` from `@shared`
- install another dependency (`eslint-import-resolver-custom-alias` or `eslint-import-resolver-alias`) and configure to fix it

both solution seem too troublesome, so I turned off the rule instead.

I don't really understand — how were we able to import dependencies that we had not explicitly installed? were they installed as part of peer dependencies? so weird.

### DIFF
--- a/backend/.eslintrc.js
+++ b/backend/.eslintrc.js
@@ -96,7 +96,7 @@ module.exports = {
     'import/first': 'error',
     'import/newline-after-import': 'error',
     'import/no-duplicates': 'error',
-    // 'import/no-extraneous-dependencies': 'error',
+    'import/no-extraneous-dependencies': 'error',
   },
   overrides: [
     // disable rules specifically for JavaScript files
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -96,7 +96,9 @@ module.exports = {
     'simple-import-sort/exports': 'error',
     'import/first': 'error',
     'import/newline-after-import': 'error',
-    'import/no-duplicates': 'error',
-    // 'import/no-extraneous-dependencies': 'error',
+    'import/no-duplicates': 'error', 
+    // fixed errors flagged by import/no-extraneous-dependencies, but keep getting 
+    // 'shared' should be listed in the project's dependencies
+    // 'import/no-extraneous-dependencies': 'error'*,
   },
 }
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -6,6 +6,7 @@
     "": {
       "name": "frontend",
       "dependencies": {
+        "@lingui/core": "^3.14.0",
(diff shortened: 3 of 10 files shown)


tan-zx:
tested on staging, works

stanleynguyen:
> turned on the `import/no-extraneous-dependencies` rule in ESLint in `shared`, `backend`, `frontend`, and `worker`.
> 
> Got this error for frontend, so while I fixed the errors flagged by installing the dependencies, I turned the rule off. image
> 
> based on my research, the solutions are to:
> 
> * drop the `@` from `@shared`
> * install another dependency (`eslint-import-resolver-custom-alias` or `eslint-import-resolver-alias`) and configure to fix it
> 
> both solution seem too troublesome, so I turned off the rule instead.
> 
> I don't really understand — how were we able to import dependencies that we had not explicitly installed? were they installed as part of peer dependencies? so weird.

It's installed and bundled in the build of `shared` folder (the `shared` code will be pointed to import from `node_modules` of `shared`). It's actually harmful to have them listed in other parts of our system that uses `shared` as now we have to maintain the versioning there as well (The only time we would want to use this is when we actually import those modules in that part of the system as well - for e.g. we have `dd-trace` in `shared` but we're directly importing `dd-trace` for usage in our workers as well that's why we have in both places)

stanleynguyen:
> Ohh no, I think my original comment was unclear but there are two separate issues.
> 
> The first is the missing dependencies that we were able to import — so if you look at the diffs these are packages like:
> 
> * frontend: `@lingui/core`, `immutable`, `make-plural`
> * backend: N/A
> * worker: `validator`
> * shared: `opentracing`
>   These are all currently in the `package-lock.json` but not the `package.json` files of the respective folders. I think we should include them in the `package.json` file explicitly? I don't think we are duplicating dependencies that are used in `shared` in the other folders.

ah ok yup sounds good

> 
> The second is how to fix the remaining errors which flag up imports from `@shared`, which is caused by the ESLint rules misunderstanding how our alias works. I think this is the part where I wasn't sure how to fix and was wondering if we should just turn off the rule (while still implementing the change in the first part).
> 
> Does that make sense or am I missing something?

For this one we will need to include `module-alias` in eslint parser. Try fixing it with these 2 approaches (the first one is the better approach if it works):
- Include `import 'module-alias/register';` or `require('module-alias/register');` on top of `eslintrc.js`
- Try [this module](https://www.npmjs.com/package/eslint-import-resolver-alias) (but u will have to duplicate the path resolution from tsconfig)

PR #1855, opened 2022-10-25
feat(backend): show presence of authorization header in request log instead of blacklisting

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Combine-log-statements-and-include-cookie-API-key-user-info-82c71a7b32884cdd83e4c9778169f841)

## Solution

Instead of blacklisting, we will just redact the value of `authorization` header

## Deployment Checklist

- [ ] N/A


tan-zx (approved):
lgtm

PR #1856, opened 2022-10-26
feat(frontend): introduce feedback modal after sending campaign

richard-yak:
## Problem

To enable users to submit feedback via embedded formsg link

## Solution

- Create new FeedbackModal component to allow more modular use in case any other component needs to embed feedback form in other areas, with url as props
- Built feedbackmodal directly into ConfirmModal instead of the 3 separate campaign triggers, so that the modal logic only needs to be written once
- Control the feedbackmodal behavior with props from confirm modal to reduce dependancy.

### DIFF
--- a/frontend/src/components/common/confirm-modal/ConfirmModal.tsx
+++ b/frontend/src/components/common/confirm-modal/ConfirmModal.tsx
@@ -8,6 +8,7 @@ import ConfirmImage from 'assets/img/confirm-modal.svg'
 
 import { PrimaryButton, TextButton, ErrorBlock } from 'components/common'
 import FeedbackModal from 'components/common/feedback-modal'
+import { AuthContext } from 'contexts/auth.context'
 import { ModalContext } from 'contexts/modal.context'
 
 const ConfirmModal = ({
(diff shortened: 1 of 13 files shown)


frontend/src/components/common/confirm-modal/ConfirmModal.tsx
```diff
@@ -29,15 +33,34 @@ const ConfirmModal = ({
...
+  needFeedback?: boolean
+  feedbackUrl?: string
 }) => {
   const modalContext = useContext(ModalContext)
+  const { email } = useContext(AuthContext)
   const [errorMessage, setErrorMessage] = useState('')
 
+  // feedback modal is built directly into confirm modal as confirm modal is the component that is commonly used across the campaigns
+  const openFeedbackModal = () => {
+    if (feedbackUrl !== null && needFeedback) {
```
stanleynguyen, line 45: Can we use a single arg instead? Cos if `feedbackUrl` is present, it also means we want feedback right

stanleynguyen (changes requested):

frontend/src/components/common/confirm-modal/ConfirmModal.tsx
richard-yak: good point, have changed to use only 1 arg!

stanleynguyen (approved):

PR #1859, opened 2022-11-02
feat: list email messages endpoint

tan-zx:
[Ticket](https://www.notion.so/opengov/Implement-list-transactional-messages-endpoint-610554e007ef4e84ad23bd1a30ceb9a1)

The general strategy I have adopted is to:
- Apply validation at the routes level using `celebrate`
- Convert the validated query params into type-safe arguments in the middleware layer
- Pass these arguments to `EmailTransactionalService`, which uses sequelize to retrieve the requested data

**Implementation Notes:**
- For extensibility, instead of passing in `created_at` directly, I have transformed it into a `TimestampFilter` that could be extended to filter by other timestamps
- I considered whether to build `sortBy` in such a way that allow for concurrent sorts in the future, but seeing as that is not in spec and it's unclear whether our API users need it (compared to more filters, which would almost certainly be needed), I did not build it that way. 
- To ensure DRY-ness as much as possible, I have passed in the `TransactionalEmailSortField` enum as part of the Joi regex validation of `sort_by`. I considered whether it was possible to do the same for the `created_at` validation using `TimestampFilter` and `TimestampOperator`, but wasn't sure how to do it. 

**Misc:**
-  Renamed `SortField` to `CampaignSortField` as that was only used in listing campaigns. For transactional emails, we are sorting by a different list of enums.
- The first commit in this PR is to fix mistakes in the Swagger docs that was caused by hasty copy-and-paste
- I have extended `convertMessageModelToResponse` to include `created_at` and `updated_at` since we're allowing sorting and filtering by these fields, would only make sense to expose this information to the API user too

**[Known downsides](https://www.moesif.com/blog/technical/api-design/REST-API-Design-Filtering-Sorting-and-Pagination/#downsides-3)** of this implementation:
- Not performant for large offset values
- Not consistent when new items are inserted to the table (i.e. Page drift). This is especially noticeable when we are ordering items by newest first. 

Regarding the large offset values, citing the same link, this is OK "in applications where the data set has a small upper bounds". ~~Considering this table can scale to quite large, I think we should probably note something in the API documentation or maybe set a max offset value.~~ Discussed with @stanleynguyen and, at our scale, there is no need to set a max offset value. We can deal with this in the future in the (happy) event that we get to a much larger scale.

### DIFF
--- a/backend/src/core/constants.ts
+++ b/backend/src/core/constants.ts
@@ -46,19 +46,17 @@ export enum Ordering {
 export enum TransactionalEmailSortField {
   Created = 'created_at',
   Updated = 'updated_at',
-  // can extend if we want to sort by other fields (to remove and write in PR)
 }
 
 export interface TimestampFilter {
-  createdAt: DateOperator
-  // can extend if we filter by other timestamp (to remove and write in PR)s
+  createdAt: TimestampOperator
 }
 
-export interface DateOperator {
-  gt: Date
-  gte: Date
-  lt: Date
-  lte: Date
+export interface TimestampOperator {
+  gt?: Date
+  gte?: Date
+  lt?: Date
+  lte?: Date
 }
 /**
  * @swagger
--- a/backend/src/core/constants.ts
+++ b/backend/src/core/constants.ts
@@ -33,7 +33,7 @@ export enum DefaultCredentialName {
   Telegram = 'Postman_Telegram_Demo',
 }
 
-export enum SortField {
+export enum CampaignSortField {
   Created = 'created_at',
   Sent = 'sent_at',
 }
@@ -43,6 +43,23 @@ export enum Ordering {
   DESC = 'DESC',
 }
 
+export enum TransactionalEmailSortField {
+  Created = 'created_at',
+  Updated = 'updated_at',
+  // can extend if we want to sort by other fields (to remove and write in PR)
+}
+
+export interface TimestampFilter {
+  createdAt: DateOperator
+  // can extend if we filter by other timestamp (to remove and write in PR)s
+}
+
+export interface DateOperator {
+  gt: Date
+  gte: Date
+  lt: Date
+  lte: Date
+}
 /**
  * @swagger
  * components:
--- a/backend/src/core/constants.ts
+++ b/backend/src/core/constants.ts
@@ -49,14 +49,14 @@ export enum TransactionalEmailSortField {
 }
 
 export interface TimestampFilter {
-  createdAt: TimestampOperator
+  createdAt: ComparisonOperator<Date>
 }
 
-export interface TimestampOperator {
-  gt?: Date
-  gte?: Date
-  lt?: Date
-  lte?: Date
+export interface ComparisonOperator<T> {
+  gt?: T
+  gte?: T
+  lt?: T
+  lte?: T
 }
 /**
  * @swagger
--- a/backend/src/email/middlewares/email-transactional.middleware.ts
+++ b/backend/src/email/middlewares/email-transactional.middleware.ts
@@ -18,12 +18,18 @@ import {
 } from '@email/models'
 
 import crypto from 'crypto'
+import {
+  Ordering,
+  TimestampFilter,
+  TransactionalEmailSortField,
+} from '@core/constants'
 
 export interface EmailTransactionalMiddleware {
   saveMessage: Handler
   sendMessage: Handler
   rateLimit: Handler
   getById: Handler
+  listMessages: Handler
 }
 
 export const RATE_LIMIT_ERROR_MESSAGE =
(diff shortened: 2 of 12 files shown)


backend/src/core/constants.ts
```diff
@@ -43,6 +43,21 @@ export enum Ordering {
...
+export interface TimestampFilter {
+  createdAt: TimestampOperator
+}
+
+export interface TimestampOperator {
+  gt?: Date
+  gte?: Date
+  lt?: Date
+  lte?: Date
+}
```
stanleynguyen, line 60:
```suggestion
export interface TimestampFilter {
  createdAt: ComparisonOperator<Date>
}

export interface ComparisonOperator<T> {
  gt?: T
  gte?: T
  lt?: T
  lte?: T
}
```

stanleynguyen: This is not only applicable for timestamp, use generics

tan-zx: good catch ser
image

tan-zx:
I updated the PR description with a known downside section and have some suggestions, lmk what you think

I will write the tests in a separate PR during the next sprint (rather than work on this over the weekend)

can merge in along with the tests, in case while writing tests we find any other problems with this implementation. wdyt?

tan-zx:
adding @richard-yak to review this PR — the tests are still WIP but the feature part is done

I do have one question though, i.e. whether we should add a max `offset` value. Repeating the question I asked in the PR description here:

> **[Known downsides](https://www.moesif.com/blog/technical/api-design/REST-API-Design-Filtering-Sorting-and-Pagination/#downsides-3)**:
> - Not performant for large offset values
> - Not consistent when new items are inserted to the table (i.e. Page drift). This is especially noticeable when we are ordering items by newest first. 
> 
> From that link, this is OK "in applications where the data set has a small upper bounds". Considering this table can scale to quite large, I think we should probably note something in the API documentation or maybe set a max offset value. What do you think? @stanleynguyen

stanleynguyen:
> adding @richard-yak to review this PR — the tests are still WIP but the feature part is done
> 
> I do have one question though, i.e. whether we should add a max `offset` value. Repeating the question I asked in the PR description here:
> 
> > **[Known downsides](https://www.moesif.com/blog/technical/api-design/REST-API-Design-Filtering-Sorting-and-Pagination/#downsides-3)**:
> > 
> > * Not performant for large offset values
> > * Not consistent when new items are inserted to the table (i.e. Page drift). This is especially noticeable when we are ordering items by newest first.
> > 
> > From that link, this is OK "in applications where the data set has a small upper bounds". Considering this table can scale to quite large, I think we should probably note something in the API documentation or maybe set a max offset value. What do you think? @stanleynguyen

Nope, max `offset` doesn't make any sense at all. Those performance downsides are negligible for the application that we're working with

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -722,7 +724,306 @@ describe('POST /transactional/email/send', () => {
...
+    expect(res.body.data.length).toBe(5)
+
+    const res2 = await request(app)
+      .get(
+        `${endpoint}?created_at[gt]=${messages[0].createdAt.toISOString()}&created_at[lt]=${messages[4].createdAt.toISOString()}`
+      )
+      .set('Authorization', `Bearer ${apiKey}`)
+    expect(res2.status).toBe(200)
+    expect(res2.body.has_more).toBe(false)
+    expect(res2.body.data.length).toBe(4) // not sure why this is 4 and not 3
```
tan-zx, line 969: honestly don't know why this is 4 and not 3 (sorry @stanleynguyen I know you said I should remove comments from the code, but I needed to include this to mark out the weird statement from the sea of tests... will remove once this is resolved)

tan-zx:
Ready for review

backend/src/email/routes/tests/email-transactional.routes.test.ts
tan-zx: oops think it's a bug let me fix

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -212,6 +221,31 @@ export const InitEmailTransactionalMiddleware = (
...
+      ? Ordering.ASC
+      : Ordering.DESC // default to descending order even without '-' prefix
+    const { hasMore, messages } = await EmailTransactionalService.listMessages({
+      userId,
+      limit: +(limit as string),
+      offset: +(offset as string),
+      sortBy: sortBy as TransactionalEmailSortField,
+      orderBy,
+      status: status as TransactionalEmailMessageStatus,
+      filterByTimestamp: filter as unknown as TimestampFilter,
```
richard-yak, line 241: is this as unknown necessary? tried it on my IDE it works without as unkown.

richard-yak (commented):
core logics look good to me! but just clarify if need to clean up as unknown, just wanted to avoid hacks if possible. i know on top uses as unknown as well, will investigate to see if can remove that as well

backend/src/email/middlewares/email-transactional.middleware.ts
tan-zx: ohh weird you're right it works now (I'm almost sure I added it because I was prompted by my IDE).

good catch, have removed it.

richard-yak (approved):
LGTM

stanleynguyen (approved):

PR #1860, opened 2022-11-02
refactor: campaign service, middleware, and routes + tests

tan-zx:
Trigger: notice code here is a bit hard to understand while working on [this](https://github.com/opengovsg/postmangovsg/pull/1859)

Changes:
- Refactored to use IIFE so dependencies among the different variables are clearer
- Added additional tests

### DIFF
--- a/backend/src/core/middlewares/campaign.middleware.ts
+++ b/backend/src/core/middlewares/campaign.middleware.ts
@@ -4,7 +4,7 @@ import {
   CampaignSortField,
   ChannelType,
   Ordering,
-  Status,
+  CampaignStatus,
 } from '@core/constants'
 import { CampaignService, UploadService } from '@core/services'
 import { Campaign } from '@core/models'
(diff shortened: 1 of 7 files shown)


tan-zx:
Updated @jhibberd-gov

backend/src/core/middlewares/campaign.middleware.ts
```diff
@@ -130,7 +130,7 @@ const createCampaign = async (
 const listCampaigns = async (
   req: Request,
   res: Response,
-  next: NextFunction
+  next: NextFunction // todo: remove the error handling?
```
stanleynguyen, line 133: Yea we probably don't need `next(err)` cos if `err` thrown inside the handler that's the default behavior I think

stanleynguyen (approved):
Should we just remove the `next(err)` thing? Can test out the behavior by intentionally throwing an error inside and compare the behavior with vs without the `next(err)`

PR #1862, opened 2022-11-03
tempfix(backend): return 202 for legacy transactional users

stanleynguyen:
## Problem

Some of our users are relying on 202 specifically

## Solution

Return 202 for legacy users

## Deployment Checklist

- [ ] Add `5378` to the comma-separated list in prod

### DIFF
--- a/backend/src/core/config.ts
+++ b/backend/src/core/config.ts
@@ -152,6 +152,7 @@ interface ConfigSchema {
     maxAttachmentSize: number
     maxAttachmentNum: number
   }
+  legacyTransactional202Users: string
 }
 
 convict.addFormats({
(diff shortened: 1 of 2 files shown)


tan-zx:
Testing on staging

I think we should update the [description here](https://github.com/opengovsg/postmangovsg/blob/dfe6b2db923384ea6ff59d5c440b014bf04a04ba/backend/src/email/routes/email-transactional.routes.ts#L99-L104), which still refers to Accepted (which is status code 202).

Instead, we should make clear in the docs (assuming this is where our users will go to get info about our API):
- What is being created
- That this doesn't imply the email has been sent successfully

tan-zx (approved):
image

tested works on staging; added env var to prod

PR #1863, opened 2022-11-07
feat(backend): rendering hyperlinks

richard-yak:
## Problem

Hyperlinks in E-mails were not rendering as clickable hyperlinks.

## Solution

This solution will read the column header of the templated field, and if the field ends with "_url" or "_link", will automatically wrap the field with a tags so that the HTML on the E-mail can render as hyperlinks

## To Note
In future as we have more "template rules", can easily refactor here to check based on column strings and handle accordingly using switch cases or a new function.


stanleynguyen:
Can we


It's worth adding a test case to `tests/templating/template-client.test.ts`

stanleynguyen (changes requested):
Can we


It's worth adding a test case to `tests/templating/template-client.test.ts`

richard-yak:
@stanleynguyen done adding test cases, plz re-review

stanleynguyen (approved):

PR #1864, opened 2022-11-07
fix: use logger.warn instead of logger.error for unhandled events

tan-zx:
Use `warn` so it won't clutter up our error logs

We choose to do this (instead of simply disabling these events on AWS) because these events could be helpful if we ever want to get more info that is not shown in our database. (@stanleynguyen is this correct?)


richard-yak (approved):
definitely no issues code-wise

stanleynguyen (approved):

PR #1865, opened 2022-11-08
fix(frontend): fix auto-scrolling of protected page

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Bug-on-password-protect-page-491d1060ac2e475f8aa5664a70eac23f)

## Solution

- Setting font of input field to `16px` explicitly which is the same as the default value (i.e. not affecting anything) to prevent iOS from zooming in
- Scroll to top after protected content is loaded
- Reset zoom to 1 after protected content is loaded

## Deployment Checklist

- N/A


stanleynguyen:
https://user-images.githubusercontent.com/12974927/200490316-bde27bfc-14bb-4395-9f69-3e4f1036a465.MP4

Tested in staging courtesy of @richard-yak

richard-yak (approved):
a-ok

PR #1867, opened 2022-11-08
fix: logo rendering in footer on SGMail

tan-zx:
[Ticket](https://www.notion.so/opengov/Postman-footer-not-rendering-properly-on-SGMail-6bb4b499f2e34f67b954e614e63b1682)

## Issue

- Fundamentally, Microsoft Outlook renders email HTML differently, so hacky workarounds in our mustache template are needed. 
- Proximately, this issue was introduced by [this PR](https://github.com/opengovsg/postmangovsg/pull/1743), specifically this change:
image

## Solution

To see the HTML of the email, forward the email to Gmail, click "Show original", change `view=om` to `view=lg`. Now you can save it as a `.html`. [Source](https://webapps.stackexchange.com/a/117154)

Changes made:
- The first change is to add a missing closing `</div>` tag. This was picked up by my IDE though ostensibly it didn't make any difference.
- The second change is to add `height="24"` to the OGP logo `` tag. While duplicative of what's alr in `style`, this is necessary in order for the logo to render at a correct size. Without this, the logo is 4x the size of what it should be. (See screenshots below).
- The third change is to fix the logos' relative position, which are vertically instead of horizontally side-by-side. To do so, I tried to follow the style of the existing `mustache` file and not make significant changes (since I'm not very familiar with Mustache). I tested every change made to make sure that it is absolutely necessary. At a high level, the idea is to use `table` instead of `div` for Outlook to render the email properly.
- The fourth change is to add `&nbsp;` into the `div` that was previously being ignored by Outlook. [Source](https://stackoverflow.com/a/71293011).

## Screenshots

Before:
image

After:
image

### DIFF
--- a/shared/src/theme/email-theme.mustache
+++ b/shared/src/theme/email-theme.mustache
@@ -213,11 +213,13 @@
           <br />    
           <a href="{{{unsubLink}}}">Unsubscribe</a>
           <div class="footer-vertical-spacer-top" style="line-height: 16px; height: 16px;">&#8202;</div>
-          <div>
-            <img src="https://file.go.gov.sg/postman-logo-color.png" alt="Postman.gov.sg Logo" height="24" style="height: 24px; width: 109.55px;">
-            <div style="margin: 0 10px; border-left: 1px solid #B5C4FF; height: 24px; display: inline-block;width: 0;"></div>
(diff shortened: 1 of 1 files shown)


richard-yak (approved):
looks like a low risk css change

PR #1868, opened 2022-11-09
refactor(backend): revise authentication for endpoints

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Revise-cookie-API-authentication-1fcff2fca43d425a9e735350f5734e0e)

## Solution

- Make auth middleware modular
- Restrict non-API-exposed endpoints to be cookie-only
- Remove non-API-exposed from our published swagger doc

## Deployment Checklist

- N/A

### DIFF
--- a/backend/src/core/middlewares/auth.middleware.ts
+++ b/backend/src/core/middlewares/auth.middleware.ts
@@ -8,8 +8,13 @@ export interface AuthMiddleware {
   getOtp: Handler
   verifyOtp: Handler
   getUser: Handler
-  isCookieOrApiKeyAuthenticated: Handler
+  getAuthMiddleware: (authTypes: AuthType[]) => Handler
   logout: Handler
+}
+
+export enum AuthType {
+  Cookie = 'COOKIE',
+  ApiKey = 'API_KEY',
 }
 
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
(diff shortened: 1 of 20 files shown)


backend/src/core/middlewares/auth.middleware.ts
```diff
@@ -110,48 +115,70 @@ export const InitAuthMiddleware = (authService: AuthService) => {
...
-    try {
-      if (authService.checkCookie(req)) {
-        logger.info({
-          message: 'User authenticated by cookie',
-          action: 'isCookieOrApiKeyAuthenticated',
-          user: req.session?.user,
-        })
-        return next()
+  type AuthenticatorType = (req: Request) => Promise<boolean>
+  const authenticators: Record<AuthType, AuthenticatorType> = {
```
richard-yak, line 119: wow this is some insane use of generics to do what i would've written as a basic switch case to handle. nice.

richard-yak (approved):
i've learnt

tan-zx:
oh btw since you're making these changes, could you also add in the `security: bearerAuth[]` as shown below?

image

it'll help make the Swagger docs look nicer by clearly distinguishing public routes and API-authenticated routes with the little lock sign as shown below

image

richard-yak (approved):
2nd review np

PR #1870, opened 2022-11-16
test(cypress): add test case for txnal emails

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Add-API-calls-in-Cypress-test-12723e30029e4c8ba79bd22437d073df)

## Solution

Implement it

## Deployment Checklist

- [ ] N/A


tan-zx:
haha you should write up your experience debugging this thing :P

richard-yak (approved):
LGTM tbh

tan-zx:
@stanleynguyen if you don't mind, I'm gonna try to clean this up a bit, e.g.
- move the cypress credentials and stuff from the root folder to the cypress folder
- look into sharing the login states across the frontend tests (so we don't need to login once for every test)

can be part of this PR or another PR; think this would be a good way for me to become more familiar with cypress

stanleynguyen:
> @stanleynguyen if you don't mind, I'm gonna try to clean this up a bit, e.g.
> 
> * move the cypress credentials and stuff from the root folder to the cypress folder
> * look into sharing the login states across the frontend tests (so we don't need to login once for every test)
> 
> can be part of this PR or another PR; think this would be a good way for me to become more familiar with cypress

Actually cypress is quite 💩  (Almost died working on this PR). Let's propose to move to another framework?

tan-zx:
Let's use Playwright hehe

PR #1871, opened 2022-11-17
Richardyak/refactor/common models

stanleynguyen:
For giving comments to the refactor attempt


backend/jest.config.js
```diff
@@ -5,6 +5,8 @@ module.exports = {
   testPathIgnorePatterns: ['<rootDir>/build/', '<rootDir>/node_modules/'],
   moduleNameMapper: {
     '@core/(.*)': '<rootDir>/src/core/$1',
+    '@models/(.*)': '<rootDir>/../shared/src/core/models/$1',
+    '@interfaces/(.*)': '<rootDir>/../shared/src/core/interfaces/$1',
```
stanleynguyen, line 9: We don't need these as we have and are using `@shared`

backend/src/core/loaders/sequelize.loader.ts
```diff
@@ -3,7 +3,7 @@ import { Sequelize, SequelizeOptions } from 'sequelize-typescript'
 import config from '@core/config'
 import { initializeModels } from '@core/models'
 
-import { loggerWithLabel } from '@core/logger'
+import { loggerWithLabel } from '@shared/core/logger'
```
stanleynguyen, line 6: I rmb intentionally left this out of the shared folder when I was doing some refactoring previously cos they loggers are not the same

stanleynguyen: Could you double check?

shared/src/core/models/sms/sms-message.ts
```diff
@@ -6,8 +6,8 @@ import {
   Table,
   BelongsTo,
 } from 'sequelize-typescript'
-import { Campaign } from '@core/models/campaign'
-import { MessageStatus } from '@core/constants'
+import { Campaign } from '@shared/core/models/campaign'
+import { MessageStatus } from '../../constants'
```
stanleynguyen, line 10: Can we alias `@core` in the `shared module too?

shared/src/core/models/user/tests/user.test.ts
```diff
@@ -0,0 +1,151 @@
+// import { Sequelize } from 'sequelize-typescript'
```
stanleynguyen, line 1: Why are we commenting out everything for this test file

stanleynguyen:
Will be part of the database refactor epic

PR #1872, opened 2022-11-17
refactor: cypress

tan-zx:
WIP


tan-zx:
to migrate away from Cypress

PR #1877, opened 2022-11-26
build: update package-lock.json

tan-zx:
Changes when I run `npm i` and `npm ci` on my laptop from the root folder
- I got rid of the frontend's `package-lock.json` changes in order for the CI/CD to pass.
- I think the behavior is different depending on whether I run `npm i` from the root or from the specific folder (e.g. backend) — but I think these changes should be correct?


richard-yak (approved):
changes seem small enough

PR #1879, opened 2022-11-29
feat(frontend): scheduled campaign detail page

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/New-Scheduled-state-info-page-and-relevant-campaign-components-954b060e5d7a4568b9fa84ea4bee3e43)

## Solution

Implement it duh

## Deployment Checklist

- [ ] N/A


richard-yak (approved):
let's gooo

PR #1880, opened 2022-11-29
refactor: drop unnecessary message_id columns

tan-zx:
## Problem

- `message_id` columns unused and unnecessary
- current SES event helper methods are very hard to read

## Solution

- remove `message_id` columns and stop existing code from writing to it
- refactor

## Deployment Checklist

- [ ] Deploy new backend
- [ ] Then run db migration


stanleynguyen (dismissed):

backend/src/database/migrations/20221129065356-drop-email-message-tx-message-id-column.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, _) => {
+    await queryInterface.removeColumn(
+      'email_messages_transactional',
+      'message_id'
+    )
+    // removing this column will lose data irreversibly
```
stanleynguyen, line 9: Are we storing data in this column? If so, we need to remove the code that's setting this column's value too

stanleynguyen: Btw next time for comment like this, can u make a github comment instead of a code comment? This comment doesn't provide any extra value for the codebase

tan-zx: yesser I thought I was just following this 😛

image

stanleynguyen: > > Are we storing data in this column? If so, we need to remove the code that's setting this column's value too
> 
> Ya I was just confused why we're storing data in this column if we don't need the data... will poke around and investigate this later (don't know whether we should put this in the sprint tasks?)

Yea probably give yourself a point for this

backend/src/email/services/email.service.ts
```diff
@@ -154,17 +154,18 @@ const getCampaignMessage = async (
 const sendEmail = async (
   mail: MailToSend,
   opts?: SendEmailOpts
-): Promise<string | void> => {
+): Promise<boolean> => {
```
jhibberd-gov, line 157: It looked like all the callsites for `sendEmail` were only interested in using the return value (i.e. the SES message ID) as an indicator for whether the function succeeded or not. So I felt it was clearer just to return a `boolean` to indicate success. This more clearly communicates that we don't actually use the SES message ID.

shared/src/clients/mail-client.class/index.ts
```diff
@@ -51,11 +48,11 @@ export default class MailClient {
         },
         ...(option?.extraSmtpHeaders || {}), // guard against undefined extraSmtpHeaders value
       } as { [key: string]: any }
-      if (input.referenceId !== undefined) {
```
jhibberd-gov, line 54: `referenceId` felt like another potentially confusing name for a variable that holds the message ID. I felt it was cleaner to simply use the term `messageId` now that this term will always refer to the table primary key value and not the SES ID -- every channel (i.e. SMS, Telegram) seem to use the term `messageId` to refer to the canonical, unique identifier for a message, so this feels consist.

backend/src/email/utils/callback/update-status.ts
```diff
@@ -39,7 +39,7 @@ export const updateBouncedStatus = async (
     errorCode,
     errorSubType,
     timestamp: metadata.timestamp,
-    id: metadata.id,
```
jhibberd-gov, line 42: Similar to [this comment](https://github.com/opengovsg/postmangovsg/pull/1880/commits/d49222a7cb867138799b06916391106311e6e7d8#r1109204869), I think it's generally best to avoid using very generic variable names (like `id`) unless the context is super clear (i.e. "the ID of what?"). In most cases for this diff I felt it was clearer to qualify the variable name as `messageId`.

tan-zx:
Actually, looking at this latest email with AWS support team, I wonder if it might be a good idea to instead save all the `messageId` from SES, in case it might be helpful to such correspondence with AWS support team. 

image

jhibberd-gov:
> Actually, looking at this latest email with AWS support team, I wonder if it might be a good idea to instead save all the `messageId` from SES, in case it might be helpful to such correspondence with AWS support team.
> 
> image

A few thoughts:

- To keep work manageable I think it's important to limit PRs to a single concept. To me, this PR seems to be about removing the SES ID from the *mechanics* of sending an email (e.g. including it in headers for future reference, including it in operational tables). 
- I think the suggestion to log SES ID *somewhere* for diagnostics makes sense, but I feel this should be part of a second PR, where the concept is making email troubleshooting easier (and doesn't affect the mechanics of sending emails).
- Perhaps we could simply log the `messageId` (primary key) and its corresponding SES ID in the `sendEmail` function in `email.service.ts`?

tan-zx:
Oops, sorry I think my original message is ambiguous!

So I'm thinking maybe we should do the complete opposite of this PR, where we save the SES `messageId` in the database (not just logging it). That way, we have a persistent 1-to-1 association between our messages and what we sent to AWS SES and the `messageId` could help with possible debugging/support requests. (The reason we hadn't thought of this previously is because, prior to recently upgrading our support plan with AWS SES, our correspondence with the AWS support team has been almost completely useless...)

Currently, the `messageId` column in the `email_transactional_messages` is completely empty, so we would need to make some code changes.

Happy to discuss further @jhibberd-gov @stanleynguyen

richard-yak:
leaving a comment not directly related to the PR but something that popped up while I was doing the Transactional SMS Callback handling feature: I had to use message_id column for SMS transactional message tables because that was the only way I could uniquely identify my records (there was no way of using PK as they were not created before being sent, and creating them in advance would result in double read-write retrieval overhead + risk of drop offs). 

Not sure if transactional e-mail would face the same problems if we tried to handle delayed callbacks from SES but just dropping a possible use-case here! 

P.S. agree with @tan-zx latest suggestion of actually saving message_id into the column because i think it would definitely come in handy eventually. Can perhaps add index to message_id too while we are at it :>

tan-zx:
To-dos after discussion with @jhibberd-gov :
- [ ] Still remove the `message_id` columns. Keep data schema 
- [ ] Ensure that Cloudwatch is set up properly
  - [ ] Persists logs for good amount of time (e.g. 6 months)
  - [ ] Ensure that logs contain both primary key ID and AWS SES's `message_id` and it's easy to search via Cloudwatch **[separate PR]**

shared/src/clients/mail-client.class/index.ts
tan-zx: **[change request, subjective]** given our latest discussion — should we use a different name for the table primary key id?
- the word `messageId` is quite strongly associated with AWS SES's ID (see [here](https://docs.aws.amazon.com/ses/latest/dg/send-email-concepts-process.html)) and I think is the source of confusion for when I was tracing through the codebase on my own
- we will be logging the AWS SES's ID as part of the logging (and it's returned by the `sendEmail` method), so we'll have both variable names in the codebase

I think sth verbose would be the most clear, e.g. `emailMessagesId` or `emailTransactionalMessagesId`

jhibberd-gov: - After this commit the name `messageId` -- within the context of email -- should only ever refer to the table primary key. The `sendEmail` message now returns a success/failure boolean rather than the SES ID because no callsites were using the SES ID.
- The other channels (e.g. SMS, Telegram) also appear to use the term `messageId` to refer to the primary identifier for a message.
- On the basis of the above two points I think it's okay to stick with `messageId`? For the one case where we want to log the SES ID I think we can use an appropriately qualified variable name (e.g. `sesMessageId`) but it will only ever be used in the local function scope (e.g. `sendEmail`), so I don't think we're at risk of it causing confusion with the `messageId` variable.

jhibberd-gov:
> To-dos after discussion with @jhibberd-gov :
> 
> * [ ]  Still remove the `message_id` columns. Rationale: keep data schema as simple as possible, to only contain what is needed for day-to-day operations **[todo in this PR]**
> 
> The subsequent checkboxes can be handled as a separate task + PR! [Ticket here](https://www.notion.so/opengov/fd466213f133486d85e4117f1ded521d?v=94cf83ccaa9c4fa4a15235cdf10ba9f4&p=03254019c59f4a6d92af62a11043c6fc&pm=s)
> 
> * [ ]  Ensure that Cloudwatch is set up properly, i.e. logs are persisted logs for good amount of time (e.g. 6 months)
> * [ ]  Ensure that logs contain both primary key ID (used for handling callbacks) and AWS SES's `message_id` and it's easy to search via Cloudwatch
> * [ ]  Document how to use Cloudwatch to search (for when diagnosis is necessary)

Is there any further work to be done to remove the `message_id` columns? Isn't that all covered by the existing commits in this PR?

tan-zx:
> Is there any further work to be done to remove the message_id columns? Isn't that all covered by the existing commits in this PR?

Yup, you're right, I meant in the context of this PR. Let me take some time to properly review this PR this couple of days!

shared/src/clients/mail-client.class/index.ts
tan-zx: as spoken, I still prefer referring to the primary key id directly because in all the other instances, the `messageId` always refers to the ID given by our downstream provider, so it's easy to get confused! nonetheless, am ok to sticking with this naming as long as it's commented properly in our code :)

jhibberd-gov (approved):
Tested in staging with lots of help from @tan-zx!

backend/src/email/services/email-transactional.service.ts
```diff
@@ -98,17 +98,18 @@ async function sendMessage({
...
+    messageId: emailMessageTransactionalId.toString(),
     attachments: sanitizedAttachments,
-    referenceId: emailMessageTransactionalId.toString(),
   }
   logger.info({
     message: 'Sending transactional email',
     action: 'sendMessage',
   })
-  const messageId = await EmailService.sendEmail(mailToSend, {
+  // receive from SES, but not saving to DB
```
tan-zx, line 108: we should remove this comment

backend/src/email/utils/callback/parsers/sendgrid.ts
```diff
@@ -67,9 +67,8 @@ const parseRecord = async (record: SendgridRecord): Promise<void> => {
     return
   }
   const metadata = {
-    id: record.message_id,
+    messageId: record.message_id,
```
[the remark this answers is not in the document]
tan-zx: (or alternatively remove this entire Sendgrid fallback...)

backend/src/email/utils/callback/parsers/ses.ts
```diff
@@ -131,14 +131,12 @@ const shouldBlacklist = ({
   return isHardBounced || isNegativelyComplained
 }
 
+// get rid of the anys
```
tan-zx, line 134: part of larger refactor of SES parser

PR #1881, opened 2022-11-29
feat: email body limit

tan-zx:
## Problem

[Ticket](https://www.notion.so/opengov/Implement-body-size-limit-to-Postman-transactional-email-d6bf865ffd9c49d1a096cec7b8bba4fb)

Currently, there is an inconsistency over what the size of email body our API will accept. 
- If the body is sent via JSON, there is a default 100KB limit. If this is exceeded, a 500 error "PayloadTooLarge" is thrown.
- If the body is sent via multipart-form, there is a default limit of 1MB. If this is exceeded, no error is thrown but the field will be truncated.

This inconsistent behaviour is unintended and should be rectified.

## Solution
1. First, an env var is used to determine what the size of the limit should be. This is set in number of bytes for precision.
2. Second, we want the validation to be done by Joi (which gives informative error message and should be the single source of truth for validation) and avoid relying on the default behavior of limits described above. To that end:
  - For the JSON, we have set the limit to be 10x of the env var's limit. This is arbitrarily chosen, but the idea is to avoid the uninformative 500 error, which doesn't tell you what the size is.
  - For the multi-part form, we have set the limit to be 1 byte more than the env var. This means any truncated field will fail the Joi validation test.
3. Having made the Joi validation more complex, we cannot rely on the default error message and thus wrote a custom one.
  - One downside of using Joi is that it returns the more generic 400 error rather than the more specific 413 error.
  - But I think that is fine as users should be able to figure out why they're getting the error

## Known Issues

There is a behavior I can't seem to reconcile when calling the API using a JSON payload whose `body` field exceeds the limit set in `app.use(express.json())` .
- When I call the API against my local server (and the same is true for staging and prod), I get a response code of 500 (though a `PayloadTooLargeError` message is logged) 
- However, when I write the tests, I get a 413 error instead of the 500 error

However, I think this is not a big deal and, as long as it's communicated properly to the user. 

## Deployment Checklist

N/A


backend/src/core/loaders/express.loader.ts
```diff
@@ -71,7 +71,13 @@ const expressApp = ({ app }: { app: express.Application }): void => {
...
 
-  app.use(express.json())
+  app.use(
+    express.json({
+      // this must be significantly bigger than transactionalEmail.bodySizeLimit so that users who exceed limit
+      // will get 404 error informing them of the size of the limit, instead of 500 error
+      limit: config.get('transactionalEmail.bodySizeLimit') * 10,
+    })
+  )
   app.use(express.urlencoded({ extended: false }))
```
stanleynguyen, line 81: Actually they can still use urlencoded format to send a request larger than `bodySizeLimit` actually

backend/src/core/loaders/express.loader.ts
```diff
@@ -71,7 +71,13 @@ const expressApp = ({ app }: { app: express.Application }): void => {
   // in the parseEvent() handle before parsing the SES event.
   app.use('/v1/callback/email', express.text({ type: 'application/json' }))
 
-  app.use(express.json())
+  app.use(
+    express.json({
+      // this must be significantly bigger than transactionalEmail.bodySizeLimit so that users who exceed limit
+      // will get 404 error informing them of the size of the limit, instead of 500 error
```
stanleynguyen, line 77: Why are we returning 404, that's a bit weird

stanleynguyen: If it's because we're leaving it to the default, what could probably work is to update the [error handler](https://github.com/opengovsg/postmangovsg/blob/aa7cf4f580b0e557ccd3f0c6ba9a27064eba769e/backend/src/core/loaders/express.loader.ts#L128-L141) following [this](https://stackoverflow.com/questions/58134287/catch-error-for-bad-json-format-thrown-by-express-json-middleware) to handle the error

stanleynguyen: This should return 400 with a message informing user about the size limit

tan-zx: oh wait I think the `404` might be a typo, I think it's really 400.

let me verify later ps

backend/src/core/loaders/express.loader.ts
tan-zx: ohh I didn't realise that — let me investigate this later and replicate on my API client

tan-zx: OK replicated and fixed.

backend/src/core/loaders/express.loader.ts
tan-zx: Yup it's 400, sorry typo

backend/src/test-utils/server.ts
```diff
@@ -15,7 +16,13 @@ const initialiseServer = (session?: boolean): express.Application => {
...
   sessionLoader({ app })
-  app.use(express.json())
+  app.use(
+    express.json({
+      // this must be bigger than transactionalEmail.bodySizeLimit so that users who exceed limit
+      // will get 404 error informing them of the size of the limit, instead of 500 error
+      limit: config.get('transactionalEmail.bodySizeLimit') * 10,
+    })
+  )
   app.use(express.urlencoded({ extended: false }))
```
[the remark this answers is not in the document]
tan-zx: oh ya you're right — I will add that and the test cases (not sure whether `supertest` can do URLencoded calls, will look into that)

tan-zx:
oh no I think my force-push removed the tests and stuff from the earlier commits let me fix that later...

backend/src/test-utils/server.ts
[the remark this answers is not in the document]
tan-zx: OK fixed (similar behavior with the JSON payload — the tests give 413 error, but the actual error the user gets is 500 — weird)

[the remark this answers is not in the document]
stanleynguyen: Wait where did the 413 come from? How come it's suddenly in the picture? We should be consistent with the status code that we're giving for this same error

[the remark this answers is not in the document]
tan-zx: Honestly idk where it comes from. 
- On my API client, I only get the 500 error. This is the actual error code for the user. 
- I thought it might be from parts of the codebase where we set the 413 error code, but I changed those (to, e.g. use 400), but that has no effect on this test's behavior

So I think it's OK to return a 500 error code in this case, because the user is far exceeding our stated limits and, if they keep doing this, I don't think there is any problem with catching this weird behavior via our alerts.

stanleynguyen (approved):
lgtm. can merge after our sync if good from product standpoint

tan-zx:
rebased to prepare for merging tmr

PR #1884, opened 2022-12-07
Feat: Scheduled Sending PR 

richard-yak:
## Problem

[Ticket](https://www.notion.so/opengov/Scheduled-Sending-46f8bff861f549198bb701134e50185e)

## Solution

- Allow users to schedule their campaigns by adding a `visible_at` column in `job_queue` and `campaigns` table, that the workers will pick up when their time to send reaches.
- Also allow users to edit and cancel scheduling after they are scheduled.
- This PR also includes the feature of sending an e-mail notification on successful sending of a scheduled e-mail. 

## Deployment Checklist

- [ ] Connect to Prod DB and run migrations scripts first
- [ ] After full successful deployment, run separate migration script to remove the old insert_job DB function


backend/src/core/services/job.service.ts
```diff
@@ -14,14 +14,40 @@ import { ListService } from '.'
...
+  scheduledTiming?: Date
 }): Promise<number | undefined> => {
+  if (!scheduledTiming) {
+    // if its not a scheduled campaign, just throw in current timing. will not affect main flow of logic
+    scheduledTiming = new Date()
+  }
+
+  //format scheduled Timing to a formatted string
+  // i think i needa offset 8 hrs man
+  const stringTiming =
```
stanleynguyen, line 30: cmiiw but u can just pass the `Date` object in directly without having to convert it to a string

backend/src/core/services/job.service.ts
```diff
@@ -14,14 +14,40 @@ import { ListService } from '.'
...
   rate: number
+  scheduledTiming?: Date
 }): Promise<number | undefined> => {
+  if (!scheduledTiming) {
+    // if its not a scheduled campaign, just throw in current timing. will not affect main flow of logic
+    scheduledTiming = new Date()
+  }
+
+  //format scheduled Timing to a formatted string
+  // i think i needa offset 8 hrs man
```
stanleynguyen, line 29: For comments that are for review, can u open a PR and then comment there instead of commenting directly in the code, so we can:
- Ensure that no such comment will get committed to the codebase
- Document our discussion on Github

backend/src/core/models/campaign.ts
```diff
@@ -104,6 +104,12 @@ export class Campaign extends Model<Campaign> {
   })
   demoMessageLimit!: number
 
+  @Column({
+    type: DataType.DATE,
+    allowNull: true,
+  })
+  visibleAt?: Date
```
stanleynguyen, line 111: Why are we double-saving here, this field can and should be inferred from corresponding job queue records

backend/src/database/migrations/20221116100713-update-stop-jobs.js
```diff
@@ -0,0 +1,31 @@
...
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.createFunction(
+      'stop_jobs',
+      [{ type: 'integer', name: 'selected_campaign_id' }],
+      'void',
+      'plpgsql',
+      `
+UPDATE job_queue SET status = 'STOPPED', updated_at = clock_timestamp(), visible_at = clock_timestamp() WHERE campaign_id = selected_campaign_id AND status <> 'LOGGED';
```
stanleynguyen, line 11: why do we need to update `visible_at` in `stop_job` again?

backend/src/database/migrations/20221130131614-create-notification-data-by-campaign-id-function.js
```diff
@@ -0,0 +1,44 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    queryInterface.createFunction(
+      'get_notification_data_by_campaign_id',
```
stanleynguyen, line 6: hmm why do we need to put this in a sub-routine? Can't we just write raw query inside the code?

frontend/src/components/dashboard/create/email/EmailDetail.tsx
```diff
@@ -94,14 +98,20 @@ const EmailDetail = () => {
           from={emailCampaign.from}
         />
         <div className="separator"></div>
-        {stats.status && (
+        {stats.status && stats.status !== Status.Scheduled && (
```
stanleynguyen, line 101: shouldn't we just use `campaign.status` here instead of forcing `status` to be returned in the get stats endpoint?

stanleynguyen (changes requested):
Let's sched a live review soon so we can discuss the technical decisions made

backend/src/database/migrations/20221117031825-update-insert-job.js
```diff
@@ -0,0 +1,54 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    // drop all existing instances of insert_job
+    await queryInterface.dropFunction('insert_job', [
```
stanleynguyen, line 6: dont need to drop here

richard-yak:
to be fixed and re-raised after adding in UIUX feedback as well

backend/src/core/services/job.service.ts
richard-yak: tried this before settling on this implementation, if pass as date object, need to manually offset 8 hours ( think its due to DB being localized as GMT+8, so db auto adds 8 hours to timestamps passed in). but if i pass in as a UTC string as cast it as timestamp, then DB's auto convert will add the GMT timing to it

backend/src/core/services/job.service.ts
richard-yak: removed

backend/src/database/migrations/20221116100713-update-stop-jobs.js
richard-yak: oh yes good point, removed.

backend/src/database/migrations/20221130131614-create-notification-data-by-campaign-id-function.js
richard-yak: this is a personal implementation choice, as raw queries poses severe security vulnerabilities and extremely high chances of error. unless i can use ORM, this is the next safest method. to be refactored once models are made available to worker

backend/src/database/migrations/20221117031825-update-insert-job.js
richard-yak: removed

frontend/src/components/dashboard/create/email/EmailDetail.tsx
richard-yak: fixed

richard-yak: status was previously returned in stats endpoint already, but have changed to refer to campaign status instead.

backend/src/core/services/job.service.ts
stanleynguyen: Eh no leh the Date object should contain timezone information, can u show us this issue in our pair review?

richard-yak: wait, im trying it rn and it seems to be working now. i was pressure sure that the offset was screwing me over back when i tried it

backend/src/database/migrations/20221117031825-update-insert-job.js
```diff
@@ -0,0 +1,35 @@
...
+`,
+      [],
+      {
+        force: true,
+      }
+    )
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.dropFunction('insert_job', [
```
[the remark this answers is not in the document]
tan-zx: Actually, more generally, why don't we use a single migration instead of splitting them across four different files? Since I presume they're all or nothing and are not meant to be incrementally deployed?

backend/src/core/middlewares/job.middleware.ts
```diff
@@ -19,15 +19,35 @@ const sendCampaign = async (
...
-    const { rate } = req.body
+    // also need to retrieve if its to be scheduled
+    const { rate, scheduledTiming } = req.body
     const logMeta = { action: 'sendCampaign', campaignId }
 
+    // convert scheduled timing to date object
+    // if passed in then convert, if not just pass empty / undefined
+    const formattedTiming = scheduledTiming
+      ? new Date(scheduledTiming)
+      : undefined
```
tan-zx, line 30: `null` is a better choice for a value that's intentionally missing (`undefined` is for unintentionally missing values)

backend/src/core/services/job.service.ts
```diff
@@ -14,14 +14,25 @@ import { ListService } from '.'
...
   campaignId,
   rate,
+  scheduledTiming = undefined,
 }: {
   campaignId: number
   rate: number
+  scheduledTiming?: Date
 }): Promise<number | undefined> => {
+  if (!scheduledTiming) {
+    // if its not a scheduled campaign, just throw in current timing. will not affect main flow of logic
```
tan-zx, line 24: typo "it's"

backend/src/core/models/job-queue.ts
```diff
@@ -34,4 +34,11 @@ export class JobQueue extends Model<JobQueue> {
     allowNull: false,
   })
   status!: JobStatus
+
+  @Column({
+    type: DataType.DATE,
+    allowNull: false,
+    defaultValue: new Date(),
+  })
+  visibleAt?: Date
```
tan-zx, line 43: actually I'm a little confused, if `visibleAt` is not nullable and has a `defaultValue`, surely it should not be undefinable?

would be good to refactor the other columns in accordance to the PRs associated with [this issue](https://github.com/opengovsg/postmangovsg/issues/1696)

backend/src/core/services/job.service.ts
```diff
@@ -52,10 +63,14 @@ const sendCampaign = async ({
   campaignId,
   rate,
   userId,
+  scheduledTiming = undefined,
 }: {
   campaignId: number
   rate: number
   userId: number
+  // scheduled timing is intended to be a pass through variable. just pass and don't handle
```
tan-zx, line 71: very helpful comment!

backend/src/database/migrations/20221116075840-update-get-next-job.js
```diff
@@ -0,0 +1,94 @@
...
+      WHERE
+      -- worker is not already operating on a job
+      NOT EXISTS (SELECT 1 FROM job_queue q WHERE q.worker_id = worker LIMIT 1)
+      AND id = ( SELECT q.id
+          FROM job_queue q, campaigns p, credentials c
+          WHERE q.status = 'READY'
+          AND q.campaign_id = p.id
+          AND p.cred_name = c.name
+          AND is_credential_used(p.id) IS NULL
+          AND q.visible_at <= clock_timestamp()
```
tan-zx, line 22: this line is the key change right? would be good to have comments summarising the diff so it's easier to follow 😅

backend/src/sms/routes/sms-campaign.routes.ts
```diff
@@ -176,5 +262,66 @@ export const InitSmsCampaignRoute = (
...
+  /**
+   * @swagger
+   * paths:
+   *  /campaign/{campaignId}/email/cancel:
+   *    post:
+   *      security:
+   *        - bearerAuth: []
+   *      tags:
+   *        - Email
+   *      summary: Select the list of recipients from an existing managed list
```
tan-zx, line 274: summary looks wrong

cypress/e2e/email_scheduled.cy.js
```diff
@@ -0,0 +1,106 @@
...
+    //write csv test file
+    // check if not exist first
+    cy.task("readFileMaybe", CSV_FILENAME).then((res) => {
+      if (!res) {
+        const CSV_CONTENT = "recipient,name\n" + EMAIL + ",postman";
+        cy.writeFile("cypress/fixtures/".concat(CSV_FILENAME), CSV_CONTENT);
+      }
+    });
+
+    //log in via OTP
```
tan-zx, line 43: haha should have space after `//`

tan-zx: got a few more later with the same issue

cypress/e2e/sms.cy.js
```diff
@@ -1,5 +1,5 @@
-describe('SMS Test', () => {
-  it('initiate email campaign', () => {
+describe("SMS Test", () => {
+  it("initiate sms campaign", () => {
```
tan-zx, line 2: same comment as above about single quotes

worker/src/core/loaders/message-worker/index.ts
```diff
@@ -248,6 +256,111 @@ const enqueueAndSend = async (): Promise<void> => {
   }
 }
 
+// refactor to use actual raw queries
```
tan-zx, line 259: you mean refactor to use ORM instead of raw queries?

worker/tsconfig.json
```diff
@@ -34,11 +36,21 @@
     "moduleResolution": "node",
     "baseUrl": "./src",
     "paths": {
-      "@core/*": ["core/*"],
-      "@sms/*": ["sms/*"],
-      "@email/*": ["email/*"],
-      "@telegram/*": ["telegram/*"],
-      "@shared/*": ["../../shared/src/*"]
+      "@core/*": [
```
tan-zx, line 39: is there some ESLint change that resulted in these formatting changes?

tan-zx (commented):
Looked through the backend code briefly — will pay more attention to the worker + frontend code after our 1:1 session!

backend/src/core/middlewares/job.middleware.ts
richard-yak: changed

backend/src/core/services/job.service.ts
richard-yak: changed

cypress.config.js
```diff
@@ -32,26 +32,29 @@ module.exports = defineConfig({
...
               include_body: true,
-            }, // Maximum poll interval (in seconds). If reached, return null, indicating the completion of the task().
+            } // Maximum poll interval (in seconds). If reached, return null, indicating the completion of the task().
           );
           return email;
         },
         findDownloaded: (path) => {
           return fs.readdirSync(path);
         },
+        readFileMaybe: (filename) => {
```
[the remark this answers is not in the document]
richard-yak: create if not exist

cypress.config.js
```diff
@@ -1,27 +1,27 @@
-const { defineConfig } = require('cypress');
-const path = require('path');
-const gmail = require('gmail-tester');
-const fs = require('fs');
+const { defineConfig } = require("cypress");
+const path = require("path");
+const gmail = require("gmail-tester");
+const fs = require("fs");
```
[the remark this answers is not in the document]
richard-yak: fixed

cypress/e2e/email_scheduled.cy.js
```diff
@@ -0,0 +1,106 @@
...
+    );
+    const SUBJECT_NAME = "sub_".concat(DATETIME).concat(RANDOM_STRING);
+    const MSG_CONTENT = Cypress.env("MSG_CONTENT").concat(RANDOM_STRING);
+
+    const OTP_SUBJECT = Cypress.env("OTP_SUBJECT");
+    const EMAIL = Cypress.env("EMAIL");
+    const MAIL_SENDER = Cypress.env("MAIL_SENDER");
+    const WAIT_TIME = Cypress.env("WAIT_TIME");
+
+    //write csv test file
```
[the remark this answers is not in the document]
richard-yak: fixed

cypress/e2e/email_scheduled.cy.js
richard-yak: fixed

cypress/e2e/sms.cy.js
richard-yak: fixed

worker/src/core/loaders/message-worker/index.ts
richard-yak: modified comment to make it clearer

worker/tsconfig.json
richard-yak: fixed from code styles

backend/src/core/models/job-queue.ts
tan-zx:
```suggestion
  visibleAt: Date
```

backend/src/database/migrations/20221117031825-update-insert-job.js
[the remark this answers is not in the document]
tan-zx: Pros and cons of single migration.

Pros:
- Easy to see all the db changes needed for scheduled sending to work in a single file
- Easier to revert (just run once)

Cons:
- If a specific step in the migration goes wrong, harder to undo (need to comment out the other un-related down steps)

cypress.config.js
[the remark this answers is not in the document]
tan-zx: Discussed; this is so that `.csv` file for Cypress test will only be generated if it doesn't yet exist.

tan-zx:
I'll review the frontend and worker code in detail subsequently!

Some stuff we discussed:
- Should we try to communicate to user that scheduled sending is not a guarantee that sending occurs **at** the scheduled time, but rather **on or after** that time? How might this interact with the "delayed" sending UI?
- Should we strictly only show date and time in terms of SGT (and indicate accordingly), so as to avoid edge cases where the user is not in Singapore?
- Cypress test should send to `internal-testing` email rather than Richard's email

Can I also check what's the the status of the tests?
- The frontend tests that are failing now will be fixed?
- The end-to-end tests include scheduled sending and we'll be able to test the feature on staging before merging into master right?

Also, I think there is a third state (i.e. besides halted and "delayed") that we might need to consider — i.e. the error state Kenneth was talking about, where the campaign does not run to completion and the user has an UI to retry. I don't think that's halted, but I'm not 100% sure.

backend/src/email/routes/email-campaign.routes.ts
```diff
@@ -192,5 +193,67 @@ export const InitEmailCampaignRoute = (
     emailTemplateMiddleware.selectListHandler
   )
 
+  /**
+   * @swagger
```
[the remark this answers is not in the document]
richard-yak: done

backend/src/sms/routes/sms-campaign.routes.ts
```diff
@@ -176,5 +262,66 @@ export const InitSmsCampaignRoute = (
     SmsTemplateMiddleware.selectListHandler
   )
 
+  /**
```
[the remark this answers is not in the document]
richard-yak: done

backend/src/telegram/routes/telegram-campaign.routes.ts
```diff
@@ -172,5 +173,67 @@ export const InitTelegramCampaignMiddleware = (
...
+  /**
+   * @swagger
+   * paths:
+   *  /campaign/{campaignId}/email/cancel:
+   *    post:
+   *      security:
+   *        - bearerAuth: []
+   *      tags:
+   *        - Email
+   *      summary: Select the list of recipients from an existing managed list
```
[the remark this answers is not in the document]
richard-yak: done

backend/src/database/migrations/20221116075840-update-get-next-job.js
richard-yak: done

backend/src/core/models/job-queue.ts
richard-yak: done

backend/src/core/services/job.service.ts
richard-yak: fixed

backend/src/core/models/campaign.ts
richard-yak: fixed

cypress.config.js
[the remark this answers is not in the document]
stanleynguyen: Yo this function should just be name `doesFileExist`, the `readFileMaybe` name is a bit misleading, at first I thought it's gonna return the file content

backend/src/database/migrations/20221117031825-update-insert-job.js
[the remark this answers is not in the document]
stanleynguyen: @tan-zx @richard-yak just wanna give my 2 cent, usually I do 1 migration max per PR. For this one @tan-zx , I think part of the reason why we have multiple migration is because this is an aggregate of multiple PRs

stanleynguyen (changes requested):
Eh ps I just realised, why do we need 3 diff cancel endpoints for different types of campaign ah? I thought we can just do it with `campaignId` without its type?

richard-yak:
@stanleynguyen changed to use only 1 route from campaignsRoutes

cypress.config.js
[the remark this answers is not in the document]
richard-yak: fixed

backend/openapi.yaml
```diff
@@ -328,7 +328,65 @@ paths:
         '525':
           description: SSL handshake failed
         '526':
-          description: Invalid SSL certificateÏ
+          description: Invalid SSL certificate
+
+  /campaign/{campaignId}/cancel:
```
tan-zx, line 333: this should be `campaigns` now? dk whether you need to move the relative position of the doc

backend/src/core/middlewares/campaign.middleware.ts
```diff
@@ -28,6 +28,12 @@ const canEditCampaign = async (
       CampaignService.hasJobInProgress(+campaignId),
       UploadService.getCsvStatus(+campaignId),
     ])
+    logger.info({
+      message: 'Successfully created new campaign',
+      hasJob,
+      csvStatus,
+      action: 'createCampaign',
```
tan-zx, line 35: I think the `action` should be aligned with the method name, i.e. should be `canEditCampaign` rather than `createCampaign`.

think this is true for the other `logger.info`s too

backend/src/core/middlewares/campaign.middleware.ts
```diff
@@ -37,6 +43,30 @@ const canEditCampaign = async (
...
+): Promise<Response | void> => {
+  try {
+    const { campaignId } = req.params
+    const [sentJobs] = await Promise.all([
+      CampaignService.hasAlreadyBeenSent(+campaignId),
+    ])
+    logger.info({
+      message: 'Checking can send campaign',
+      sentJobs,
+      action: 'sendCampaign',
```
tan-zx, line 59: `canSendCampaign`

backend/src/core/middlewares/job.middleware.ts
```diff
@@ -98,8 +116,28 @@ const retryCampaign = async (
...
+const cancelScheduledCampaign = async (
+  req: Request,
+  res: Response,
+  next: NextFunction
+): Promise<Response | void> => {
+  try {
+    const { campaignId } = req.params
+    const logMeta = { campaignId, action: 'cancelScheduledCampaign' }
+    logger.info({
+      message: 'Cancel Scheduled Campaign...',
```
tan-zx, line 128: nit, but would suggest removing the ellipsis to keep the style consistent

backend/src/core/models/job-queue.ts
```diff
@@ -34,4 +34,11 @@ export class JobQueue extends Model<JobQueue> {
     allowNull: false,
   })
   status!: JobStatus
+
```
tan-zx, line 37: think the current column comment added in the migration should be here (easier to reference compared to the db)

backend/src/core/services/campaign.service.ts
```diff
@@ -188,9 +213,11 @@ const listCampaigns = ({
 
   if (status) {
     let operation: any = {}
```
tan-zx, line 215: might be out of scope for this PR, but do you think you can refactor `listCampaigns` by building on [the PR here](https://github.com/opengovsg/postmangovsg/pull/1860) — I think the PR there is mostly done, can just copy and paste and add your changes

the current `listCampaigns` is really quite bad code 😅

backend/src/core/middlewares/job.middleware.ts
```diff
@@ -19,15 +19,33 @@ const sendCampaign = async (
   try {
     const userId = req.session?.user?.id
     const { campaignId } = req.params
-    const { rate } = req.body
+    // also need to retrieve if its to be scheduled
+    const { rate, scheduledTiming } = req.body
     const logMeta = { action: 'sendCampaign', campaignId }
 
+    // convert scheduled timing to date object
+    // if passed in then convert, if not just pass empty / undefined
```
tan-zx, line 27: should fix comment here, is `null`

backend/src/core/services/job.service.ts
```diff
@@ -123,9 +141,36 @@ const retryCampaign = (campaignId: number): Promise<any> | undefined => {
   })
 }
 
+const updateScheduledCampaign = async (
```
tan-zx, line 144: would prefer naming this to make clear it is updating the `scheduledTime` of a campaign

backend/src/sms/routes/sms-campaign.routes.ts
```diff
@@ -139,10 +139,96 @@ export const InitSmsCampaignRoute = (
 
   router.get('/preview', smsMiddleware.previewFirstMessage)
 
+  /**
```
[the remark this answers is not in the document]
tan-zx: to delete since we're using a single endpoint now?

cypress.config.js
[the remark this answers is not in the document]
tan-zx: think you didn't commit the changes 😅

tan-zx (commented):
took another pass at the backend code, apologies!

backend/src/core/services/job.service.ts
tan-zx: e.g. `updateCampaignScheduledTime` or sth

worker/src/core/loaders/message-worker/index.ts
```diff
@@ -248,6 +256,111 @@ const enqueueAndSend = async (): Promise<void> => {
...
+        created_at: createdAt,
+        unsent_count: unsentCount,
+        error_count: errorCount,
+        sent_count: sentCount,
+        invalid_count: invalidCount,
+        notification_email: notificationEmail,
+        halted: halted,
+      } = jsonRes
+
+      let mail: MailToSend | void
```
tan-zx, line 291: this should be `null` instead of `void` (see https://www.tutorialsteacher.com/typescript/typescript-void)

actually, when will `mail` not be falsy? there should always be a mail sent? so the nested`if (mail)` statement can be removed?

worker/src/core/loaders/message-worker/index.ts
```diff
@@ -248,6 +256,111 @@ const enqueueAndSend = async (): Promise<void> => {
   }
 }
 
+// to refactor to use orm
+const sendFinalizedNotification = (campaignId: number): void => {
```
tan-zx, line 260: Is it OK to refactor the code here? 
- I don't think using `void` here is correct since we are actually using the result returned?
- The nested if statements are also quite hard to reason about

I would suggest doing sth like this:

```ts
const results = await connection.query(...)
const { id: ..., campaign_name... } =  get(result,...)
```

Then to get rid of the nested if statements:
- if `campaignId` doesn't exist, I'd think we should throw an error right? So we can use a guard clause to get rid of that?
- the final `if` statement is fine I guess haha

worker/src/core/loaders/message-worker/index.ts
```diff
@@ -158,6 +164,8 @@ const finalize = tracer.wrap(
             campaignId,
             action: 'finalize',
           })
+          // for each campaign id, send email confirmation
+          sendFinalizedNotification(campaignId)
```
tan-zx, line 168: I think here is where you should use `void`

worker/src/core/loaders/message-worker/index.ts
```diff
@@ -248,6 +256,111 @@ const enqueueAndSend = async (): Promise<void> => {
...
+          mail =
+            await NotificationService.generateHaltedCampaignNotificationEmail(
+              client,
+              notificationEmail,
+              campaignName
+            )
+        } else {
+          // for scheduled, visible_at must be after created_at
+          // normal flow, pull out visibleAt from job_queue table.
+
```
tan-zx, line 305: unnecessary extra line?

worker/src/core/services/notification.service.ts
```diff
@@ -0,0 +1,105 @@
...
+      showMasthead: true,
+    }),
+    subject,
+    ...(config.get('mailFrom') ? { replyTo: config.get('mailFrom') } : {}),
+  }
+}
+
+export const NotificationService = {
+  sendEmail,
+  // eventually to move to a theme folder for all relevant transactional maielrs
```
tan-zx, line 102: typo

worker/src/core/services/notification.service.ts
```diff
@@ -0,0 +1,105 @@
...
+    '<p>Postman.gov.sg</p>'
+  const body = client.template(templateBody as string, params)
+  return assembleNotificationMail(recipient, subject, body)
+}
+
+const generateHaltedCampaignNotificationEmail = async (
+  client: TemplateClient,
+  recipient: string,
+  campaignName: string
+): Promise<MailToSend | void> => {
```
tan-zx, line 66: why is there `| void` in the return signature? think `assembleNotificationMail`'s return signature is `Promise<MailToSend>`

worker/src/core/loaders/message-worker/index.ts
```diff
@@ -248,6 +256,111 @@ const enqueueAndSend = async (): Promise<void> => {
...
+                errorCount,
+                sentCount,
+                invalidCount
+              )
+          }
+        }
+        if (mail) {
+          // Send email using node mailer
+          // cannot use singleton mailclient like backend cuz worker will self destruct
+          // instantiate it only when needed
```
tan-zx, line 333: can add the other comment you mentioned, which is that this is only initialized here because not all workers will be sending mails

actually I'm not very sure about a logger worker's life cycle. if a logger worker sends multiple emails, will this cause `mailClient` to be unnecessarily initialized over and over again?

worker/src/core/loaders/message-worker/index.ts
```diff
@@ -248,6 +256,111 @@ const enqueueAndSend = async (): Promise<void> => {
...
+          const mailClient = new MailClient(
+            config.get('mailOptions'),
+            config.get('mailOptions.callbackHashSecret'),
+            config.get('mailFrom'),
+            config.get('mailConfigurationSet')
+          )
+          const isEmailSent = await NotificationService.sendEmail(
```
tan-zx, line 340: iinw if `isEmailSent` is null, something has gone wrong and we should throw an error? (I think the underlying method would already throw an error, but we should also do it here?)

worker/src/core/services/notification.service.ts
```diff
@@ -0,0 +1,105 @@
...
+const sendEmail = async (
+  mailClient: MailClient,
+  mail: MailToSend,
+  opts?: SendEmailOpts
+): Promise<string | void> => {
+  try {
+    return mailClient.sendMail(mail, opts)
+  } catch (e) {
+    logger.error({
+      message: 'Error while sending test email',
```
tan-zx, line 21: I think "test email" is a bit of a misnomer?

worker/src/core/services/notification.service.ts
```diff
@@ -0,0 +1,105 @@
...
+  client: TemplateClient,
+  recipient: string,
+  campaignName: string,
+  unsentCount: number,
+  errorCount: number,
+  sentCount: number,
+  invalidCount: number
+): Promise<MailToSend | void> => {
+  const subject =
+    'Your scheduled campaign "' + campaignName + '" was successfully sent out'
```
tan-zx, line 39:
```suggestion
    `Your scheduled campaign "${campaignName}" was successfully sent out`
```

tan-zx (commented):
took a look at the worker!

richard-yak:
@tan-zx sorry but I think there is a bit of a misunderstanding on what a PR review is supposed to be.

I strongly suggest to review based on what actually needs to be resolved before releasing (i.e. design patterns or code inefficiencies etc, things that you are absolutely sure needs to be done), but do be mindful of the fact that this PR has been open for 2 weeks, and the intended release date was today.

backend/src/core/middlewares/job.middleware.ts
```diff
@@ -19,15 +19,33 @@ const sendCampaign = async (
...
+    const formattedTiming = scheduledTiming ? new Date(scheduledTiming) : null
     if (await JobService.canSendCampaign(+campaignId)) {
-      const jobIds = await JobService.sendCampaign({
-        campaignId: +campaignId,
-        rate: +rate,
-        userId,
-      })
+      let jobIds
```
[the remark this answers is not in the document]
richard-yak: this was done this way to persist the original writing style of when before the change was made.

backend/src/core/middlewares/campaign.middleware.ts
richard-yak: fixed w next push

backend/src/core/middlewares/campaign.middleware.ts
richard-yak: fixed w next push

backend/src/core/middlewares/job.middleware.ts
richard-yak: removing which ellipsis?

backend/src/core/models/job-queue.ts
richard-yak: fixed w next push

backend/src/core/services/campaign.service.ts
richard-yak: suggest leaving out of this PR first

backend/src/core/middlewares/job.middleware.ts
richard-yak: fixed w next push

backend/src/core/services/job.service.ts
richard-yak: scheduledTime is a property of the campaign, so you are updating the campaign. Also see the rest of the function names in this file, it is to be consistent.

worker/src/core/loaders/message-worker/index.ts
richard-yak: this is keeping consistent with how it has been implemented elsewhere...

worker/src/core/loaders/message-worker/index.ts
richard-yak: removed campaignId if checker.

we are not using the return value of sendFinalizedNotification; it is a self contained function. we use function chaining so that we only execute the next portion of code after the the query has been completed and retrieve the results.

worker/src/core/loaders/message-worker/index.ts
richard-yak: this function does return void

worker/src/core/loaders/message-worker/index.ts
richard-yak: fixed w next push

worker/src/core/services/notification.service.ts
richard-yak: fixed w next push

worker/src/core/services/notification.service.ts
richard-yak: again, this is done to keep consistent w the rest of the codebase

worker/src/core/loaders/message-worker/index.ts
richard-yak: pls suggest the copy changes if the above comment is not clear enough about the behavior of mailClient

worker/src/core/loaders/message-worker/index.ts
richard-yak: there may not necessarily be an error even if the mail was not successfully sent. even if there was, the function sendEmail already catches the error. 

On the side once again, this archetype is done referencing the existing code that has been already been written and tested to be working.

worker/src/core/services/notification.service.ts
richard-yak: fixed w next push

worker/src/core/services/notification.service.ts
richard-yak: fixed w next push

backend/src/sms/routes/sms-campaign.routes.ts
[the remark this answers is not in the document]
richard-yak: removed

prestonlimlianjie:
Thanks for the PR @richard-yak! Three suggestions for the next time we work on a big feature like this (no need to make big changes for this one)
- It would be great to create an `epic` tag and break the entire feature down into small GitHub issues associated with the epic. This would really help make the review process a lot easier, since there would be smaller PRs to review.
- I would also recommend syncing the RFC with the GitHub issues so that it's easier to track the final engineering design for big features like these - i.e. the RFC write up in Notion would have a clear breakdown of the issues + PRs involved
- This last suggestion is up to the team and @stanleynguyen feel free to make the final call - I was wondering if we should have a 1 hour RFC discussion for future big features like these, so that all engineers can pre-read and then discuss it. We did this last time for a feature that @tan-zx worked on and it helped the entire team align on what the task should look like.

backend/src/database/migrations/20221117031825-update-insert-job.js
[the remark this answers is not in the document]
tan-zx: > @tan-zx @richard-yak just wanna give my 2 cent, usually I do 1 migration max per PR. For this one @tan-zx , I think part of the reason why we have multiple migration is because this is an aggregate of multiple PRs

Occurred to me that, if we have a single migration, we can wrap the different steps in a single transaction, so that if any part fails, the whole thing can automatically rollback 🤔

also, forgot to note: think in my chat with Richard ytd:
- we are keeping the old `insert_job` so that we can have backwards compatibility
- we will the new `insert_job` differently so that it's clearer which is meant to run at any one point

backend/src/core/middlewares/job.middleware.ts
tan-zx: oh the `...` haha just a nit

worker/src/core/loaders/message-worker/index.ts
tan-zx: sorry, to elabroate, a function can return `void` (i.e. it doesn't return anything), but `void` is not a valid type in TypeScript

> this is keeping consistent with how it has been implemented elsewhere...
ohh which parts? haha should flag so that we can refactor those away

worker/src/core/loaders/message-worker/index.ts
tan-zx: my apologies, I just realized I'm highlighting the wrong line here

I meant we shouldn't use `void connection.query(...)` — the return type of `void` is correct!

worker/src/core/loaders/message-worker/index.ts
tan-zx: oh I meant like the [void operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void)

backend/src/database/migrations/20221117031825-update-insert-job.js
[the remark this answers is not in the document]
prestonlimlianjie: Tackle in a separate issue: https://github.com/opengovsg/postmangovsg/issues/1893

worker/src/core/services/notification.service.ts
richard-yak: done

stanleynguyen (approved):

stanleynguyen:
Don't forget to do the same checklist items and deploy to staging as well yea

PR #1885, opened 2022-12-09
feat: bcc campaign messages to author

stanleynguyen:
## Problem

Closes [Ticket](https://www.notion.so/opengov/BCC-on-Postman-852b225df35a42b2bb034eaf38efd2c6)

## Solution

Implement it duh

**Some extra notes**: This PR was implemented with future "bcc to anyone" possibility in mind, hence going through the hoop of putting `bcc` inside the params. However, we haven't handled bounces and other notifications that are related to the bcc instead of the main email properly yet but for now since it's only `bcc`-ed to the main user email, we can perhaps live with this

## Deployment Checklist

- [ ] Migrate DB

### DIFF
--- a/backend/src/core/models/campaign.ts
+++ b/backend/src/core/models/campaign.ts
@@ -104,6 +104,12 @@ export class Campaign extends Model<Campaign> {
   })
   demoMessageLimit!: number
 
+  @Column({
+    type: DataType.BOOLEAN,
+    defaultValue: false,
+  })
+  shouldBccToMe!: boolean
+
   // Sets key in s3Object json
   static async updateS3ObjectKey(
     id: number,
--- a/backend/src/database/migrations/20221208074426-add-should-campaign-should-bcc.js
+++ b/backend/src/database/migrations/20221208074426-add-should-campaign-should-bcc.js
@@ -1,0 +1,15 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('campaigns', 'should_bcc_to_me', {
+      type: Sequelize.DataTypes.BOOLEAN,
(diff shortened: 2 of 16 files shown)


backend/src/database/migrations/20221208074426-add-should-campaign-should-bcc.js
```diff
@@ -0,0 +1,14 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('campaigns', 'should_bcc_to_me', {
+      type: Sequelize.DataTypes.BOOLEAN,
```
richard-yak, line 6: recommend adding `allowNull: false` due to use-case definition of this field as sequelize default allowNull param is true.

backend/src/core/models/campaign.ts
```diff
@@ -104,6 +104,12 @@ export class Campaign extends Model<Campaign> {
   })
   demoMessageLimit!: number
 
+  @Column({
+    type: DataType.BOOLEAN,
+    defaultValue: false,
+  })
+  shouldBccToMe!: boolean
```
richard-yak, line 111: side note for schema design consideration: although not immediately imperative, this is an e-mail campaign unique field. in strict DAO sense, this field should be in an extension table unique to e-mail campaigns.

richard-yak (changes requested):
highly recommend non nullable in new col as best practice

backend/src/core/models/campaign.ts
stanleynguyen: I'd prefer we keep thing simple first as there's not enough benefits in doing this

richard-yak (approved):
LGTM

PR #1888, opened 2022-12-14
feat: save transactional sms

tan-zx:
[Ticket](https://www.notion.so/opengov/fd466213f133486d85e4117f1ded521d?v=052d3609255c4dfcb2cf173aae7f7c5d&p=5114247758dd4f08997924f32e9f99c3&pm=s)

[RFC](https://www.notion.so/opengov/2022-12-13-RFC-for-Saving-SMSes-Sent-via-API-1eefc5346aae48d996786ae04ad57437)

Reviewers, please read the RFC + review the implementation.

Remaining tasks are:
- [x] Add tests
- [x] Update OpenAPI schema
- [x] Test on staging (works as intended!)

Deployment notes:
- [ ] Run database migration first (alr migrated on staging; to reverse if there are any changes to the migration history caa 16 Dec 2022)
- [ ] Need to do the following, need to grant permissions to `flamingo_iam` first (because new table is created). Specifically:
  - [ ] SSH to jumphost (`npm run connectjumphost:${staging|prod}`), connect to RDS via `psql -h <XXX>.rds.amazonaws.com -p 5432 -U <user_name><db_name> -W`
  - [ ] Run `GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO flamingo_iam;` (NB semi colon is important)
  - [ ] Run `GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA public TO flamingo_iam;` (NB semi colon is important)

### DIFF
--- a/backend/src/sms/middlewares/sms-transactional.middleware.ts
+++ b/backend/src/sms/middlewares/sms-transactional.middleware.ts
@@ -3,32 +3,75 @@ import { SmsTransactionalService } from '@sms/services'
 import { loggerWithLabel } from '@core/logger'
 import { TemplateError } from '@shared/templating'
 import { RateLimitError, InvalidRecipientError } from '@core/errors'
+import { SmsMessageTransactional } from '@sms/models/sms-message-transactional'
 
 const logger = loggerWithLabel(module)
+
+function convertMessageModelToResponse(message: SmsMessageTransactional) {
+  return {
+    id: message.id,
+    credentialsLabel: message.credentialsLabel,
+    recipient: message.recipient,
+    body: message.body,
+    message_id: message.messageId,
+    created_at: message.createdAt,
+    updated_at: message.updatedAt,
+  }
+}
+
+async function saveMessage(
+  req: Request,
+  _: Response,
+  next: NextFunction
+): Promise<void> {
+  const action = 'saveMessage'
+  logger.info({ message: 'Saving SMS', action })
+  const { recipient, body, label } = req.body
+
+  const smsMessageTransactional = await SmsMessageTransactional.create({
+    userId: req.session?.user?.id,
+    recipient,
+    body,
+    credentialsLabel: label,
+    messageId: null,
+    // not sure why unknown is needed to silence TS (yet other parts of the code base can just use `as Model` directly hmm)
+  } as unknown as SmsMessageTransactional)
+  req.body.smsMessageTransactionalId = smsMessageTransactional.id // for subsequent middlewares to distinguish whether this is a transactional SMS
+  next()
+}
 
 async function sendMessage(
   req: Request,
   res: Response,
   next: NextFunction
 ): Promise<void> {
-  const ACTION = 'sendMessage'
+  const action = 'sendMessage'
 
   const { credentials } = res.locals
   try {
-    const { recipient, body } = req.body
-
-    logger.info({ message: 'Sending SMS', action: ACTION })
-    await SmsTransactionalService.sendMessage({
+    const smsMessageTransactional = await SmsMessageTransactional.findByPk(
+      req.body.smsMessageTransactionalId
+    )
+    if (!smsMessageTransactional) {
+      throw new Error('Unable to find entry in sms_messages_transactional')
+    }
+    logger.info({ message: 'Sending transactional SMS (middleware)', action })
+    const messageId = await SmsTransactionalService.sendMessage({
       credentials,
-      recipient,
-      body,
+      recipient: smsMessageTransactional.recipient,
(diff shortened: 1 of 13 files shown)


backend/src/sms/middlewares/sms-transactional.middleware.ts
```diff
@@ -3,32 +3,75 @@ import { SmsTransactionalService } from '@sms/services'
...
-
-    logger.info({ message: 'Sending SMS', action: ACTION })
-    await SmsTransactionalService.sendMessage({
+    const smsMessageTransactional = await SmsMessageTransactional.findByPk(
+      req.body.smsMessageTransactionalId
+    )
+    if (!smsMessageTransactional) {
+      throw new Error('Unable to find entry in sms_messages_transactional')
+    }
+    logger.info({ message: 'Sending transactional SMS (middleware)', action })
```
stanleynguyen, line 58: This log could be a bit more useful if we replace/add on top of `action` the message object

backend/src/sms/middlewares/sms.middleware.ts
```diff
@@ -153,22 +145,58 @@ export const InitSmsMiddleware = (
...
-      action: 'getCredentialsFromLabel',
+      action: 'getCredentialsFromLabelCampaign',
     }
     try {
       /* Determine if credential name can be used */
       const campaign = campaignId
         ? await SmsService.findCampaign(+campaignId, userId)
         : null
-      const credentialName = campaign?.demoMessageLimit
+      const canUseDemoCredentials = !!campaign?.demoMessageLimit // false if (1) campaign is null or (2) demoMessageLimit is 0
```
stanleynguyen, line 155: Can we use campaign?.demoMessageLimit && campaign?.demoMessageLimit > 0 instead? I know this is equivalent of the existing code, but it's quite confusing

backend/src/sms/middlewares/sms.middleware.ts
```diff
@@ -153,22 +145,58 @@ export const InitSmsMiddleware = (
...
+      )
+      res.locals.credentialName = credentialName
+      return next()
+    } catch (err) {
+      const errAsError = err as Error
+      logger.error({
+        ...logMeta,
+        message: `${errAsError.stack}`,
+      })
+      return res.status(400).json({ message: `${errAsError.message}` })
```
stanleynguyen, line 177: We shouldn't just return the error message out plainly just in case it contain confidential information. For HTTP errors to be returned with specific message, the try/catch should be very specific to the block containing error that u're tryna catch too

backend/src/sms/middlewares/sms-transactional.middleware.ts
stanleynguyen: But I know it's legacy and we're perhaps better off refactoring all of these in one shot

tan-zx: we create a separate issue to deal with this to overhaul `logger.info`
- we alr have trace id (on DataDog), so we can correlate logs across different functions, so we don't need to log message in detail
- no point logging these happy paths `logger.info` across the entire codebase

I'll create an issue after this

backend/src/sms/middlewares/sms.middleware.ts
```diff
@@ -153,22 +145,58 @@ export const InitSmsMiddleware = (
...
+    const logMeta = {
+      label,
+      action: 'getCredentialsFromLabelTransactional',
+    }
+    try {
+      /* Determine if credential name can be used */
+      const credentialName = await getCredentialName(label, +userId)
+
+      /* Get credential from the name */
+      res.locals.credentials = await credentialService.getTwilioCredentials(
```
tan-zx, line 197: Separating out `getCredentialsFromLabel` into `getCredentialsFromLabelCampaign` and `getCredentialsFromLabelTransactional` so as to avoid weird campaign-error related error if the label is wrong.

The new logic for `getCredentialsFromLabelTransactional` simply checks the credential directly using the `label` and `userId`, so it won't be able to use demo credentials at all. (So the overall logic remains unchanged.)

Current `try/catch` clause is probably too broad but this is consistent with original code, so not going to change in this PR.

stanleynguyen (approved):

backend/src/sms/middlewares/sms-transactional.middleware.ts
tan-zx: created https://github.com/opengovsg/postmangovsg/issues/1902

stanleynguyen:
I cancelled the non-serverless deployment. Let's migrate and deploy this tmr @tan-zx ?

PR #1889, opened 2022-12-19
fix(backend): truncate attachment content from logs

stanleynguyen:
## Problem

Attachments logged in full taking up our diskspace

Closes [insert issue #]

## Solution

Truncate their content from logs

## Deployment Checklist

- N/A


tan-zx (approved):
lgtm

PR #1892, opened 2022-12-22
docs(api): removed .doc, .ppt, .xls from attachment file types supported

lisatjide:
the library file-type we use currently does not accept cfb file types because these file types are old formats of Microsoft (.doc, .xls, .ppt) can be weaponized and exploited for malicious usecases. more on this [here](https://www.decalage.info/en/book/export/html/55).

the team decided to disallow cfb file types for 2 reasons:

keeping our system more secure by avoiding highly exploitable file types
aligning our practices with go.gov.sg which we intend to transit over our attachments in the longer run
list of cfb file types that we will not support:

.doc - Microsoft Word 97-2003 Document
.xls - Microsoft Excel 97-2003 Document
.ppt - Microsoft PowerPoint97-2003 Document

### DIFF
--- a/backend/src/core/services/file-extensions.service.ts
+++ b/backend/src/core/services/file-extensions.service.ts
@@ -6,7 +6,6 @@ const DEFAULT_ALLOWED_EXTENSIONS = [
   'bmp',
   'csv',
   'dgn',
-  'doc',
   'docx',
   'dwf',
   'dwg',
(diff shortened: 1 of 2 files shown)


tan-zx (approved):
lgtm! added a commit to remove from `DEFAULT_ALLOWED_EXTENSIONS` too

PR #1897, opened 2022-12-26
build(actions): revert application version upon e2e failure

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Postman-CI-CD-E2E-Rolling-Back-Was-Canary-RFC-ee42f3356a264cdf8832649bb988b569)

## Solution

As stated in the ticket's comment.
To be followed up with another PR for notifying on Slack on build successes & failures

## Deployment Checklist

- [ ] N/A

### DIFF
--- a/.github/workflows/deploy-eb.yml
+++ b/.github/workflows/deploy-eb.yml
@@ -1,5 +1,13 @@
 name: Deploy backend to AWS Elastic Beanstalk
-on: workflow_call
+on:
+  workflow_call:
+    outputs:
+      revert_command_backend:
+        description: "Command to revert backend deployment"
+        value: ${{ jobs.deploy_backend.outputs.revert_cmd }}
+      revert_command_callback:
+        description: "Command to revert callback server deployment"
+        value: ${{ jobs.deploy_callback.outputs.revert_cmd }}
 env:
   # Update this common config
   DIRECTORY: backend
@@ -14,9 +22,9 @@ jobs:
         run: |
           echo "Running on branch ${{ github.ref }}"
           if [ "${{ github.ref }}" = "refs/heads/master" ]; then
-            echo "::set-output name=current_env::production"
+            echo "current_env=production" >> $GITHUB_OUTPUT
           else
-             echo "::set-output name=current_env::staging"
+             echo "current_env=staging" >> $GITHUB_OUTPUT
           fi
 
   lint-test:
@@ -51,7 +59,7 @@ jobs:
         env:
           NODE_OPTIONS: "--max-old-space-size=4096"
         run: |
-          cd shared 
+          cd shared
           npm ci
           cd ..
           cd $DIRECTORY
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy-frontend.yml
```diff
@@ -16,14 +25,21 @@ jobs:
...
           fi
         id: extract_branch
+      - name: Generate revert command
+        id: generate-revert-command
+        env:
+          BRANCH: ${{ steps.extract_branch.outputs.branch }}
+        run: |
+          aws amplify list-jobs --app-id $APP_ID --branch-name $BRANCH --max-items 1 | jq '.jobSummaries[0]' | echo
+          JOB_ID=$(aws amplify list-jobs --app-id $APP_ID --branch-name $BRANCH --max-items 1 | jq '.jobSummaries[0].jobId')
+          echo "revert_cmd=aws amplify start-job --app-id $APP_ID --branch-name $BRANCH --job-id $JOB_ID --job-type RETRY" >> $GITHUB_OUTPUT
```
jhibberd-gov, line 40: I'm new to GitHub Actions workflows but my understanding is that for each component (backend, frontend, etc.) we're generating a revert command by querying the current code revision deployed for each component. I wonder whether doing this on a component-by-component basis could cause inconsistencies, where we accidentally revert different components to different code revisions. Is there any way we can query for the last revision successfully deployed, and generate the revert commands based on that, to ensure all components will be reverted to the same revision.

NB There's a reasonable chance I misunderstood the code, and this isn't an issue.

.github/workflows/deploy-worker.yml
```diff
@@ -44,11 +52,19 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
+      - name: Set environment variables
+        run: |
+          if [$CURRENT_ENV == 'production']
+          then
+            echo "SECRET_ID=env/eb/postmangovsg-production-amz2" >> $GITHUB_ENV
+          else
+            echo "SECRET_ID=env/eb/postmangovsg-staging-amz2" >> $GITHUB_ENV
+          fi
```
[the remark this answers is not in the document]
stanleynguyen: Actually for our case yes, but it doesn't represent the correct nature of these IDs imo

.github/workflows/deploy-eb.yml
```diff
@@ -90,35 +98,35 @@ jobs:
...
           sed -i -e "s|@REPO|$ECR_REPOSITORY|g" Dockerrun.aws.json 
           docker build -f Dockerfile -t $ECR_REPOSITORY:$IMAGE_TAG .
-          docker tag $ECR_REPOSITORY:$IMAGE_TAG $ECR_REPOSITORY:$BRANCH
+          docker tag $ECR_REPOSITORY:$IMAGE_TAG $ECR_REPOSITORY:$CURRENT_ENV
           docker push -a $ECR_REPOSITORY
           zip -r "$IMAGE_TAG.zip" .ebextensions .platform Dockerrun.aws.json
 
       - name: Copy to S3
         env:
-          BUCKET_NAME: ${{ secrets.EB_BUCKET_NAME }}
```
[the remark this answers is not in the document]
stanleynguyen: Yup it's actually a refactor that's bundled into this PR as well

.github/workflows/deploy-frontend.yml
stanleynguyen: yup I share this concern too, but right now there's no way to figure out `last successful deployment` that's why I'm resorting to this

stanleynguyen: But I dont think it's a big concern cos we will be notified of failed deployments and will revert the code before the next deployment attempt

jhibberd-gov:
Thanks for responding to all my comments!

jhibberd-gov (approved):

PR #1900, opened 2022-12-28
build(slack): send messages to slack to notify about builds

stanleynguyen:
## Problem

Continued effort from #1897 

## Solution

Use slack gh actions

## Deployment Checklist

- [x] Make sure #1897 has been merged
- [ ] Rebase master into this branch before merging

### DIFF
--- a/.github/workflows/non-serverless-deploy.yml
+++ b/.github/workflows/non-serverless-deploy.yml
@@ -1,7 +1,7 @@
 # This action is to orchestrate the non-serverless stuffs deployment
 # to make sure cypress e2e test is run post-deployments.
 # The reason we need to do this hack is because there's currently no way
-# to use `workflow_run` trigger with requirement being all depdency workflows'
+# to use "workflow_run" trigger with requirement being all depdency workflows'
 # complete. More here: https://github.com/community/community/discussions/16059
 # Side note: we need to follow .github/workflows convention https://docs.github.com/en/actions/using-workflows/reusing-workflows
 name: Deploy and E2E test non-serverless stuffs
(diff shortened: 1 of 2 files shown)


.github/workflows/non-serverless-deploy.yml
```diff
@@ -14,6 +14,51 @@ on:
...
+  slack-started:
+    name: Send slack message when deployment starts
+    runs-on: ubuntu-latest
+    outputs:
+      message_id: ${{ steps.slack.outputs.ts }}
+    steps:
+      - uses: slackapi/slack-github-action@v1.23.0
+        id: slack
+        with:
+          channel-id: C04H0FTGS5P # postman-deploys
```
jhibberd-gov, line 26: nit: Better as an environment variable, similar to `SLACK_BOT_TOKEN`?

.github/workflows/non-serverless-deploy.yml
```diff
@@ -34,3 +79,104 @@ jobs:
...
+                      }
+                  ],
+                  "footer_icon": "https://github.githubassets.com/favicon.ico",
+                  "footer": "<https://github.com/opengovsg/postmangovsg | opengovsg/postmangovsg>"
+                }
+              ]
+            }
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+  slack-failure:
```
jhibberd-gov, line 132: nit: Inconsistency on whether jobs are preceded by newline (see above).

.github/workflows/non-serverless-deploy.yml
```diff
@@ -34,3 +79,104 @@ jobs:
...
+      - deploy-frontend
+      - deploy-worker
+      - e2e-test
+    if: success()
+    name: Send Slack message about successful build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: slackapi/slack-github-action@v1.23.0
+        with:
+          channel-id: C04H0FTGS5P # postman-deploys
```
jhibberd-gov, line 96: nit: Looks like Git Actions workflows supports "workflow level" environment variables. Since this value is used across multiple jobs does it make sense to define it once at the workflow level?
https://docs.github.com/en/actions/learn-github-actions/contexts#env-context

jhibberd-gov (approved):
Overall this LGTM (based on my knowledge of GitHub Actions and Slack). There's quite a bit of repeated config (e.g. `needs` and `payload`) but unless there's a clear way to abstract these values into an env var I'm not sure what else we could do to make it more maintainable (i.e. updating `needs` for one job but forgetting to make the same update to the other jobs).

PR #1904, opened 2023-01-04
docs: update README

tan-zx:
Removed
- Out-of-date description of our release procedures
- Description of our architecture (the db schema is certainly out-of-date...I say we should just delete all this since there's no point maintaining these — not like we're getting forked or receiving meaningful open-source contributions) 

Not sure whether the following parts are still up-to-date:
- Downtime procedure
- Description of our features (seems to be missing Telegram, should probably add scheduled sending, but probably better place is to update landing page?)


richard-yak (approved):

tan-zx:
> Not sure whether the following parts are still up-to-date:
> 
> * Downtime procedure
> * Description of our features (seems to be missing Telegram, should probably add scheduled sending, but probably better place is to update landing page?)

Any thoughts on these? 
- For downtime procedure, I don't think there is any value in keeping the files in the repo (both the README in root and the `index.md` file it points to) and we should just remove them, since they are liable to become outdated/confusing
- Same with features, don't think anyone reads them, better to just keep the landing page updated

tan-zx:
I suspect this is outdated too:

tan-zx:
gonna merge in first — can deal with the rest in the future

PR #1907, opened 2023-01-10
fix: attachment array bug

tan-zx:
## Problem

We previously modified the code to redact attachment data in the log to ensure our logs will be reasonably sized. However, the modification assumes that the `req.body.attachments` field will be an array, which is not the case.

## Solution

Have created an utility function `ensureAttachmentsFieldIsArray` to use in both part of the codebase that does what it says.

### DIFF
--- a/backend/src/core/utils/attachment.ts
+++ b/backend/src/core/utils/attachment.ts
@@ -1,0 +1,12 @@
+import { UploadedFile } from 'express-fileupload'
+
+const ensureAttachmentsFieldIsArray = (
+  attachments: UploadedFile | UploadedFile[]
+) => {
+  if (!Array.isArray(attachments)) {
+    return [attachments]
+  }
+  return attachments
+}
+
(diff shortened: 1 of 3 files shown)


backend/src/core/utils/attachment.ts
```diff
@@ -0,0 +1,12 @@
+import { UploadedFile } from 'express-fileupload'
+
+const ensureAttachmentsFieldIsArray = (
+  attachments: UploadedFile | UploadedFile[]
```
stanleynguyen, line 4: Why do we have to declared `UploadedFile` type ah? We're returning array for all cases

tan-zx: Oh this is the argument type, not the return type. Yup returning array for all cases. 

Strictly the UploadedFile type is not necessary la, but I thought it be good to show which library this is meant to be working with

stanleynguyen (approved):

PR #1921, opened 2023-01-18
fix(backend): handle malformed request gracefully

stanleynguyen:


tan-zx:
Based on [this](https://github.com/ntedgi/express-body-parser-error-handler)

tan-zx (approved):

PR #1923, opened 2023-01-25
feat: forbid attachments from default Postman from address

tan-zx:
## Problem

See [description here](https://opengovproducts.slack.com/archives/C049QPXT27M/p1673860380333379)

## Solution

Rejects request to send tx API email with attachments if user is using default Postman email address

## Deployment Checklist
- [x] Test on staging
- [x] Check to ensure existing users are not sending attachments using default Postman email address


backend/src/core/middlewares/file-attachment.middleware.ts
```diff
@@ -48,7 +51,36 @@ function preprocessPotentialIncomingFile(
...
+  if (req.files?.attachments) {
+    const { from } = req.body
+    if (isDefaultFromAddress(from)) {
+      res.status(403).json({
+        message: 'Attachments are not allowed for default email address',
+      })
+      return
+    }
+    const { fromAddress } = parseFromAddress(from)
+    const exists = await CustomDomainService.existsFromAddress(fromAddress)
```
stanleynguyen, line 70: This should stop them from even sending the email no?

tan-zx: oh I was thinking that is the intended behavior (unless you mean we should just remove the attachment and still send the email?) I think that's a bit trickier since, if the user isn't testing properly, things might fail silently (they send an email talking about some non-existent attachment)

not sure if I'm misunderstanding sth

backend/src/core/middlewares/file-attachment.middleware.ts
```diff
@@ -48,7 +51,36 @@ function preprocessPotentialIncomingFile(
...
+async function checkAttachmentPermission(
+  req: Request,
+  res: Response,
+  next: NextFunction
+): Promise<void> {
+  if (req.files?.attachments) {
+    const { from } = req.body
+    if (isDefaultFromAddress(from)) {
+      res.status(403).json({
+        message: 'Attachments are not allowed for default email address',
```
stanleynguyen, line 65: Maybe we can add a bit more info here `Attachments are not allowed for Postman default from-email-address`.

tan-zx: done

backend/src/core/middlewares/file-attachment.middleware.ts
tan-zx: discussed, to simplify requirements to forbidding attachments from default Postman from address

otherwise, the extra check is a bit confusing (considering we alr do the same check later on)

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -427,14 +449,19 @@ describe(`${emailTransactionalRoute}/send`, () => {
       .spyOn(FileExtensionService, 'hasAllowedExtensions')
       .mockResolvedValue(false)
 
+    await EmailFromAddress.create({
+      email: user.email,
+      name: 'Agency ABC',
+    } as EmailFromAddress)
+
```
tan-zx, line 456: @stanleynguyen just curious, do you think I should shift this to `beforeEach`, or leaving it here is OK

technically not required for the test to pass (with the latest change), but good to leave it in so it's clear what the "control variable" is?

stanleynguyen (approved):

PR #1924, opened 2023-01-25
fix: use DataType.Text for sms transactional body

tan-zx:
## Problem

Previously, `DataType.String` defaults to 255 chars, which may be too short.

Related:
- https://stackoverflow.com/questions/4848964/difference-between-text-and-varchar-character-varying
- https://dba.stackexchange.com/questions/189876/size-limit-of-character-varying-postgresql

## Solution

Use `DataType.Text` instead

## Deployment Checklist
- [x] Test on staging
- [x] Order of running migration vs merging in PR doesn't matter

Backend code unchanged

### DIFF
--- a/backend/src/database/migrations/20230125083046-convert-tx-sms-body-datatype.js
+++ b/backend/src/database/migrations/20230125083046-convert-tx-sms-body-datatype.js
@@ -1,0 +1,17 @@
+'use strict';
+
+module.exports = {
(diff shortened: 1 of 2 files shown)


tan-zx:
btw @stanleynguyen I removed the `'long'` because I realized when I ran the db migration, I didn't include it and looking at the table again, for Postgres it doesn't make a difference

image

richard-yak (approved):

PR #1928, opened 2023-02-06
docs(openapi): add transactional SMS get endpoints

stanleynguyen:


tan-zx:
Note to subsequent readers: key changes are at line 3311 onwards and line 5092 onwards

backend/openapi.yaml
```diff
@@ -3298,13 +3302,225 @@ paths:
...
+              - lt: less than
+              - lte: less than or equal
+          required: false
+          schema:
+            type: object
+            minProperties: 1
+            properties:
+              gt:
+                type: string
+                format: date-time (ISO 8601)
```
tan-zx, line 3362: I think the `date-time` field is inconsistent between `date-time` and `date-time (ISO 8601)`. think we can standardise on the latter if it's more precise? also can specify our time zone to be more precise?

backend/openapi.yaml
```diff
@@ -4881,22 +5097,46 @@ components:
...
+          example: 'Hello world'
         created_at:
           type: string
           format: date-time
         updated_at:
           type: string
           format: date-time
+        accepted_at:
+          type: string
+          format: date-time
```
tan-zx, line 5115: yeah this is the part where it's `date-time` instead of `date-time (ISO 8601)`

think we should make it consistent for the tx email API also

backend/openapi.yaml
```diff
@@ -3298,13 +3302,225 @@ paths:
...
+            default: 0
+        - in: query
+          name: status
+          description: status of messages to filter for
+          required: false
+          schema:
+            type: array
+            items:
+              type: string
+              enum: [UNSENT, ACCEPTED, SENT, DELIVERED, ERROR]
```
tan-zx, line 3344: how are we aligning these with [the statuses supported by Twilio](https://support.twilio.com/hc/en-us/articles/223134347-What-are-the-Possible-SMS-and-MMS-Message-Statuses-and-What-do-They-Mean-)? 

(not sure whether these are the same as the ones in the delivery reports, but don't think aligning with that is that important)

backend/openapi.yaml
```diff
@@ -4881,22 +5097,46 @@ components:
...
+          format: date-time
+        errored_at:
+          type: string
+          format: date-time
+        error_code:
+          type: string
+        status:
+          type: string
+          enum: [UNSENT, ACCEPTED, SENT, DELIVERED, ERROR]
+          description: >
```
tan-zx, line 5130: **[question/suggestion]** how are we aligning these with [the statuses supported by Twilio](https://support.twilio.com/hc/en-us/articles/223134347-What-are-the-Possible-SMS-and-MMS-Message-Statuses-and-What-do-They-Mean-)? maybe we can link to the article in the description here? some notes:
- some SMSes that are actually delivered may be stuck at `SENT` because the telcos don't provide callbacks
- I think we can just put `failed` and `undelivered` under `ERROR` and put more info in the `error_code` field? 

**[thinking out loud]** not sure whether these are the same as the ones in the camapgin delivery reports, but don't think aligning with that is that important

**[change request]** there are some typos below using email instead of SMS haha

**[nit]** would be good to use `SMS` instead of `sms`

tan-zx (changes requested):
left some comments, looks good on the whole!

backend/openapi.yaml
stanleynguyen: They're the same thing, `date-time` is `ISO 8601`

backend/openapi.yaml
stanleynguyen: ERROR is for both `failed` and `undelivered` in the current design. What I'm doing is exactly what u're suggesting above

tan-zx: think it's good to add a link to the Twilio, so the implementation/API user will know

backend/openapi.yaml
```diff
@@ -4881,22 +5097,46 @@ components:
...
-          example: "ogp-twilio-creds"
+          example: 'ogp-twilio-creds'
         recipient:
           type: string
-          example: "98765432"
+          example: '98765432'
         body:
           type: string
-          example: "Hello world"
-        message_id:
```
tan-zx, line 4891: just to clarify, the idea here is we'll still store this in our db, but we won't be returning this to the user? why or why not? (I was thinking with the `message_id` in theory they can go to Twilio to get more info from them?)

stanleynguyen: I can guarantee u no one will go directly to Twilio

tan-zx: fair enough

backend/openapi.yaml
```diff
@@ -4881,22 +5097,47 @@ components:
...
+        error_code:
+          type: string
+          description: This error code corresponds with [error codes from Twilio](https://www.twilio.com/docs/sms/api/message-resource#delivery-related-errors).
+        status:
+          type: string
+          enum: [UNSENT, ACCEPTED, SENT, DELIVERED, ERROR]
+          description: >
+            * `UNSENT` - initial state of a newly created transactional sms (this status is not returned in the course of a successful request to send an email)
+
+            * `ACCEPTED` - email has been accepted by our sms provider (this status is returned in the course of a successful request to send an sms)
```
tan-zx, line 5134: this one should say SMS instead of email

backend/openapi.yaml
```diff
@@ -4881,22 +5097,47 @@ components:
...
+          type: string
+          format: date-time
+        error_code:
+          type: string
+          description: This error code corresponds with [error codes from Twilio](https://www.twilio.com/docs/sms/api/message-resource#delivery-related-errors).
+        status:
+          type: string
+          enum: [UNSENT, ACCEPTED, SENT, DELIVERED, ERROR]
+          description: >
+            * `UNSENT` - initial state of a newly created transactional sms (this status is not returned in the course of a successful request to send an email)
```
tan-zx, line 5132: this one should say SMS instead of email

backend/openapi.yaml
tan-zx: actually let me fix all these nits directly

tan-zx (approved):

tan-zx:
OK approved

PR #1933, opened 2023-02-18
feat: rate limited emails should not be saved to db

tan-zx:
## Problem

- Previously, rate-limited emails are sent to database. This is a problem as we scale up as it means even rate-limited emails are causing db writes. This issue is thankfully flagged by @stanleynguyen prior to our upcoming big deployment.

## Solution

- Move rate-limit up, so that it precedes writing to database
- Remove tests that checks that rate-limited emails have been written to db

## Deployment Checklist

N/A just deploy only


tan-zx:
Actually sth interesting to check is whether we have any rate-limited error codes in our db so far

stanleynguyen (approved):
lgtm ser

PR #1936, opened 2023-02-21
feat: sms transactional API and callback status handling 

richard-yak:
## Problem

This is part of [bringing Postman API to v1](https://www.notion.so/opengov/Bring-Postman-API-to-v1-0-78283c4ac4e3445ca8c4f1b20019a45c?pvs=4)

## Solution

Implementing of the relevant endpoints that were lacking, and also adding callback handling based on Twilio's responses to our system. 

## Deployment Checklist

- [x] verify that BACKEND_URL is in secrets manager for production server
- [x] Run migration on production DB


backend/src/sms/models/sms-message-transactional.ts
```diff
@@ -38,4 +39,13 @@ export class SmsMessageTransactional extends Model<SmsMessageTransactional> {
 
   @Column({ type: DataType.STRING, allowNull: true })
   messageId: string | null
+
+  @Column({
+    type: DataType.ENUM(...Object.values(MessageStatus)),
```
stanleynguyen, line 44: It was intentional that we want to introduce a better message status system (than the existing `MessageStatus`) for transactional usage with an eventual goal of moving campaign message status to that as well

stanleynguyen: Could you follow the API specs for the above reason?

backend/src/sms/services/sms-transactional.service.ts
```diff
@@ -31,9 +41,70 @@ async function sendMessage({
...
+  offset,
+  sortBy,
+  orderBy,
+  status,
+  filterByTimestamp,
+}: {
+  userId: string
+  limit?: number
+  offset?: number
+  sortBy?: TransactionalEmailSortField
```
stanleynguyen, line 63: Should create a separate `TransactionalSmsSortField`

backend/src/sms/services/sms-transactional.service.ts
```diff
@@ -31,9 +41,70 @@ async function sendMessage({
...
+  orderBy?: Ordering
+  status?: TransactionalEmailMessageStatus
```
stanleynguyen, line 65: `TransactionalSmsMessageStatus` instead

backend/src/sms/services/sms.service.ts
```diff
@@ -82,10 +88,13 @@ const sendMessage = (
     throw new InvalidRecipientError('Invalid phone number')
   }
 
-  const client = config.get('smsFallback.activate')
-    ? new SnsSmsClient()
-    : new TwilioClient(credential)
-  return client.send(recipient, message)
+  if (config.get('smsFallback.activate')) {
+    const client = new SnsSmsClient()
```
stanleynguyen, line 92: I actually think we should just deprecate Sns fallback. Based on my observations from the last few incidents, we didn't even activate this option at all. Planning to include this deprecation as part of the refactor but it could be started here as well.
Lmk if you got a differing opinion @tan-zx

richard-yak: done

backend/src/sms/services/sms-transactional.service.ts
richard-yak: done

backend/src/sms/services/sms-transactional.service.ts
richard-yak: done

backend/src/sms/models/sms-message-transactional.ts
richard-yak: have refactored

backend/src/database/migrations/20230216082014-update-transactional-sms.js
```diff
@@ -0,0 +1,27 @@
...
+const {MessageStatus} = require('@core/constants');
+const {TransactionalSmsMessageStatus} = require('@sms/models');
+module.exports = {
+    up: async (queryInterface, Sequelize) => {
+        await queryInterface.addColumn('sms_messages_transactional', 'status', {
+            type: Sequelize.DataTypes.ENUM(
+                Object.values(TransactionalSmsMessageStatus)
+            ),
+            allowNull: false,
+            defaultValue: TransactionalSmsMessageStatus.Delivered
```
stanleynguyen, line 12: `defaultValue` should be `UNSENT`

backend/src/sms/middlewares/sms-callback.middleware.ts
```diff
@@ -48,7 +65,26 @@ const parseEvent = async (
...
 
+const parseTransactionalEvent = async (
+  req: Request,
+  res: Response,
+  next: NextFunction
+): Promise<Response | void> => {
+  try {
+    await SmsCallbackService.parseTransactionalEvent(req)
+    logger.info({
+      message: 'Successfully parsed transactional event',
```
stanleynguyen, line 76: I dont think we need to log at this point of success, if any, we should log before the processing/parsing

backend/src/sms/middlewares/sms-callback.middleware.ts
```diff
@@ -48,7 +65,26 @@ const parseEvent = async (
...
+      action: 'parseTransactionalEvent',
+    })
+    return res.sendStatus(200)
+  } catch (err) {
+    return next(err)
+  }
```
stanleynguyen, line 82: No need for this try/catch wrapping as it's already the default behavior

backend/src/sms/routes/sms-callback.routes.ts
```diff
@@ -30,4 +31,24 @@ router.post(
...
+ *      tags:
+ *        - Settings
+ *      responses:
+ *        200:
+ *          description: OK
+ *        400:
+ *          description: Bad Request
+ */
+router.post(
+  '',
```
stanleynguyen, line 50: nit: Use `/` instead of ''

backend/src/sms/services/sms-callback.service.ts
```diff
@@ -94,7 +114,65 @@ const parseEvent = async (req: Request): Promise<void> => {
...
+      }
+    )
+  } else {
+    // longer messages are delivered in multiple segments
+    // each segment has a separate delivery status
+    // Update the message as successful only if there does not exist previous failed status
+    // update only if sent or delivered
+    const mappedStatus = twilioMessageStatus.toUpperCase()
+    if (
+      mappedStatus == TransactionalSmsMessageStatus.Sent ||
```
stanleynguyen, line 155: *`===`

backend/src/sms/services/sms-callback.service.ts
```diff
@@ -94,7 +114,65 @@ const parseEvent = async (req: Request): Promise<void> => {
...
+      mappedStatus == TransactionalSmsMessageStatus.Delivered
```
stanleynguyen, line 156: *`===`

backend/src/sms/services/sms-callback.service.ts
```diff
@@ -94,7 +114,65 @@ const parseEvent = async (req: Request): Promise<void> => {
...
+    ) {
+      await SmsMessageTransactional.update(
+        {
+          updatedAt: new Date(),
+          status: twilioMessageStatus.toUpperCase(),
```
stanleynguyen, line 161: I feel like we should have a function to cover twilioMessageStatus to our native txnmessagestatus then use across this handler

backend/src/sms/services/twilio-client.class.ts
```diff
@@ -1,6 +1,6 @@
 import twilio from 'twilio'
```
stanleynguyen, line 1: Do we still need to keep this file if we're moving the twilio client to shared?

backend/src/database/migrations/20230216082014-update-transactional-sms.js
richard-yak: the issue with defaultValue unsent in migration script would be that on column creation, the old records would all be unsent. Then i would have to run another update script to update to delivered. currently i init the column to unsent when i create the record

backend/src/sms/middlewares/sms-callback.middleware.ts
richard-yak: done

backend/src/sms/routes/sms-callback.routes.ts
richard-yak: done

backend/src/sms/services/sms-callback.service.ts
richard-yak: done

backend/src/sms/services/sms-callback.service.ts
richard-yak: done

backend/src/sms/services/sms-callback.service.ts
richard-yak: used a switch case to map back, given that it's not a 1 to 1 relation i will only capture statuses which are the same, and the rest defaulted to unsent state

backend/src/sms/services/twilio-client.class.ts
richard-yak: removed, side note: in future to also allow worker to use shared twilio when bandwidth allows (have included this in the comments)

backend/src/sms/middlewares/sms-callback.middleware.ts
richard-yak: moved

backend/src/database/migrations/20230216082014-update-transactional-sms.js
richard-yak: prod has 724 records, if we are okay with running raw sql in prod to manually update to delivered, i can change the migration script to default to unsent, what do you think? @stanleynguyen

stanleynguyen: > prod has 724 records, if we are okay with running raw sql in prod to manually update to delivered, i can change the migration script to default to unsent, what do you think? @stanleynguyen

I dont think we need to touch the existing records, just need to update the current column definition manually to have `UNSENT` as default and also update the js migration file (without running it)

richard-yak: hmmm status is a new column, so if I set default value as unsent, it'll automatically set all previous records status to unsent.

stanleynguyen: wait hang on, I think we're confusing ourselves here. Can huddle?

shared/src/clients/twilio-client.class/index.ts
```diff
@@ -0,0 +1,96 @@
...
+          return this.send(recipient, message, true)
+        }
+        return Promise.reject(new Error(error.message))
+      })
+  }
+
+  // you dont need the message id... it is passed in the payload
+  // you also dont need campaign id..
+  private async generateStatusCallbackUrl(): Promise<string | undefined> {
+    console.log('callbackSecret IS: ', this.callbackSecret)
```
stanleynguyen, line 68: This needs cleanup

shared/src/clients/twilio-client.class/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    console.log('callbackBaseUrl IS: ', this.callbackBaseUrl)
```
stanleynguyen, line 69: This needs cleanup

backend/src/database/migrations/20230216082014-update-transactional-sms.js
stanleynguyen: Eh it just clicked for me what u're talking about. U're concerned that users might see historical records in `UNSENT` state and wonder why right? I think that's ok, if they ask we will can provide the information that we only migrated on the deployment date so all historical messages prior to that are defaulted to UNSENT

backend/src/sms/middlewares/sms-callback.middleware.ts
```diff
@@ -34,6 +34,23 @@ const isAuthenticated = (
...
+const isAuthenticatedTransactional = (
+  req: Request,
+  res: Response,
+  next: NextFunction
+): Response | void => {
+  if (!req.get('authorization')) {
+    res.set('WWW-Authenticate', 'Basic realm="Twilio"')
+    return res.sendStatus(401)
+  }
+  if (
```
tan-zx, line 46: **[nit]** I think it's clearer to use guard clauses to handle all the 4xx cases and leave the happy path at the end

backend/src/database/migrations/20230216082014-update-transactional-sms.js
tan-zx: another alternative is to make the column nullable and make historical messages null.

we should probably should note this behavior somewhere obvious (in the data model?) in case someone else comes along and runs some SQL analysis giving them weird results

backend/src/core/config.ts
```diff
@@ -568,6 +571,13 @@ const config: Config<ConfigSchema> = convict({
       format: 'required-string',
       sensitive: true,
     },
+    callbackUrl: {
```
tan-zx, line 574: **[request for more context]** curious about adding this (haven't traced the code fully myself), this is used for all callbacks and not just tx SMS callbacks right? how did we handle callbacks previously for campaign SMSes/why do we only need to add this now?

tan-zx: ohh saw the changes at `backend/src/sms/routes/sms-callback.routes.ts` → seems like this is only used for tx SMS? if so should make clear in the `doc` field here. 

still wish to get more context on how this setup differs from campaign SMSes' callback though

backend/src/sms/services/sms-callback.service.ts
```diff
@@ -94,7 +114,81 @@ const parseEvent = async (req: Request): Promise<void> => {
     )
   }
 }
+
+// for future context: we only care about when the twilio callback is a finalized status
+// we don't want to update the status pre-emptively until we receive the finalized status
+// initial state --> UNSENT
+const parseTransactionalEvent = async (req: Request): Promise<void> => {
+  const {
+    MessageStatus: twilioMessageStatus,
```
tan-zx, line 123: **[nit/change request]** know this part is consistent with existing code, but is it possible to use types from Twilio's library to make this part type-safe? 

or if that is too much of a hassle, maybe add a comment here to highlight this part is not type-safe and any changes here should be treated with caution

backend/src/database/migrations/20230216082014-update-transactional-sms.js
richard-yak: aaahaha yes! ok if there's no problem then i'll set it to default unsent

richard-yak: done

shared/src/clients/twilio-client.class/index.ts
richard-yak: oops wth

shared/src/clients/twilio-client.class/index.ts
richard-yak: removed

backend/src/sms/services/sms-callback.service.ts
```diff
@@ -94,7 +114,81 @@ const parseEvent = async (req: Request): Promise<void> => {
...
+            errorCode: { [Op.eq]: null },
+          },
+        }
+      )
+    }
+  }
+}
+
+const mapTwilioMessageStatus = (
+  status: string
```
tan-zx, line 176: similarly, would be good if this can be an `enum`, ideally imported from Twilio's library (if it exists there) and if not, it should be `enum` that we define ourselves. this will allow the `switch` case to ensure we have exhausted all the possibilities.

backend/src/sms/services/sms-callback.service.ts
```diff
@@ -94,7 +114,81 @@ const parseEvent = async (req: Request): Promise<void> => {
...
+
+  if (twilioErrorCode) {
+    await SmsMessageTransactional.update(
+      {
+        errorCode: twilioErrorCode,
+        status: TransactionalSmsMessageStatus.Error,
+      } as SmsMessageTransactional,
+      {
+        where: {
+          messageId: messageId,
```
tan-zx, line 144: **[nit]** can simply be `messageId`

backend/src/sms/services/sms.service.ts
```diff
@@ -65,6 +68,7 @@ const getHydratedMessage = async (
  * @param credential
  * @param recipient
  * @param message
+ * @param campaignId
```
tan-zx, line 71: this seems unnecessary? (actually all these JSDocs stuff should be removed since we have TypeScript...but that's for another day)

tan-zx:
Sorry I think my review clashed with your previous commit.

backend/src/sms/services/sms-callback.service.ts
```diff
@@ -94,7 +114,81 @@ const parseEvent = async (req: Request): Promise<void> => {
...
+    case 'delivered':
+      return TransactionalSmsMessageStatus.Delivered
+    case 'accepted':
+      return TransactionalSmsMessageStatus.Accepted
+    default:
+      return TransactionalSmsMessageStatus.Unsent
+  }
+}
 export const SmsCallbackService = {
   isAuthenticated,
```
tan-zx, line 190: **[suggestion for future PR]** should we rename the campaign callback methods to specify they're for campaigns? would be clearer for future maintenance too imo

tan-zx:
Mostly looks good! Just a bunch of nits/code quality issues from me

backend/src/core/config.ts
richard-yak: if we look over to the `worker` codebase, we'd notice that the variable that worker was referring to was the `backend_url` variable. this is simply explicitly labelling the `backend_url` correctly as the callback_url. 

If you scroll slightly up, you'd see that this is already nested under `smsCallback`, as it indicates that it is only used for smsCallbacks.

backend/src/sms/services/sms-callback.service.ts
richard-yak: this is... already an enum?

tan-zx: referring to `status: string`

backend/src/sms/services/sms-callback.service.ts
richard-yak: added the comment.

P.S. there is much to be desired for a cleaner twilio integration but yes, there is only that much refactoring i can do in one PR before it becomes a risk, and better to have a separate PR to properly clean up twilio integration. The first step of moving twilio to shared folder is already gonna be huge in this refactoring should we decide to do it.

backend/src/sms/services/sms-callback.service.ts
richard-yak: removed

backend/src/sms/services/sms.service.ts
richard-yak: removed

backend/src/core/config.ts
tan-zx: Thanks for the clarification! Could you help to fix the `config.ts` [file here](https://github.com/opengovsg/postmangovsg/blob/a0ba537cb5908ae5c8eb5e554ce0313b023a3d79/worker/src/core/config.ts#L228-L233) so that it's aligned to these changes? So what happens in practice is the worker and backend actually are loaded with the same set of env variables, so it'd be good to keep them aligned.

Also can consider whether we should rename `backend_url` as it's specifically the SMS callback URL

backend/src/sms/middlewares/sms-callback.middleware.ts
richard-yak: have updated, take a look

tan-zx: lgtm thanks!

backend/src/sms/services/sms-callback.service.ts
tan-zx: agreed!

backend/src/core/config.ts
richard-yak: woah, ok would recommend not to change the worker file this release yet no matter how small, if this is really imported i will change it in another PR after the release. 

We probably should indeed rename `backend_url` but that would mean changing on secrets manager as well.

backend/src/sms/services/sms-callback.service.ts
richard-yak: same as above comment, let's do this in the twilio refactoring should the day come, it's currently stored as an array of strings so changing to enum right now wouldn't make sense unless we refactor the whole bit.

stanleynguyen (approved):

PR #1938, opened 2023-02-22
feat(backend): api keys database setup

stanleynguyen:
## Problem

Closes [ticket](https://www.notion.so/opengov/Setting-up-api_keys-support-5af9e208c9834481a5bcce135b5ee405?pvs=4)

## Solution

- Create `api_keys` database table
- Implement double write to`api_keys` table

## Deployment Checklist

- [ ] Run DB migration
- [ ] Granting privileges to backend IAM user on the new table and sequence

**Post-deployment**
- [ ] Run `migrate-api-keys` script


backend/src/migrate-api-keys.ts
```diff
@@ -0,0 +1,29 @@
...
+import sequelizeLoader from '@core/loaders/sequelize.loader'
+import { User } from '@core/models'
+import { ApiKey } from '@core/models/user/api-key'
+import { Op } from 'sequelize'
+void (async function main() {
+  await sequelizeLoader()
+  const users = await User.findAll({
+    where: { apiKeyHash: { [Op.not]: null } },
+  })
+  for (const u of users) {
```
richard-yak, line 14: can be optimized if performance is a concern, but if it's a one time script to be ran after hours then no issues

backend/src/database/migrations/20230221073046-create-api-keys-table.js
```diff
@@ -0,0 +1,46 @@
...
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.createTable('api_keys', {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.BIGINT,
+      },
+      user_id: {
```
richard-yak, line 12: probably a worthwhile effort to create index on this col now before the user size api key table grows big.

also remember there was a set active / inactive functionality briefly discussed before, i'm guessing that's no longer required!

richard-yak (approved):
seemsGood, minor optimizations can be considered but likely irrelevant till the table gets huge

backend/src/migrate-api-keys.ts
stanleynguyen: Actually this was intended, cos we don't have a lot of user records to begin with (can be queried within a single query without limit) and we're narrowing down to `api_key_hash not null`

backend/src/database/migrations/20230221073046-create-api-keys-table.js
stanleynguyen: - The indexing piece is interesting 🤔 we might actually need cos in the future we expiring keys
- Yea the expiry, but we're not doing it here yet. This is just for multi API keys first

backend/src/migrate-api-keys.ts
tan-zx: we have 46 active API users haha

richard-yak: oh then defo no issue

tan-zx:
Haven't looked through in detail, but I was thinking the rate-limit should be at the key level, rather than the user level. That way, ICA can have multiple sub-systems, each with their own rate limit.

Doesn't have to be in this PR, but I was thinking an RFC visualising the long-term state of the API key system would be helpful! Then we can decide how to implement these features in parts.

backend/src/migrate-api-keys.ts
```diff
@@ -0,0 +1,29 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
tan-zx, line 1: just curious — this seems like a weird place to put this script, why don't we run it as a database migration and keep track of it with all the other migration scripts there?

tan-zx (approved):
small nit re: placement of migration script, otherwise lgtm

backend/src/migrate-api-keys.ts
stanleynguyen: Cos it's not a schema migration, there's no `down`, it's a one-way-street kinda activity

PR #1943, opened 2023-02-27
feat/sms-api-tests

richard-yak:
Fixes [ticket](https://www.notion.so/opengov/Add-tests-for-SMS-API-PR-55474f71d49247749de5a52403801593?pvs=4)

Providing new test cases that were previously left out in the SMS transactional API release.


backend/src/sms/routes/tests/sms-callback.routes.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      ErrorCode: 'ERRORBOI',
+    }
+    sampleTwilioCallback.ErrorCode = ''
+
+    jest
+      .spyOn(SmsCallbackService, 'isAuthenticatedTransactional')
+      .mockReturnValue(true)
+    let callbackRes = await request(app)
+      .post('/callback/sms')
+      .set('Authorization', `Basic ${1234}`)
```
tan-zx, line 118: **[nit]** can we abstract away the callback credentials? `Basic ${1234}` looks weird haha

backend/src/sms/routes/tests/sms-callback.routes.test.ts
```diff
@@ -0,0 +1,168 @@
...
+    const sampleTwilioCallback = {
+      SmsSid: mockSendMessageResolvedValue,
+      SmsStatus: 'sent',
+      MessageStatus: 'sent',
+      To: '+1512zzzyyyy',
+      MessageSid: mockSendMessageResolvedValue,
+      AccountSid: 'ACxxxxxxx',
+      From: '+1512xxxyyyy',
+      ApiVersion: '2010-04-01',
+      ErrorCode: 'ERRORBOI',
```
tan-zx, line 109: lol

backend/src/sms/routes/tests/sms-callback.routes.test.ts
```diff
@@ -0,0 +1,168 @@
...
+    }
+    sampleTwilioCallback.ErrorCode = ''
```
tan-zx, line 111: **[nit / change request]**

actually just wondering, why not instead of mutating the object, we create new object with different error codes? that way it's clearer.
so what I'd do is:
- default `sampleTwilioCallback` is the happy path one
- new `sampleTwilioCallbackError` contains error code, e.g.

```JS
const sampleTwilioCallbackError = {
      ...sampleTwilioCallback
      MessageStatus: 'failed',
      ErrorCode: 'ERRORBOI',
}
```

think it's easier to follow that mutating the object!

tan-zx (approved):
lgtm! just some nits

backend/src/sms/routes/tests/sms-callback.routes.test.ts
richard-yak: it needs to be basic auth otherwise it'll reject the request, i'll change it to something more readable

backend/src/sms/routes/tests/sms-callback.routes.test.ts
richard-yak: good idea, completely forgot about spread, changed

backend/src/sms/routes/tests/sms-callback.routes.test.ts
tan-zx: Ohh ya I meant put 1234 as a variable. Previously 1234 was in backticks and `${}` but it's not a variable. 

So right now (after your latest commit), it's still a bit weird since we're using backticks but there are no variables. 

Just a nit!

richard-yak: ohh ok i get it, updated

tan-zx (approved):

PR #1944, opened 2023-02-28
fix: remove cypress test temporarily

tan-zx:
## Problem

- Currently, Cypress test is flaky and, combined with the reversion after failure, is causing dev issues
- We also discovered the Cypress test coverage is less than we expected (something to fix soon)

## Solution

- We will be temporarily disabling the Cypress test until we fix the flakiness. We should have a subsequent PR to fix the coverage issue too.


jhibberd-gov:
If we plan to only temporarily remove this logic should we comment it out, rather than deleting it, to make that clear?

jhibberd-gov (approved):

tan-zx:
good point!

PR #1945, opened 2023-02-28
fix: make slack build messages meaningful again

tan-zx:
sorry my bad with the previous PR


tan-zx:
Stanley fixing in separate PR

PR #1951, opened 2023-03-03
chore: log channel provider message id to aid debugging

jhibberd-gov:
Task:
https://www.notion.so/opengov/Ensure-AWS-SES-message_id-are-logged-with-table-primary-key-ID-03254019c59f4a6d92af62a11043c6fc

## Solution
- Message AWS SES ID written to CloudWatch Logs alongside "native" message ID (i.e. DB table primary key value).
- Tested in staging (both campaign and tx).
- [Wrote guide on how to resolve AWS SES ID from native message ID.](https://www.notion.so/opengov/Resolving-AWS-SES-ID-from-Native-Message-ID-c6d47be3edcf433daf1bc8cd66b69c05)
- Confirmed both log groups have sufficient retention (currently infinite).

### DIFF
--- a/backend/src/core/services/auth.service.ts
+++ b/backend/src/core/services/auth.service.ts
@@ -210,10 +210,7 @@ export const InitAuthService = (redisService: RedisService): AuthService => {
    * @param email
    * @param ipAddress originating IP address that requests for OTP.
    */
-  const sendOtp = async (
-    email: string,
-    ipAddress: string
-  ): Promise<string | void> => {
+  const sendOtp = async (email: string, ipAddress: string): Promise<void> => {
     const otp = generateOtp()
     const hashValue = await bcrypt.hash(otp, SALT_ROUNDS)
     const hashedOtp: HashedOtp = {
@@ -224,7 +221,7 @@ export const InitAuthService = (redisService: RedisService): AuthService => {
     await saveHashedOtp(email, hashedOtp)
 
     const appName = config.get('APP_NAME')
-    return MailService.mailClient.sendMail({
+    await MailService.mailClient.sendMail({
       from: config.get('mailFrom'),
       recipients: [email],
       subject: `One-Time Password (OTP) for ${appName}`,
--- a/backend/src/core/services/auth.service.ts
+++ b/backend/src/core/services/auth.service.ts
@@ -11,7 +11,7 @@ import { Transaction } from 'sequelize/types'
 
 export interface AuthService {
   canSendOtp(email: string): Promise<void>
-  sendOtp(email: string, ipAddress: string): Promise<boolean>
+  sendOtp(email: string, ipAddress: string): Promise<string | void>
   verifyOtp(input: VerifyOtpInput): Promise<boolean>
   findOrCreateUser(email: string): Promise<User>
   findUser(id: number): Promise<User>
@@ -213,7 +213,7 @@ export const InitAuthService = (redisService: RedisService): AuthService => {
   const sendOtp = async (
     email: string,
     ipAddress: string
-  ): Promise<boolean> => {
+  ): Promise<string | void> => {
     const otp = generateOtp()
     const hashValue = await bcrypt.hash(otp, SALT_ROUNDS)
     const hashedOtp: HashedOtp = {
--- a/backend/src/core/services/auth.service.ts
+++ b/backend/src/core/services/auth.service.ts
@@ -11,7 +11,7 @@ import { Transaction } from 'sequelize/types'
 
 export interface AuthService {
   canSendOtp(email: string): Promise<void>
-  sendOtp(email: string, ipAddress: string): Promise<string | void>
+  sendOtp(email: string, ipAddress: string): Promise<void>
   verifyOtp(input: VerifyOtpInput): Promise<boolean>
   findOrCreateUser(email: string): Promise<User>
   findUser(id: number): Promise<User>
@@ -221,7 +221,7 @@ export const InitAuthService = (redisService: RedisService): AuthService => {
     await saveHashedOtp(email, hashedOtp)
 
     const appName = config.get('APP_NAME')
-    await MailService.mailClient.sendMail({
+    void MailService.mailClient.sendMail({
       from: config.get('mailFrom'),
       recipients: [email],
       subject: `One-Time Password (OTP) for ${appName}`,
--- a/backend/src/email/services/email.service.ts
+++ b/backend/src/email/services/email.service.ts
@@ -156,14 +156,14 @@ const sendEmail = async (
   opts?: SendEmailOpts
 ): Promise<boolean> => {
   try {
-    const channelProviderMessageId = await MailService.mailClient.sendMail(
+    const serviceProviderMessageId = await MailService.mailClient.sendMail(
       mail,
       opts
     )
     logger.info({
       message: 'Message sent to channel provider.',
       nativeMessageId: mail.messageId,
-      channelProviderMessageId: channelProviderMessageId,
+      serviceProviderMessageId: serviceProviderMessageId,
     })
   } catch (e) {
     logger.error({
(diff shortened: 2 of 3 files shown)


backend/src/core/services/auth.service.ts
```diff
@@ -213,7 +213,7 @@ export const InitAuthService = (redisService: RedisService): AuthService => {
   const sendOtp = async (
     email: string,
     ipAddress: string
-  ): Promise<boolean> => {
+  ): Promise<string | void> => {
```
tan-zx, line 216: I took a look at the [call site here](https://github.com/opengovsg/postmangovsg/blob/b0729fac7a6a2d2de450846ed0239da15f018f5d/backend/src/core/middlewares/auth.middleware.ts#L36). It seems like it is actually not using the return value at all. If the intention for `sendOtp` is not to use the `string` returned all the time, then As such, the return type should be `Promise<void>`.

As for the rest of the function, I personally think it would be clearer to do `void MailService.mailClient.sendMail({..})` followed by `return`.  


```suggestion
): Promise<void> => {
```

tan-zx: then the portion below should be:

```TypeScript
    void MailService.mailClient.sendMail({
      from: config.get('mailFrom'),
      recipients: [email],
      subject: `One-Time Password (OTP) for ${appName}`,
      body: `Your OTP is <b>${otp}</b>. It will expire in ${Math.floor(
        OTP_EXPIRY / 60
      )} minutes.
    Please use this to login to your ${appName} account. <p>If your OTP does not work, please request for a new OTP.</p>
    <p>This login attempt was made from the IP: ${ipAddress}. If you did not attempt to log in to ${appName}, you may choose to investigate this IP address further.</p>
    <p>The ${appName} Support Team</p>`,
    })
    return
```

backend/src/email/services/email.service.ts
```diff
@@ -156,7 +156,15 @@ const sendEmail = async (
   opts?: SendEmailOpts
 ): Promise<boolean> => {
   try {
-    await MailService.mailClient.sendMail(mail, opts)
+    const channelProviderMessageId = await MailService.mailClient.sendMail(
```
tan-zx, line 159: I'm guessing it's intentionally vague? i.e. you didn't want to specify it's AWS SES? not sure if we want to add comments here so it's clearer, or is the idea that you want the code to be agnostic of the actual provider. 

maybe `emailServiceProvider` might be clearer?

shared/src/clients/mail-client.class/index.ts
```diff
@@ -38,8 +38,11 @@ export default class MailClient {
     this.configSet = configSet
   }
 
-  public sendMail(input: MailToSend, option?: SendEmailOpts): Promise<boolean> {
-    return new Promise<boolean>((resolve, reject) => {
+  public sendMail(
+    input: MailToSend,
+    option?: SendEmailOpts
+  ): Promise<string | void> {
+    return new Promise<string | void>((resolve, reject) => {
```
tan-zx, line 45: I think `Promise<string>` makes more sense, if the idea is it should be up to the call-site to decide if they don't want to use the `string` that's returned. if you look at the actual `resolve` clause, it always resolves as a string.

(The code below is actually not type-safe... but that's a battle for another time I guess)

```suggestion
  public sendMail(input: MailToSend, option?: SendEmailOpts): Promise<string> {
    return new Promise((resolve, reject) => {
```

tan-zx (changes requested):

tan-zx:
lgtm mostly just nits!

backend/src/email/services/email.service.ts
jhibberd-gov: Yeah, the name of the module (`email.service.ts`) is non-AWS specific so I felt variable names should reflect this. Although I do prefer your suggestion of at least being concrete about the channel (i.e. `emailProviderMessageId`). Will make that change.

backend/src/email/services/email.service.ts
```diff
@@ -156,7 +156,15 @@ const sendEmail = async (
...
   try {
-    await MailService.mailClient.sendMail(mail, opts)
+    const serviceProviderMessageId = await MailService.mailClient.sendMail(
+      mail,
+      opts
+    )
+    logger.info({
+      message: 'Message sent to channel provider.',
+      nativeMessageId: mail.messageId,
+      serviceProviderMessageId: serviceProviderMessageId,
```
tan-zx, line 166: **[nit]** can be simplified thus:
```suggestion
      serviceProviderMessageId,
```

tan-zx (approved):
lgtm

backend/src/core/services/auth.service.ts
```diff
@@ -11,7 +11,7 @@ import { Transaction } from 'sequelize/types'
 
 export interface AuthService {
   canSendOtp(email: string): Promise<void>
-  sendOtp(email: string, ipAddress: string): Promise<boolean>
+  sendOtp(email: string, ipAddress: string): Promise<string | void>
```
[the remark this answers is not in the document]
jhibberd-gov: Hmm, I was relying on the type system to flag inconsistencies. Wonder why this wasn't flagged?

backend/src/email/services/email.service.ts
jhibberd-gov: Nice :)

backend/src/core/services/auth.service.ts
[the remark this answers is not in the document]
tan-zx: Actually in TypeScript, for application code, it's generally not a great idea to explicitly annotate the return type and better to rely on the type inference, with the exception being if your type is very complex. 

See [this](https://www.youtube.com/watch?v=I6V2FkW1ozQ) for a great explainer.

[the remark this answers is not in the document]
tan-zx: haha this is what I meant by the legacy code in Postman being quite poorly written TypeScript, esp since it was converted from a JavaScript codebase (not helped by the fact that the ppl who did the conversion weren't super familiar with TS imo)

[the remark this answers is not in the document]
tan-zx: the TLDR here is annotating return type allows you to "lie", i.e. it won't throw an error as long as your return type doesn't **violate** the constraint (even if it might be looser than the constraint). See [this excerpt](https://www.youtube.com/watch?v=I6V2FkW1ozQ&t=200s).

This is also why `void` is kinda weird. It doesn't correspond to a type in JS and is closer to telling the callsite "this function might return something, but you're supposed to just ignore it".

backend/src/core/services/auth.service.ts
```diff
@@ -224,7 +221,7 @@ export const InitAuthService = (redisService: RedisService): AuthService => {
     await saveHashedOtp(email, hashedOtp)
 
     const appName = config.get('APP_NAME')
-    return MailService.mailClient.sendMail({
+    await MailService.mailClient.sendMail({
```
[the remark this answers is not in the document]
jhibberd-gov: Hmm, I'm a little confused about the `void` operator. It seems like an esoteric feature of JS that isn't specific to calling async functions where we don't care about the return value? It also seems that Node.js may prematurely terminate async functions that aren't explicitly being awaited?
https://stackoverflow.com/questions/46966890/what-happens-when-a-promise-never-resolves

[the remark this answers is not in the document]
tan-zx: So the NodeJS event loop is a little too complicated to address here ([this](https://www.youtube.com/watch?v=8aGhZQkoFbQ) is the go-to explainer). 

You're right to say that JS's `void` operator (which is distinct from the `void` return type in TS) has a lot of weird history that tbh I'm not super familiar wtih. 

For our purposes, because `sendMail` itself does the error handling and we're not using the return value in this function, it's OK to use `void`. See example discussions [here](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md), [here](https://github.com/Microsoft/TypeScript/issues/13376#issuecomment-273289748), and [here](https://github.com/palantir/tslint/issues/4653). We have also used this in the codebase previously to fulfill this linting rule, and things seem to work fine!

PR #1952, opened 2023-03-03
ci(e2e): add end to end tests using puppeteer

stanleynguyen:
## Problem

e2e test for transactional email endpoint is not working.
We also want to migrate away from Cypress to achieve higher stability in terms of frameworks.

## Solution

Move to puppeteer and use our swaggerUI instead of sending requests straight to backend.

## Deployment Checklist

- [ ] N/A


tan-zx:
just wondering, is the long-term plan to move all e2e tests to puppeteer, or we just moving the tx email API ones

stanleynguyen:
> just wondering, is the long-term plan to move all e2e tests to puppeteer, or we just moving the tx API ones

All tests ser

tan-zx:
do we want to merge this in first and decide on Puppeteer vs Playwright in the future?

so currently the e2e tests we have are:
- Cypress for encrypted email?
- Puppeteer for transactional email?
- no tests for regular emails and SMSes etc.?

a bit messy hor

stanleynguyen:
> do we want to merge this in first and decide on Puppeteer vs Playwright in the future?
> 
> so currently the e2e tests we have are:
> 
> * Cypress for encrypted email?
> * Puppeteer for transactional email?
> * no tests for regular emails and SMSes etc.?
> 
> a bit messy hor

- Cypress for encrypted email and SMSes
- Puppeteer for txn

But we're not running cypress after this change but rather puppeteer. It will be more stable. I'd love to still have some sort of e2e post-deployment to have early signals on bugs, esp major ones

tan-zx:
OK, as long as we've verified the Cypress ones actually work, then we're all good. 

also can we move everything e2e related into the `e2e` folder, think this was a leftover issue from Rae's PR that annoyed me haha

tan-zx (approved):

stanleynguyen:
> OK, as long as we've verified the Cypress ones actually work, then we're all good.

Err actually we're not using cypress again and moving away from it

> 
> also can we move everything e2e related into the `e2e` folder, think this was a leftover issue from Rae's PR that annoyed me haha

yup it's setup that way for puppeteer

PR #1953, opened 2023-03-03
feat: use api key table api records instead

richard-yak:
resolves [ticket](https://www.notion.so/opengov/Update-API-authentication-middleware-to-use-api_keys-table-874cbd2343c34aad85fe20883e205eef?pvs=4)

included
- rewrote logic flow to move all sad path above
- find the key from api key table, then find in user that has that api key

### DIFF
--- a/backend/src/core/services/auth.service.ts
+++ b/backend/src/core/services/auth.service.ts
@@ -3,7 +3,7 @@ import bcrypt from 'bcrypt'
 import { Request } from 'express'
 import config from '@core/config'
 import { loggerWithLabel } from '@core/logger'
-import { User } from '@core/models'
+import { ApiKey, User } from '@core/models'
 import { validateDomain } from '@core/utils/validate-domain'
 import { ApiKeyService, MailService, RedisService } from '@core/services'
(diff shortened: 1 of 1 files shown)


backend/src/core/services/auth.service.ts
```diff
@@ -3,19 +3,25 @@ import bcrypt from 'bcrypt'
...
-import { User } from '@core/models'
+import { ApiKey, User } from '@core/models'
 import { validateDomain } from '@core/utils/validate-domain'
 import { ApiKeyService, MailService, RedisService } from '@core/services'
 import { HashedOtp, VerifyOtpInput } from '@core/interfaces'
 import { Transaction } from 'sequelize/types'
 
 export interface AuthService {
   canSendOtp(email: string): Promise<void>
+
```
tan-zx, line 14: **[nit]** don't think these line breaks are necessary?

tan-zx (approved):
lgtm!

backend/src/core/services/auth.service.ts
richard-yak: removed the line breaks! @tan-zx

PR #1960, opened 2023-03-14
feat: update get settings endpoint

richard-yak:
As per [ticket](https://www.notion.so/opengov/Update-GET-settings-endpoint-to-use-api_keys-bf0496d0131442378542aa7b8afd83fc?pvs=4)

Included in this release:
1. The primary logic of updating existing GET settings endpoint to return the has_api_key flag based on data from api_keys table instead of users table
2. Refactoring of previous change [here](https://github.com/opengovsg/postmangovsg/pull/1953) to use a general service method instead of writing it directly into function, to accommodate future changes with the conditions in case of API key deactivation / expiry

### DIFF
--- a/.github/workflows/non-serverless-deploy.yml
+++ b/.github/workflows/non-serverless-deploy.yml
@@ -93,14 +93,14 @@ jobs:
     uses: ./.github/workflows/deploy-frontend.yml
     secrets: inherit
 
-  e2e-test:
-    needs:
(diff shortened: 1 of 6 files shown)


tan-zx (approved):
lgtm

do we want to add tests within this PR too?

UPDATE: oh nvm it's not part of the DoD, it's ok then

PR #1961, opened 2023-03-14
ci(e2e): setup playwright

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Migrate-e2e-away-from-Cypress-6419b67862b14a79800e0a7c758abaa4?pvs=4)

## Solution

## Deployment Checklist

- [ ] Add secrets in Github Actions for `E2E_API_KEY` (both staging & prod), `E2E_TWILIO_ACC_SID`, `E2E_TWILIO_AUTH_TOKEN` (shared across envs)


e2e/playwright.config.ts
```diff
@@ -0,0 +1,37 @@
+import type { PlaywrightTestConfig } from '@playwright/test';
+
+require('dotenv').config();
+
+const config: PlaywrightTestConfig = {
+  testDir: './tests',
+  /* Maximum time one test can run for. */
+  timeout: 120 * 1000,
```
tan-zx, line 8: any particular reason for the choice of these constants for timeouts, e.g.
- `120 * 1000` (this is 2 mins right)
- `20000` (this is 20 mins?)

e2e/playwright.config.ts
```diff
@@ -0,0 +1,37 @@
...
+  globalSetup: require.resolve('./global-setup'),
+  use: {
+    /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
+    actionTimeout: 0,
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    // baseURL: 'http://localhost:3000',
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: 'on-first-retry',
+  },
```
tan-zx, line 34: Do you want to run the tests for the different browsers too? My thinking is:
- GSIB users probably are on MS Edge (or maybe Chrome)
- OGP users are probably on Chromium and Webkit (e.g. I know Kenneth runs email campaigns on his phone)

This is the config from CheckWho:
```typescript
projects: [
    {
      name: 'chromium',
      use: {
        ...devices['Desktop Chrome'],
      },
    },
    {
      name: 'firefox',
      use: {
        ...devices['Desktop Firefox'],
      },
    },

    {
      name: 'webkit',
      use: {
        ...devices['Desktop Safari'],
      },
    },

    /* Test against mobile viewports. */
    {
      name: 'Mobile Chrome',
      use: {
        ...devices['Pixel 5'],
      },
    },
    {
      name: 'Mobile Safari',
      use: {
        ...devices['iPhone 12'],
      },
    },

    /* Test against branded browsers. */
    {
      name: 'Microsoft Edge', // browser used by public officers
      use: {
        channel: 'msedge',
      },
    },
  ],
```

tan-zx (dismissed):
lgtm, just a couple of comments

e2e/playwright.config.ts
stanleynguyen: - 120s cos our test might run for as long as 2 minutes
- 20k is 20s

e2e/playwright.config.ts
stanleynguyen: I don't think these browser compat is big enough an issue to revert a deployment, so not intending to test on lower-tier ones

tan-zx: so which are the browsers we are actually testing on? I actually think we should prioritise Edge + Chrome on PC (since that's what our users use)

tan-zx:
Btw I know you're removing the [Cypress README](https://github.com/opengovsg/postmangovsg/blob/f2f1d29b564f9313b2a4fdf564bf0d839333588e/cypress/README.md#L1) in [another PR](https://github.com/opengovsg/postmangovsg/pull/1965), but would be good if you could write a similar Playwright README in this PR.

That way, I can set up Playwright by following the instructions, then I can test the other PRs by running them locally.

Thanks.

e2e/playwright.config.ts
stanleynguyen: The default one is Chromium

tan-zx (approved):

PR #1964, opened 2023-03-15
ci(e2e): add test case for sms in playwright

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Migrate-e2e-away-from-Cypress-6419b67862b14a79800e0a7c758abaa4?pvs=4)

## Solution

## Deployment Checklist

- [ ] N/A

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -51,6 +51,9 @@ jobs:
           # Reusing env var from cypress
           # TODO: rename this once we've migrated away from Cypress
           API_KEY: ${{ secrets.CYPRESS_API_KEY }}
+          TWILIO_ACC_SID: ${{ secrets.CYPRESS_TWILIO_ACC_SID }}
+          TWILIO_AUTH_TOKEN: ${{ secrets.CYPRESS_TWILIO_AUTH_TOKEN }}
+          SMS_NUMBER: "+6582410222"
           MAILBOX: internal-testing@open.gov.sg
         working-directory: e2e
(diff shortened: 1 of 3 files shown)


tan-zx:
Btw I know you're removing the [Cypress README](https://github.com/opengovsg/postmangovsg/blob/f2f1d29b564f9313b2a4fdf564bf0d839333588e/cypress/README.md#L1) in [another PR](https://github.com/opengovsg/postmangovsg/pull/1965), but would be good if you could write a similar Playwright README in this PR.

That way, I can set up Playwright by following the instructions, then I can test the other PRs by running them locally.

Thanks.

.github/workflows/e2e.yml
```diff
@@ -51,6 +51,9 @@ jobs:
           # Reusing env var from cypress
           # TODO: rename this once we've migrated away from Cypress
           API_KEY: ${{ secrets.CYPRESS_API_KEY }}
+          TWILIO_ACC_SID: ${{ secrets.CYPRESS_TWILIO_ACC_SID }}
+          TWILIO_AUTH_TOKEN: ${{ secrets.CYPRESS_TWILIO_AUTH_TOKEN }}
+          SMS_NUMBER: "+6582410222"
```
richard-yak, line 56: is this your own number?!

richard-yak (approved):

.github/workflows/e2e.yml
stanleynguyen: Nope, a number that we bought on twilio for this purpose

PR #1967, opened 2023-03-17
feat: deleting api key feature with tests

richard-yak:
As per [ticket](https://www.notion.so/opengov/Delete-an-API-key-endpoint-735bc9991ea94e6c8c4da891e15c67c3?pvs=4)

Included in this release:
1. Created new apikey route for new apikey specific endpoints. 
2. Introduce endpoint to delete API key
3. Integration tests for the new endpoint

Deployment Checklist
- [x] Remove Staging PPT test 
- [x] Test on Staging
- [x] Add back Staging PPT test

### DIFF
--- a/backend/src/core/middlewares/api-key.middleware.ts
+++ b/backend/src/core/middlewares/api-key.middleware.ts
@@ -1,38 +1,33 @@
 import { loggerWithLabel } from '@core/logger'
-import { NextFunction, Request, Response } from 'express'
+import { Request, Response } from 'express'
 import { ApiKeyService } from '@core/services'
 
 const logger = loggerWithLabel(module)
 
 const deleteApiKey = async (
   req: Request,
-  res: Response,
-  next: NextFunction
+  res: Response
 ): Promise<Response | void> => {
-  try {
-    const { apiKeyId } = req.params
-    const userId = req.session?.user?.id
+  const { apiKeyId } = req.params
+  const userId = req.session?.user?.id
 
-    const deletedRows = await ApiKeyService.deleteApiKey(
-      userId.toString(),
-      +apiKeyId
-    )
-    if (deletedRows <= 0) {
-      return res.status(404).json({
-        code: 'not_found',
-        message: `Could not find API key to delete`,
-      })
-    }
-    logger.info({
-      message: 'Successfully deleted api key',
-      action: 'deleteApiKey',
+  const deletedRows = await ApiKeyService.deleteApiKey(
+    userId.toString(),
+    +apiKeyId
+  )
+  if (deletedRows <= 0) {
+    return res.status(404).json({
+      code: 'not_found',
+      message: `Could not find API key to delete`,
     })
-    return res.status(200).json({
-      api_key_id: apiKeyId,
-    })
-  } catch (err) {
-    return next(err)
   }
+  logger.info({
+    message: 'Successfully deleted api key',
+    action: 'deleteApiKey',
+  })
+  return res.status(200).json({
+    api_key_id: apiKeyId,
+  })
 }
 
 export const ApiKeyMiddleware = {
--- a/backend/src/core/middlewares/api-key.middleware.ts
+++ b/backend/src/core/middlewares/api-key.middleware.ts
@@ -1,50 +1,40 @@
 import { loggerWithLabel } from '@core/logger'
-import { Handler, NextFunction, Request, Response } from 'express'
-import { ApiKeyService, AuthService } from '@core/services'
+import { NextFunction, Request, Response } from 'express'
+import { ApiKeyService } from '@core/services'
 
-export interface ApiKeyMiddleware {
-  deleteApiKey: Handler
-}
-export const InitApiKeyMiddleware = (authService: AuthService) => {
-  const logger = loggerWithLabel(module)
+const logger = loggerWithLabel(module)
 
-  const deleteApiKey = async (
-    req: Request,
-    res: Response,
-    next: NextFunction
-  ): Promise<Response | void> => {
-    try {
-      // enforce authentication by cookie
-      const authenticated = authService.checkCookie(req)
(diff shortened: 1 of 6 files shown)


backend/src/core/middlewares/api-key.middleware.ts
```diff
@@ -0,0 +1,50 @@
...
+    res: Response,
+    next: NextFunction
+  ): Promise<Response | void> => {
+    try {
+      // enforce authentication by cookie
+      const authenticated = authService.checkCookie(req)
+      if (!authenticated) {
+        return res.sendStatus(403).json({
+          message: `Invalid cookie`,
+        })
```
tan-zx, line 22: **[change request]** this part is a bit weird, cmiiw but shouldn't authentication be handled at the route level (i.e. in `backend/src/core/routes/index.ts`), rather than here? 

reasons for this:
- at the route level, you can clearly distinguish which routes are meant to accept which modes of validation (so in this case, we only accept cookie)
- clearer separation of concerns

backend/src/core/middlewares/api-key.middleware.ts
```diff
@@ -0,0 +1,50 @@
...
+      }
+
+      const { apiKeyId } = req.params
+      const userId = req.session?.user?.id
+
+      const count = await ApiKeyService.deleteApiKey(
+        userId.toString(),
+        +apiKeyId
+      )
+      if (count <= 0) {
```
tan-zx, line 32: **[question/nit]**
- can `count` be less than zero?
- took a look at the API, seems like it returns the number of destroyed rows, maybe would be good to rename the variable to something clearer

backend/src/core/middlewares/api-key.middleware.ts
```diff
@@ -0,0 +1,50 @@
...
+    req: Request,
+    res: Response,
+    next: NextFunction
+  ): Promise<Response | void> => {
+    try {
+      // enforce authentication by cookie
+      const authenticated = authService.checkCookie(req)
+      if (!authenticated) {
+        return res.sendStatus(403).json({
+          message: `Invalid cookie`,
```
stanleynguyen, line 21: add field `code: unauthorized` as we're getting to [this](https://www.notion.so/API-Responses-in-JSON-5174d66bf5ce49a1b3d161630869eb65) soon

backend/src/core/middlewares/api-key.middleware.ts
```diff
@@ -0,0 +1,50 @@
...
+      const { apiKeyId } = req.params
+      const userId = req.session?.user?.id
+
+      const count = await ApiKeyService.deleteApiKey(
+        userId.toString(),
+        +apiKeyId
+      )
+      if (count <= 0) {
+        return res.sendStatus(404).json({
+          message: `Could not find API key to delete`,
```
stanleynguyen, line 34: add field `code: not_found` as we're getting to [this](https://www.notion.so/API-Responses-in-JSON-5174d66bf5ce49a1b3d161630869eb65) soon

backend/src/core/middlewares/api-key.middleware.ts
```diff
@@ -0,0 +1,50 @@
...
+        })
+      }
+      logger.info({
+        message: 'Successfully deleted api key',
+        action: 'deleteApiKey',
+      })
+      return res.sendStatus(200)
```
stanleynguyen, line 41: return a json object `{ api_key_id: apiKeyId}` to follow [this proposal](https://www.notion.so/opengov/API-Responses-in-JSON-5174d66bf5ce49a1b3d161630869eb65)

backend/src/core/middlewares/api-key.middleware.ts
```diff
@@ -0,0 +1,50 @@
...
+    } catch (err) {
+      return next(err)
```
stanleynguyen, line 43: not necessary as this is the default behavior of express middleware already

backend/src/core/middlewares/api-key.middleware.ts
richard-yak: good point, changing next push

backend/src/core/middlewares/api-key.middleware.ts
richard-yak: renamed

backend/src/core/middlewares/api-key.middleware.ts
richard-yak: this part has been removed, using authMiddleware now

backend/src/core/middlewares/api-key.middleware.ts
richard-yak: updated

backend/src/core/middlewares/api-key.middleware.ts
richard-yak: done

backend/src/core/middlewares/api-key.middleware.ts
richard-yak: removed

backend/src/core/middlewares/api-key.middleware.ts
```diff
@@ -0,0 +1,35 @@
...
+  res: Response
+): Promise<Response | void> => {
+  const { apiKeyId } = req.params
+  const userId = req.session?.user?.id
+
+  const deletedRows = await ApiKeyService.deleteApiKey(
+    userId.toString(),
+    +apiKeyId
+  )
+  if (deletedRows <= 0) {
```
tan-zx, line 18: **[nit]** same as previous nit, if it cannot be `< 0`, why don't we just do `deletedRows === 0`

tan-zx (approved):
lgtm

backend/src/core/middlewares/api-key.middleware.ts
richard-yak: updated

PR #1968, opened 2023-03-20
build(Dockerfiles): Refactor dockerfiles for backend and worker

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Optimizing-Dockerfiles-af57454eef2d4e90bc5e59233a32bfef?pvs=4)

## Solution

Switched to multi-step build and add cache to docker build on GHA:
- Reduced the image size by 20%
- Docker build time down from 2m47s (without cache) to 10s (with full cache) or more realistically 1m30s (without differing application code)

## Deployment Checklist

- [ ] N/A


backend/Dockerfile
```diff
@@ -1,4 +1,19 @@
...
+FROM public.ecr.aws/docker/library/node:18-alpine as builder
+WORKDIR /usr/home/postmangovsg
+
+COPY shared ../shared
+RUN cd ../shared && npm ci
+
+COPY ./package* ./
+RUN npm ci
+
+COPY tsconfig.* ./
```
tan-zx, line 10: trivial, but `tsconfig` should come before the others, since it's less likely to change?

worker/Dockerfile
```diff
@@ -1,4 +1,20 @@
...
+FROM public.ecr.aws/docker/library/node:18-alpine as builder
+WORKDIR /usr/home/postmangovsg
+
+COPY shared ../shared
+RUN cd ../shared && npm ci
+
+COPY ./package* ./
+RUN npm ci
+
+COPY tsconfig.json ./
```
tan-zx, line 10: > trivial, but `tsconfig` should come before the others, since it's less likely to change?

same as prev

tan-zx (commented):
honestly, I don't really understand Docker, so can't really review this PR — have left some best effort comments below

look forward to discussing irl

richard-yak (commented):
same here, im not too familiar with docker in general. Perhaps @hanu-ogp can consider weighing in?

backend/Dockerfile
```diff
@@ -13,24 +28,25 @@ RUN aws configure set default.region ap-southeast-1
...
+EXPOSE 4000
+COPY ./docker-entrypoint.sh ./
+RUN ["chmod", "+x", "docker-entrypoint.sh"]
 
-COPY tsconfig.json ./
-COPY tsconfig.build.json ./
-COPY openapi.yaml ./
-COPY src ./src
+COPY --from=builder /usr/home/shared/package* ../shared/
+RUN cd ../shared && npm install
```
hanu-ogp, line 41: Do we need npm install again if the folder is copied from the builder step ?

backend/Dockerfile
```diff
@@ -13,24 +28,25 @@ RUN aws configure set default.region ap-southeast-1
 
 WORKDIR /usr/home/postmangovsg
 
-COPY shared ../shared
-RUN cd ../shared && npm ci
+# this means npm will only install production deps, and optimizations in the node
+# engine will take place as well
+ENV NODE_ENV=production
```
hanu-ogp, line 33: shouldnt this step be in the builder phase ?

worker/Dockerfile
```diff
@@ -12,22 +28,24 @@ RUN aws configure set default.region ap-southeast-1
...
-RUN npm ci
+# expose port and copy these files first as they're unlikely to change
+EXPOSE 4000
+COPY ./docker-entrypoint.sh ./
+RUN ["chmod", "+x", "docker-entrypoint.sh"]
 
-COPY src ./src
-COPY tsconfig.json ./
+COPY --from=builder /usr/home/shared/package* ../shared/
+RUN cd ../shared && npm install
```
hanu-ogp, line 41: ditto as above

backend/Dockerfile
stanleynguyen: Nope, we only copy package files

backend/Dockerfile
stanleynguyen: Nope, we need devDependencies in the builder stage to build the app, plus we want this env var set in the final container as well

worker/Dockerfile
stanleynguyen: Nope, we only copy package files

tan-zx (approved):
lgtm thanks for the changes 

hope to learn more about multi stage builds and layer caching when we meet irl

backend/Dockerfile
```diff
@@ -1,4 +1,19 @@
...
+COPY tsconfig.* ./
+COPY ./package* ./
+RUN npm ci
+
+COPY src ./src
+
+RUN npm run build
+RUN cd ../shared && npm run postbuild
+
+FROM public.ecr.aws/docker/library/node:18-alpine
```
tan-zx, line 16: swap order with comment

backend/Dockerfile
tan-zx, line 33: two purposes for setting this
1. you will only download dependencies and not devDependencies
2. you will run in optimised mode

worker/Dockerfile
```diff
@@ -1,4 +1,20 @@
...
+COPY ./package* ./
+RUN npm ci
+
+COPY src ./src
+
+RUN npm run build
+RUN cd ../shared && npm run postbuild
+
+FROM public.ecr.aws/docker/library/node:18-alpine
+# postman public ecr
```
tan-zx, line 17: swap comment order

PR #1969, opened 2023-03-21
feat: added endpoint to get api key list

richard-yak:
As per [ticket](https://www.notion.so/opengov/List-API-keys-endpoint-af0d9d82748b4c38bdd4f3fca5a5bf64?pvs=4)

Included in this PR:
- Building on previous PR after middleware and services are initiated
- Creating a new endpoint with api key routing to decouple from user settings
- Implement API Key list endpoint, sorted by created date DESC, only accessible by cookie
- Test cases for this use case

### DIFF
--- a/backend/src/core/middlewares/api-key.middleware.ts
+++ b/backend/src/core/middlewares/api-key.middleware.ts
@@ -4,6 +4,14 @@ import { ApiKeyService } from '@core/services'
 
 const logger = loggerWithLabel(module)
 
+const listApiKeys = async (
+  req: Request,
+  res: Response
+): Promise<Response | void> => {
+  const userId = req.session?.user?.id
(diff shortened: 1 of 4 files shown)


backend/src/core/middlewares/api-key.middleware.ts
```diff
@@ -4,6 +4,14 @@ import { ApiKeyService } from '@core/services'
 
 const logger = loggerWithLabel(module)
 
+const listApiKeys = async (
+  req: Request,
+  res: Response
+): Promise<Response | void> => {
```
tan-zx, line 10: actually the `Promise<Response | void>` return signature is a bit weird, but it's a larger problem with the codebase, so might as well just stick with it for now.

tan-zx (approved):
lgtm!

PR #1970, opened 2023-03-21
docs(README): add instruction to setup webhooks to local

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Local-callbacks-setup-d3a78470cf8b41798c0b1951e463422a?pvs=4)

## Solution

Write docs

## Deployment Checklist

- [ ] N/A


richard-yak (approved):
yummy

PR #1971, opened 2023-03-22
feat: support soft delete for api key

richard-yak:
As per [slack discussion](https://opengovproducts.slack.com/archives/C049QPXT27M/p1679459330284049?thread_ts=1679389658.985529&cid=C049QPXT27M)

Included in this release:
- Migration file to add deleteAt column
- Paranoid: true on api_keys model 
- Updating of existing test cases to test soft delete

Deployment Checklist

- [x] Running migration file for prod db

### DIFF
--- a/backend/src/core/routes/tests/api-key.routes.test.ts
+++ b/backend/src/core/routes/tests/api-key.routes.test.ts
@@ -13,7 +13,7 @@ beforeAll(async () => {
 })
 
 afterEach(async () => {
-  await ApiKey.destroy({ where: {} })
+  await ApiKey.destroy({ where: {}, force: true })
   await User.destroy({ where: {} })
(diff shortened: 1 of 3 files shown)


backend/src/core/routes/tests/api-key.routes.test.ts
```diff
@@ -13,7 +13,7 @@ beforeAll(async () => {
 })
 
 afterEach(async () => {
-  await ApiKey.destroy({ where: {} })
+  await ApiKey.destroy({ where: {}, force: true })
```
tan-zx, line 16: ah this is required because otherwise a paranoid model will only do soft delete

tan-zx (approved):
lgtm

PR #1974, opened 2023-03-24
fix(amplify): use node version 16

stanleynguyen:
## Problem

AML2 image (default image of amplify) doesn't have the underlying required glibc version to run node 18

## Solution

- Remove repo-wide requirement to use node 18 (i.e. removing .nvmrc in root)
- Force use node v16 in amplify

## Deployment Checklist

- [ ] N/A

### DIFF
--- a/amplify.yml
+++ b/amplify.yml
@@ -13,7 +13,7 @@ frontend:
   phases:
     preBuild:
       commands:
-        - nvm install
+        - nvm install 16
         - if [ "${AWS_BRANCH}" = "master" ]; then
           export REACT_APP_BACKEND_URL=${BACKEND_URL_PRODUCTION} &&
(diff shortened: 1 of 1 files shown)


tan-zx:
can downgrade the `.nvmrc` to Node 16 instead of deleting?

it's helpful for me to quickly switch Node versions when I check out different repos haha

stanleynguyen:
> can downgrade the `.nvmrc` to Node 16 instead of deleting? it's helpful for me to quickly switch Node versions when I check out different repos haha
> 
> so does this mean we'll use Node 18 in the Docker images (backend + worker), then Node 16 for the frontend?

The requirement is only on amplify actually, so yea actually we can keep nvm

tan-zx (approved):
ok lgtm thanks

PR #1976, opened 2023-03-24
fix: Twilio invalid phone number error handling

tan-zx:
Still testing on local sth wrong hm let me try on staging


backend/src/sms/middlewares/sms-transactional.middleware.ts
```diff
@@ -91,9 +93,11 @@ async function sendMessage(
       action,
     })
 
-    const BAD_REQUEST_ERRORS = [TemplateError, InvalidRecipientError]
-    if (BAD_REQUEST_ERRORS.some((errType) => err instanceof errType)) {
-      res.status(400).json({ message: (err as Error).message })
+    if (err instanceof InvalidPhoneNumberError) {
+      res.status(400).json({
+        code: 'invalid_phone_number',
```
stanleynguyen, line 98: I think maybe make it a bit more generic `invalid_recipient`? so we can use the same code for other endpoints too

stanleynguyen (approved):

backend/src/sms/routes/tests/sms-transactional.routes.test.ts
```diff
@@ -128,42 +127,6 @@ describe('POST /transactional/sms/send', () => {
     mockSendMessage.mockReset()
   })
 
-  test('Should return a HTTP 400 when template is invalid', async () => {
```
tan-zx, line 131: removing because there is no "invalid template" for SMSes

backend/src/sms/routes/tests/sms-transactional.routes.test.ts
```diff
@@ -128,42 +127,6 @@ describe('POST /transactional/sms/send', () => {
...
-      .post('/transactional/sms/send')
-      .set('Authorization', `Bearer ${apiKey}`)
-      .send(validApiCall)
-
-    expect(res.status).toBe(400)
-    expect(mockSendMessage).toBeCalledTimes(1)
-    mockSendMessage.mockReset()
-  })
-
-  test('Should return a HTTP 400 when recipient is not valid', async () => {
```
tan-zx, line 149: invalid recipient error not caught at this point, handled separately, so this test is misleading

stanleynguyen: Eh this is actually possible

tan-zx: fixed [here](https://github.com/opengovsg/postmangovsg/compare/fix/twilio-error-2?expand=1), ps

PR #1979, opened 2023-03-27
refactor(config): use APP_ENV instead of NODE_ENV

stanleynguyen:
## Problem

Under the latest optimization #1968 , `NODE_ENV` is defaulted to `production` so that we only install production node_modules and also signal to node engines to carry out [optimizations](https://nodejs.dev/en/learn/nodejs-the-difference-between-development-and-production/) (and we want to have this in staging as well for env parity).
However, some of our application code relies on this `ENV` for business logic.

## Solution

Switching to use `APP_ENV` instead of `NODE_ENV`

**Note**: We don't switch these for serverless functions (e.g. redaction-digest, and unsubscribe-digest) as env vars these are set manually through GHA and the performance improvement is probably insignificant

## Deployment Checklist

- [ ] set `APP_ENV` in our secret manager for the app

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -259,14 +259,14 @@
         "filename": "backend/src/test-utils/test-env.ts",
         "hashed_secret": "c237a19676ad55d8904be354990dd54f92f9572c",
         "is_verified": false,
-        "line_number": 8
+        "line_number": 9
       },
       {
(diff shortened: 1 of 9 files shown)


tan-zx (approved):
lgtm but backend tests are failing hm

richard-yak (approved):

PR #1983, opened 2023-03-28
feat: add tag and classification to tx email

tan-zx:
[Ticket](https://www.notion.so/opengov/fd466213f133486d85e4117f1ded521d?v=94cf83ccaa9c4fa4a15235cdf10ba9f4&p=193debfcb61948b98e8e7b5422c8568c&pm=s)

Todo:
- [x] Test on local 
- [x] Test on staging

Deployment
- [x] Run db migration first
- [x] Deploy

Note to reviewers: the DoD doesn't include tests. If we decide to permanently add this feature, we can add tests I guess.

### DIFF
--- a/backend/src/email/routes/email-transactional.routes.ts
+++ b/backend/src/email/routes/email-transactional.routes.ts
@@ -7,7 +7,10 @@ import {
 } from '@email/middlewares'
 import { fromAddressValidator } from '@core/utils/from-address'
 import { FileAttachmentMiddleware } from '@core/middlewares'
-import { TransactionalEmailMessageStatus } from '@email/models'
+import {
+  TransactionalEmailMessageStatus,
+  TransactionalEmailClassification,
+} from '@email/models'
 import { TransactionalEmailSortField } from '@core/constants'
 import config from '@core/config'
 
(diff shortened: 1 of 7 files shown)


backend/src/email/routes/email-transactional.routes.ts
```diff
@@ -44,6 +47,11 @@ export const InitEmailTransactionalRoute = (
         .options({ convert: true })
         .lowercase(),
       attachments: Joi.array().items(Joi.object().keys().required()),
+      classification: Joi.string()
+        .uppercase()
+        .valid(...Object.values(TransactionalEmailClassification))
+        .optional(),
+      tag: Joi.string().max(255).optional(),
```
tan-zx, line 54: strictly speaking I think the `.optional()` flag is itself [optional by default](https://joi.dev/api/?v=17.8.3#:~:text=Values%20(or%20keys%20in%20case%20of%20objects)%20are%20optional%20by%20default), but I think it's good to include to make it clearer

stanleynguyen (changes requested):
Can we also update the openapi doc?

tan-zx:
Yessir sorry I forgot about that

stanleynguyen (approved):

PR #1985, opened 2023-03-29
feat(backend): add endpoint to generate new api key

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Generate-API-key-endpoint-39e2e9a06c7f44e083599e06fc4c8abe?pvs=4)

## Solution

Implement the endpoint.
- [x] Tested to make sure that newly created API key works with existing auth check

## Deployment Checklist

- [ ] N/A

### DIFF
--- a/backend/src/core/middlewares/api-key.middleware.ts
+++ b/backend/src/core/middlewares/api-key.middleware.ts
@@ -1,44 +1,66 @@
 import { loggerWithLabel } from '@core/logger'
-import { Request, Response } from 'express'
-import { ApiKeyService } from '@core/services'
+import { Handler, Request, Response } from 'express'
+import { ApiKeyService, CredentialService } from '@core/services'
 
 const logger = loggerWithLabel(module)
 
-const listApiKeys = async (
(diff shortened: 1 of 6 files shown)


richard-yak (approved):

tan-zx (commented):
pending discussion, but lgtm otherwise

tan-zx (approved):

tan-zx:
small suggestion: do you want to flag [here](https://github.com/opengovsg/postmangovsg/blob/ae7695550630fab7c45e7a20664d7b965576be00/backend/src/core/services/credential.service.ts#L357) to say that this method will be deprecated in a subsequent PR?

stanleynguyen:
> small suggestion: do you want to flag [here](https://github.com/opengovsg/postmangovsg/blob/ae7695550630fab7c45e7a20664d7b965576be00/backend/src/core/services/credential.service.ts#L357) to say that this method will be deprecated in a subsequent PR?

It's ok, this will be done this sprint as well

PR #1988, opened 2023-04-03
feat(frontend): multi API key support for API settings page

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Multi-API-key-support-for-API-settings-page-3577b7a7f4414ddca3c4104a3daa7507?pvs=4)

## Solution

Implement [the design](https://www.figma.com/file/29FkTqWnA4yLKzstnvyY6X/Design-Master-V1?t=NU4IZNYNKd6Dktsm-0)

## Deployment Checklist

- [x] Update API guide doc

### DIFF
--- a/backend/src/core/middlewares/api-key.middleware.ts
+++ b/backend/src/core/middlewares/api-key.middleware.ts
@@ -11,7 +11,7 @@ export interface ApiKeyMiddleware {
   generateApiKey: Handler
 }
 
-export const InitApikeyMiddleware = (
+export const InitApiKeyMiddleware = (
   credentialService: CredentialService
 ): ApiKeyMiddleware => {
   const listApiKeys = async (
--- a/frontend/src/components/dashboard/settings/api-key/ApiKey.tsx
+++ b/frontend/src/components/dashboard/settings/api-key/ApiKey.tsx
@@ -1,155 +1,127 @@
 import cx from 'classnames'
 
-import { useState, useRef, useContext, useEffect } from 'react'
+import { useState, useContext, useEffect } from 'react'
 
 import type { FunctionComponent } from 'react'
 
 import styles from './ApiKey.module.scss'
 
-import {
-  TextInputWithButton,
-  ConfirmModal,
-  StepHeader,
-} from 'components/common'
+import { CopyModal } from './CopyModal'
+import { CreateModal } from './CreateModal'
+
+import { ConfirmModal, PrimaryButton } from 'components/common'
 import { ModalContext } from 'contexts/modal.context'
 
-import { regenerateApiKey } from 'services/settings.service'
+import {
+  ApiKey as ApiKeyType,
+  deleteApiKey,
+  listApiKeys,
+} from 'services/api-key.service'
 
-const RESET_COPY_TIMEOUT = 1000
+const ApiKey: FunctionComponent = () => {
+  const [apiKeys, setApiKeys] = useState<ApiKeyType[]>([])
+  function addApiKey(key: ApiKeyType) {
+    setApiKeys([key, ...apiKeys])
+  }
+  function removeApiKey(id: string) {
+    setApiKeys(apiKeys.filter((k) => k.id !== id))
+  }
 
-interface ApiKeyProps {
-  hasApiKey: boolean
-  onGenerate?: () => void
-}
+  const [isPageLoading, setIsPageLoading] = useState<boolean>(false)
 
-enum ApiKeyState {
-  GENERATE = 'GENERATE',
-  COPY = 'COPY',
-  COPIED = 'COPIED',
-  REGENERATE = 'REGENERATE',
-}
-
-const ApiKey: FunctionComponent<ApiKeyProps> = ({ hasApiKey, onGenerate }) => {
-  const [apiKey, setApiKey] = useState('')
-  const [errorMsg, setErrorMsg] = useState('')
-  const [apiKeyState, setApiKeyState] = useState<ApiKeyState>(
-    ApiKeyState.GENERATE
-  )
+  useEffect(() => {
+    void (async () => {
+      setIsPageLoading(true)
+      setApiKeys(await listApiKeys())
+      setIsPageLoading(false)
+    })()
+  }, [])
 
   const modalContext = useContext(ModalContext)
-  const apiKeyRef = useRef<HTMLInputElement>()
 
-  useEffect(() => {
-    setApiKey(hasApiKey ? '*'.repeat(30) : '')
-    setApiKeyState(hasApiKey ? ApiKeyState.REGENERATE : ApiKeyState.GENERATE)
-  }, [hasApiKey])
-
-  useEffect(() => {
-    if (apiKeyState !== ApiKeyState.COPIED) return
-
-    const timeoutId = setTimeout(() => {
-      setApiKeyState(ApiKeyState.COPY)
-
-      if (!apiKeyRef.current) return
-      apiKeyRef.current.blur()
-    }, RESET_COPY_TIMEOUT)
-
-    return () => clearTimeout(timeoutId)
-  }, [apiKeyState])
-
-  async function onButtonClick() {
-    switch (apiKeyState) {
-      case ApiKeyState.GENERATE:
-      case ApiKeyState.REGENERATE:
-        if (apiKey) {
-          modalContext.setModalContent(
-            <ConfirmModal
-              title="Are you sure?"
-              subtitle="Generating a new API key will revoke your current one."
-              buttonText="Confirm"
-              onConfirm={onGenerateConfirm}
-            />
-          )
-        } else {
-          await onGenerateConfirm()
-        }
-        break
-      case ApiKeyState.COPY:
-        if (!apiKeyRef.current) return
-        apiKeyRef.current.select()
-        document.execCommand('copy')
-        setApiKeyState(ApiKeyState.COPIED)
-        break
-      default:
-        break
-    }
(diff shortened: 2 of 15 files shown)


backend/src/core/middlewares/api-key.middleware.ts
```diff
@@ -1,44 +1,66 @@
...
 
-  const deletedRows = await ApiKeyService.deleteApiKey(
-    userId.toString(),
-    +apiKeyId
-  )
-  if (deletedRows == 0) {
-    return res.status(404).json({
-      code: 'not_found',
-      message: `Could not find API key to delete`,
+export const InitApikeyMiddleware = (
```
tan-zx, line 13: **[nit]** `InitApiKeyMiddleware`

backend/src/core/services/credential.service.ts
```diff
@@ -64,6 +64,11 @@ export interface CredentialService {
     userId: number,
     announcementVersion: string | null
   ): Promise<{ announcementVersion: string }>
+
+  generateApiKey(
+    userId: number,
+    label: string
+  ): Promise<ApiKey & { key: string }>
```
tan-zx, line 71: **[nit]** 
- I had to trace the code to realise that `key` is the API key in plaintext
- would prefer a more informative variable name 🤔️ maybe `apiKeyPlaintext`

frontend/src/components/dashboard/settings/api-key/ApiKey.tsx
```diff
@@ -1,155 +1,127 @@
...
+              <th className="lg">Key name</th>
+              <th className="sm">Last 5 Digits</th>
+              <th className="sm"></th>
+            </tr>
+          </thead>
+          <tbody>
+            {apiKeys.map((k) => (
+              <tr key={k.id}>
+                <td className="lg">{k.label}</td>
+                <td className="sm">••••• {k.lastFive}</td>
```
tan-zx, line 111: **[question]** just curious, the space between the `•` and the `{k.lastFive}` is an aesthetic decision?

this is how it currently looks:

image


this is how it looks without the space:

image

tan-zx (changes requested):
just want to say this is a great job!! really awesome to be able to support multiple API keys.

I have a few nits + change requests. will review the frontend code in more detail tomorrow.

some other suggestions:
- [ ] not sure if the DoD includes tests 
- [ ] before this is deployed, we should coordinate and update the [API Guide](https://guide.postman.gov.sg/developer-guide/api-doc), which contains screenshots that would be obsoleted by this PR)

****
## API key aesthetics/security

- currently, because of how we're generating the API key, the last character is always `=`, which is a padding character in base64. this is because:
  -  we are generating 32 bytes 
  - each character is used to represent 6 bits (log2(64) = 6).
  - therefore 4 chars are used to represent 4 * 6 = 24 bits = 3 bytes.
  - the number of characters to represent 32 bytes = 4 chars of 3 bytes each *10 +  3 chars (which is 18 bits, i.e. more than enough to represent 16 bits, i.e. 2 bytes) 
- it's pointless from a security-perspective and quite unhelpful as a "last-5-digit" identifier to have a redundant character in our API key
- suggest that we get rid of this useless `=` by changing `32` to `33` [here](https://github.com/opengovsg/postmangovsg/blob/22ef05af491796d590dd951dc33d96573d2ab253/backend/src/core/services/api-key.service.ts#L11)
  - in that case, to represent 33 bytes, we need exactly 44 characters (which is the current length of our API key)

## API key list
SCR-20230403-tp7

"You can create an API key to access our programmatic email and programmatic SMS APIs. For more information, go to our [API Guide](https://guide.postman.gov.sg/developer-guide/api-doc)."

The current copy is simultaneously too vague (doesn't tell you what the API keys are for) and too specific (custom integrations, automate tasks etc.) and irrelevant (keep data more secure??). I think it's better to link to the guide, where we can properly elaborate on what can be done with API keys.

## Create API key 
SCR-20230403-tuc

Changes:
- "Key Label"
- "The label you choose for your API key..."
- "Label your key"

Reasoning:
- Consistent with how we're calling it in code
- No need to switch between "name" and "reference name"

## Delete API key
SCR-20230403-tuf

"Deleting your API key is immediate and irreversible. To ensure a smooth API key rotation, please replace this key with a new API key before deleting."

frontend/src/components/dashboard/settings/api-key/CopyModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const CopyModal = ({ apiKey }: { apiKey: ApiKey }) => {
+  const [hasCopied, setHasCopied] = useState<boolean>(false)
+  return (
+    <div className={styles.modal}>
+      <div className={styles.modalImg}>
+        <img src={successImg} alt="Success image" />
+      </div>
+      <h2 className={styles.title}>
+        Your API Key has been successfully created
```
tan-zx, line 17: **[nit]** `key` instead of `Key`

frontend/src/components/dashboard/settings/api-key/CopyModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+          }}
+          buttonLabel={
+            <>
+              {hasCopied ? 'Copied' : 'Copy'} API key
+              <i className={`bx ${hasCopied ? 'bx-check' : 'bx-copy'}`}></i>
+            </>
+          }
+        />
+        <InfoBlock className={styles.infoBlock}>
+          Please save a copy your API key immediately as this will only be shown
```
tan-zx, line 40: `this will only be shown once`

`If you leave or refresh this page, the API key will be no longer be accessible.`

("hidden" suggests that we kept a copy of it somewhere, which is not true)

frontend/src/components/dashboard/settings/api-key/CreateModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+        })()
+      }}
+    >
+      <div>
+        <h2>
+          <label htmlFor="apiKeyLabel">Create a new API key</label>
+        </h2>
+        <h4>Key Name</h4>
+        <p>
+          The name you choose for your API key will be permanent and cannot be
```
tan-zx, line 45: `is permanent`

tan-zx (commented):
more nits

frontend/src/components/dashboard/settings/api-key/ApiKey.tsx
```diff
@@ -1,155 +1,127 @@
...
+  function openDeleteModal(keyId: string) {
+    const apiKey = apiKeys.find((k) => k.id === keyId)
+    if (!apiKey) return
+    modalContext.setModalContent(
+      <ConfirmModal
+        title={`Are you sure you want to delete API key "${apiKey.label}"?`}
+        subtitle="Deleting your API key is irreversible, and any integrations using this key will need to be updated with a new API key to continue working."
+        onConfirm={() => {
+          removeApiKey(keyId)
+          void deleteApiKey(keyId)
```
richard-yak, line 66: if deleteApiKey api call fails, UI will still show that API key was removed which might be misleading, perhaps we should put removeApiKey after a successful deleteApiKey call

richard-yak (approved):
seemsGood, minor suggestion on state management of delete flow

stanleynguyen:
@tan-zx kudos on the smart suggestion on the api key generation mechanism

@tan-zx @richard-yak revised, please re-review

richard-yak (approved):
yummy

frontend/src/components/dashboard/settings/api-key/ApiKey.tsx
```diff
@@ -1,155 +1,135 @@
...
+          </thead>
+          <tbody>
+            {apiKeys.map((k) => (
+              <tr key={k.id}>
+                <td className="lg">{k.label}</td>
+                <td className="sm">••••• {k.last_five}</td>
+                <td className={cx('sm', styles.buttonContainer)}>
+                  {/* <button>
+                <i className={cx('bx bx-pencil', styles.pencil)} />
+              </button> */}
```
tan-zx, line 123: **[question]** why is this commented out and not just deleted ah?

frontend/src/components/dashboard/settings/api-key/CreateModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+        </h2>
+        <h4>Key Label</h4>
+        <p>
+          The label you choose for your API key is be permanent and cannot be
+          changed once it has been created.
+        </p>
+        <TextInput
+          id="apiKeyLabel"
+          type="text"
+          placeholder="The label that you can identify your API key with"
```
tan-zx, line 50: **[nit]** Sorry missed this earlier, strictly grammatical English should be: `The label with which you can identify your API key`

tan-zx (approved):
lgtm, just a couple of comments/nits

frontend/src/components/dashboard/settings/api-key/ApiKey.tsx
stanleynguyen: cos we're using it in the next iteration so I'm leaving there haha. During compilation these will be removed from the final code anyway

tan-zx: > cos we're using it in the next iteration so I'm leaving there haha. During compilation these will be removed from the final code anyway

coolz sgtm

let's coordinate with @kendra-wong to update the guide before merging into `master` (am aware this one is just merging into `feat/generate-api-key`)

stanleynguyen: I'll update the guide

PR #1991, opened 2023-04-05
test(e2e): remove email check from txn email test

stanleynguyen:
## Problem

Test txm emails don't arrive fast enough at our test mailbox smh, rendering our test time-out

## Solution

Considering that we might have an async model for sending out txn emails (i.e. non-guaranteed delivery latency), let's remove the check for email to stabilise our test case

## Deployment Checklist

- [ ] N/A


stanleynguyen:
[Successful run](https://github.com/opengovsg/postmangovsg/actions/runs/4616800802/jobs/8162567446)

richard-yak (approved):

tan-zx (approved):
lgtm

PR #1993, opened 2023-04-09
Feat/phonebook preview

hanu-ogp:
## Problem

Integrating phonebook with postman emails and SMS

* Invokes phonebook api to create the user and userchannels from the worker
* Only email and SMS campaigns
* Error handling to ensure Phonebook errors doesnt stop postman from delivering messages. Circuit breaker to also take care of ignoring phonebook in case of multiple failures
* Enabling the feature is controlled by an ENV variable (default = false)
* When switched on, will contact phonebook for every campaign. Phonebook will decide the eligible experiment participants and will send the contact pref link if the campaign user is part of the experiment.
* Mustasche takes care of rendering the snippet only if the link is available

## Deployment Checklist

NA


richard-yak:
remember to add into the deployment checklist

- [ ] to populate secrets on aws prior to deployment
- [ ] to deploy phonebook code
- [ ] run migrations (if any) 
- [ ] etc

worker/src/core/config.ts
```diff
@@ -310,6 +315,23 @@ const config: Config<ConfigSchema> = convict({
...
+      doc: 'Phonebook uri to fetch contact preferences',
+      default: 'phonebook.postman.gov.sg',
+      env: 'PHONEBOOK_URL',
+    },
+    apiKey: {
+      doc: 'API key for Phonebook contact preferences api',
+      default: '',
+      env: 'PHONEBOOK_API_KEY',
+    },
+  },
```
richard-yak, line 334: after reviewing the use case, i think we should use `required-string` format here

hanu-ogp:
> remember to add into the deployment checklist
> 
> * [ ]  to populate secrets on aws prior to deployment
> * [ ]  to deploy phonebook code
> * [ ]  run migrations (if any)
> * [ ]  etc

@richard-yak The config has the default value. Technically they should work even without them being created in aws. Nevertheless will create them. And the flag is disabled, so even if phonebook is unavailable its ok as the flow will never be invoked until the flag is true

worker/src/core/loaders/message-worker/util/contact-preference.ts
```diff
@@ -0,0 +1,119 @@
...
+    })
+}
+
+const breaker = new CircuitBreaker(getContactPrefLinks, options)
+
+export const getContactPrefLinksForEmail = async (
+  result: EmailResultRow[],
+  campaignId: number,
+  campaignOwnerEmail: string,
+  channel = 'Email'
```
richard-yak, line 40: beware the inconsistency, channel in postman codebase  as `campaign_type` enum label is `EMAIL`, just making sure that this is intentional

worker/src/core/loaders/message-worker/util/contact-preference.ts
```diff
@@ -0,0 +1,119 @@
...
+    userChannels,
+    postmanCampaignId: campaignId,
+    postmanCampaignOwner: campaignOwnerEmail,
+  }
+  return breaker
+    .fire(request)
+    .then((resp) => {
+      const userChannels = resp as ContactChannel[]
+      return map(result, (message) => {
+        const contactPreference = userChannels.find(
```
richard-yak, line 105: map + find operation here is O(n^2). 

is it possible to let userChannels return as map format so that we can reduce the complexity to O(n) by using key-value access.

richard-yak (changes requested):
overally seems to be ok, not too sure about circuit breaker as it is new to me, but i trust that the failsafe has been tested! 

Just need to check up on some potential issues.

richard-yak:
we shouldn't have `apiKey` as a default value in config.ts i don't think, unless `apiKey` is not required for making a request..?

tan-zx:
Will look at the code more closely later, but curious about this design:

> When switched on, will contact phonebook for every campaign. Phonebook will decide the eligible experiment participants and will send the contact pref link if the campaign user is part of the experiment.

Curious what's the thinking behind this, why not do this determination from within Postman itself?

hanu-ogp:
> Will look at the code more closely later, but curious about this design:
> 
> > When switched on, will contact phonebook for every campaign. Phonebook will decide the eligible experiment participants and will send the contact pref link if the campaign user is part of the experiment.
> 
> Curious what's the thinking behind this, why not do this determination from within Postman itself?

To avoid making changes to postman (including env variables) whenever I need to add/remove agencies for the experiment. Even if phonebook is down or slow, circuit breaker will take care of fault tolerance and ensure the campaign is delivered with max delay of 5secs

hanu-ogp:
> we shouldn't have `apiKey` as a default value in config.ts i don't think, unless `apiKey` is not required for making a request..?

Cant define a required config without having a default value - https://github.com/mozilla/node-convict/issues/29

worker/src/core/loaders/message-worker/util/contact-preference.ts
```diff
@@ -0,0 +1,119 @@
...
+    .post(url, JSON.stringify(request), {
+      headers: {
+        'Content-Type': 'application/json',
+        'x-api-key': config.get('phonebookContactPref.apiKey'),
+      },
+    })
+    .then((resp) => resp.data)
+    .catch((err) => {
+      throw err
+    })
```
stanleynguyen, line 31: THis should be rewritten to use await instead

worker/src/core/loaders/message-worker/util/contact-preference.ts
```diff
@@ -0,0 +1,119 @@
...
+        return {
+          ...row.message,
+          showMasthead,
+          contactPrefLink: contactPreference?.contactPrefLink || '',
+        }
+      })
+    })
+    .catch((error) => {
+      throw error
+    })
```
stanleynguyen, line 74: Similar to above, this should be rewritten to use await instead

worker/src/core/loaders/message-worker/util/contact-preference.ts
```diff
@@ -0,0 +1,119 @@
...
+      channelId: message.recipient,
+    }
+  })
+  const request = {
+    userChannels,
+    postmanCampaignId: campaignId,
+    postmanCampaignOwner: campaignOwnerEmail,
+  }
+  return breaker
+    .fire(request)
```
stanleynguyen, line 101: Similar to above, rewritten to use await instead

worker/src/core/loaders/message-worker/sms.class.ts
```diff
@@ -60,12 +61,47 @@ class SMS {
...
+            type: QueryTypes.SELECT,
+          }
+        )
+        if (!emailResult || emailResult.length === 0) {
+          throw new Error(
+            'Unable to fetch user email from campaign for phonebook contact preference api'
+          )
+        }
+        const userEmail = emailResult[0].email
+        return await getMessagesWithContactPrefLinks(
```
stanleynguyen, line 90: nit: unnecessary await at return statement

worker/src/core/loaders/message-worker/email.class.ts
```diff
@@ -59,18 +60,42 @@ class Email {
...
+            type: QueryTypes.SELECT,
+          }
+        )
+        if (!emailResult || emailResult.length === 0) {
+          throw new Error(
+            'Unable to fetch user email from campaign for phonebook contact preference api'
+          )
+        }
+        const userEmail = emailResult[0].email
+        return await getContactPrefLinksForEmail(result, campaignId, userEmail)
```
stanleynguyen, line 89: nit: unnecessary await at return statement

worker/src/core/loaders/message-worker/util/contact-preference.ts
hanu-ogp: Yes it is intentional

worker/src/core/loaders/message-worker/util/contact-preference.ts
hanu-ogp: Updated it to compute the map here so the api can be cleaner - input and output are same format for api

richard-yak (approved):

worker/src/core/loaders/message-worker/sms.class.ts
```diff
@@ -60,12 +61,43 @@ class SMS {
...
+      } catch (error) {
+        logger.error({
+          message: 'Unable to fetch contact preferences',
+          error,
+          workerId: this.workerId,
+        })
+        // If phonebook is down, we still want to continue sending the messages
+      }
+    }
+    return result
```
tan-zx, line 100: **[nit]** I think it's clearer to move this `return result` up, i.e.`if (!showContactPref || result.length === 0) return result`. you remove one level of nesting and the code is easier to read

worker/src/core/loaders/message-worker/util/contact-preference.ts
```diff
@@ -0,0 +1,113 @@
...
+  const userChannelsResp = response as ContactChannel[]
+  const userChannelMap = userChannelsResp.reduce(
+    (map: Map<string, ContactChannel>, contactChannel: ContactChannel) => {
+      map.set(contactChannel.channelId, contactChannel)
+      return map
+    },
+    new Map<string, ContactChannel>()
+  )
+
+  return map(result, (row) => {
```
tan-zx, line 60: curious why do you choose to use the `map` from `lodash` instead of the built-in `map`, i.e. `result.map(...)`

worker/src/core/loaders/message-worker/util/contact-preference.ts
```diff
@@ -0,0 +1,113 @@
...
+  const userChannelsResp = response as ContactChannel[]
+  const userChannelMap = userChannelsResp.reduce(
+    (map: Map<string, ContactChannel>, contactChannel: ContactChannel) => {
+      map.set(contactChannel.channelId, contactChannel)
+      return map
+    },
+    new Map<string, ContactChannel>()
+  )
+
+  return map(result, (message) => {
```
tan-zx, line 105: same here

> curious why do you choose to use the `map` from `lodash` instead of the built-in map, i.e. `result.map(...)`

from a reader's perspective, slightly confusing since you are also using the `Map()` construct

tan-zx (commented):
some nits, mostly lgtm

would recommend thoroughly testing on staging to make sure the circuit breaker works as expected before merging in. for example, to make your tests on staging closer to production, you might want to:
- increase the number of sending workers on staging (iinw there is only 1 on staging normally)
- make sure there are multiple large campaigns of different types going on at the same 

would recommend you write up a testing plan to make sure things are working as intended

worker/src/core/loaders/message-worker/util/contact-preference.ts
stanleynguyen: my wild guess is that this is reference to the existing code in other part of the workers

worker/src/core/loaders/message-worker/util/contact-preference.ts
stanleynguyen: my wild guess is that this is reference to the existing code in other part of the workers

stanleynguyen (approved):

worker/src/core/loaders/message-worker/util/contact-preference.ts
hanu-ogp: Await will break the flow, only then/catch will work with the breaker

PR #1996, opened 2023-04-11
feat: sync GitBook with repo

tan-zx:
Create `guide` branch to allow two-way sync of [our guide](https://guide.postman.gov.sg/) with our repository

# Suggested SOP for maintaining guide

##  Editing markdown file to sync back to GitBook

- If review is necessary, we can create a new branch (e.g. `docs/update-readme`) and target the `guide` branch in the PR. A preview will be generated on the PR and reviewers can comment before merging in and locking in the changes.
- If review is not necessary, we can create a commit onto the `guide` branch directly.
- Generally, editing the markdown file is preferred for significant content changes (since GitBook's web UI can be limiting)

## Making changes via GitBook's UI
- GitBbook will automatically sync and add the changes as a commit to the `guide` branch. 
- GitBook UI is good for making structural changes via the UI (reorganising pages etc.)
- Also, product team will probably use the UI to make changes

# Alternatives considered

## Syncing to `master` directly 

Pros:
- Code changes and guide changes can be made in a single PR

Cons:
- Any documentation-only changes will trigger a new deployment, which needs to be cancelled manually (don't think there is an easy way to disable this)

## Syncing `guide` with `master`

Pros:
- We can see the documentation and the rest of the codebase at the same time without switching branches.

Cons:
- Keeping `guide` and `master`, which is a pain. Similarly, when merging `guide` into `master`, we need to cancel deployment manually.

### Code changes:

To implement the alternatives, following modification to our CI/CD is necessary:
- [x] add `gitbook` to `.prettierignore` at root folder
- [x] turn off CI/CD for `guide` branch (need to use `contains(github.head_ref, 'feature')` syntax, using `${{ github.ref != 'refs/heads/guide' }}` doesn't work for some reason)

Not needed under current setup


stanleynguyen:
cmiiw, if we create a draft in gitbook, gitbook will open a PR in our repo right?

tan-zx:
I had to do [this](https://github.com/opengovsg/postmangovsg/pull/1996/commits/d175e15a2cf92f305a17788ba36e6a425ad06a40) to silence `detect-secrets`, but now [the guide](https://guide.postman.gov.sg/campaign-guide/quick-start/password-protected-email) looks like this:

image

let me see if there's some other way to fix this...

tan-zx:
OK, so I think our SOP for using this is:

## If we're just editing the guide:
- Editing markdown directly: we can create commits and merge into `guide` directly → the changes will show up automatically
- Same is true for changes made via the UI: Gitbook will automatically sync and add the commits (see above

## If we're making changes to the guide as part of a PR
1. PR will be merged into `master` → this leads to new deployment
2. Second PR should be opened that (1) first rebases `guide` against `master`; (2) makes the same changes in the first PR, so that the guide would be updated. 

@stanleynguyen let me know if this makes sense. 

I thought about syncing to `master` directly, but that doesn't make sense because then any changes product makes to the guide will trigger a new deployment (though there's probably a way around that?)

probably need more thought

tan-zx:
Not sure why [this](https://github.com/opengovsg/postmangovsg/pull/1996/commits/a019315bf6e45459b2bef806d421d6f6a4fd2188) didn't work, gonna rest now

tan-zx:
Actually, it just occurred to me that maybe there's no need to merge this into `master` at all. 

Rather, the guide can simply live in its own branch and, if we ever make changes to the documentation, we can just open a PR against this branch. 

If the above makes sense, I'm gonna close this PR and add the documentation somewhere else instead!

PR #1998, opened 2023-04-12
build: run container as non-root user

tan-zx:
[Why we should run container as non-root](https://docs.bitnami.com/tutorials/why-non-root-containers-are-important-for-security)

**Next steps of debugging**
- [ ] Run container locally and run command to see what exactly the permission issue is

- [ ] Test to make sure it works on staging
- [ ] Update `.yml` files to parameterize (instead of hard-code) Node version?


stanleynguyen (approved):

stanleynguyen:
Can fix the worker one similar to what I did for backend docker-entry, then we can move `dotenv` from prod dependencies to devDependencies for both apps

tan-zx:
> Can fix the worker one similar to what I did for backend docker-entry, then we can move `dotenv` from prod dependencies to devDependencies for both apps

OK I continue tmr

stanleynguyen:
Actually can u try this one instead `eval "export $(echo "$(aws secretsmanager get-secret-value --secret-id '@SECRET_ID' --query SecretString --output text)"   | jq -r 'to_entries | map("\(.key)=\(.value)") | @sh')"`

tan-zx:
> Actually can u try this one instead `eval "export $(echo "$(aws secretsmanager get-secret-value --secret-id '@SECRET_ID' --query SecretString --output text)" | jq -r 'to_entries | map("\(.key)=\(.value)") | @sh')"`

[still failing](https://github.com/opengovsg/postmangovsg/actions/runs/4777015309/jobs/8492498707) 😢

PR #1999, opened 2023-04-12
build(localstack): remove localstack

tan-zx:
After discussing with @richard-yak and @stanleynguyen, we don't really need `localstack` to mock AWS services, can use AWS directly, so might as well remove it.

Have also [disabled revert-on-e2e-failure on staging](https://github.com/opengovsg/postmangovsg/pull/1999/commits/3435f0f98b1db5dc469af0fc45d4edeeba7df970)

- [x] test on staging


README.md
```diff
@@ -54,56 +54,14 @@
...
-# Create the database
-createdb postmangovsg_dev
-
-# Check if you can connect to the database
-psql -h localhost -p 5432 postmangovsg_dev
-
-# Install redis
-brew install redis
-brew services start redis
-
```
richard-yak, line 85: these ones cannot remove, must have to run the local server

richard-yak (changes requested):
the redis and sql local server will be used, cannot remove

README.md
tan-zx: I'm just removing the instructions where Redis and PostgreSQL are run natively.

Still keeping the Dockerfile for running these in the Docker container.

image

richard-yak (approved):

stanleynguyen (approved):

PR #2000, opened 2023-04-13
feat: Init whatsapp service and accompanying placeholder code.

richard-yak:
As per [whatsapp](https://www.notion.so/opengov/RFC-Whatsapp-Implementation-b16ba611aced40c1baf52224169de836?pvs=4) epic.

This release has proven:

- [ ] We are able to send whatsapp messages via API.
- [ ] We can parse whatsapp API responses

Included in this release:

- [ ] New configs to be used by whatsapp service
- [ ] Placeholder files to be filled in along the way
- [ ] New whatsapp routes and services
- [ ] the core whatsapp wrapper to enable posting to graph API

Please ignore empty files / files with code in them just to enable them to be exported without errors :)

### DIFF
--- a/backend/src/core/config.ts
+++ b/backend/src/core/config.ts
@@ -699,28 +699,34 @@ const config: Config<ConfigSchema> = convict({
   whatsapp: {
     endpointUrl: {
       doc: 'Whatsapp endpoint URL',
-      default: '',
+      default: 'graph.facebook.com',
       env: 'WHATSAPP_ENDPOINT_URL',
+      format: 'required-string',
     },
     endpointVersion: {
       doc: 'Whatsapp current version',
-      default: '',
+      default: '16.0',
       env: 'WHATSAPP_ENDPOINT_VERSION',
+      format: 'required-string',
     },
     callbackUrl: {
       doc: 'Whatsapp Callback URL',
-      default: '',
+      default: '/v1/callback/whatsapp',
       env: 'WHATSAPP_CALLBACK_URL',
+      format: 'required-string',
     },
     callbackVerifyToken: {
       doc: 'Whatsapp Callback Verify Token',
-      default: '',
+      default: 'sample-token',
       env: 'WHATSAPP_CALLBACK_VERIFY_TOKEN',
+      format: 'required-string',
     },
     bearerToken: {
       doc: 'Whatsapp Permanent Access Token',
       default: '',
       env: 'WHATSAPP_BEARER_TOKEN',
+      format: 'required-string',
+      sensitive: true,
     },
   },
 })
--- a/backend/src/core/config.ts
+++ b/backend/src/core/config.ts
@@ -153,6 +153,13 @@ interface ConfigSchema {
     maxAttachmentSize: number
     maxAttachmentNum: number
   }
+  whatsapp: {
+    endpointUrl: string
+    endpointVersion: number
+    callbackUrl: string
+    callbackVerifyToken: string
+    bearerToken: string
+  }
 }
 
 convict.addFormats({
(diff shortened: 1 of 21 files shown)


backend/src/core/config.ts
```diff
@@ -689,6 +696,33 @@ const config: Config<ConfigSchema> = convict({
       format: Number,
     },
   },
+  whatsapp: {
```
stanleynguyen, line 699: Shouldn't all of these below be `required-string`?

backend/src/whatsapp/middlewares/whatsapp-callback.middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+  req: Request,
+  res: Response,
+  next: NextFunction
+): Promise<Response | void> => {
+  try {
+    await WhatsappCallbackService.parseEvent(req)
+    logger.info({ message: 'Successfully parsed event', action: 'parseEvent' })
+    return res.sendStatus(200)
+  } catch (err) {
+    return next(err)
```
stanleynguyen, line 17: no need to do this, this is the default behavior

backend/src/whatsapp/middlewares/whatsapp.middleware.ts
```diff
@@ -0,0 +1,53 @@
...
+    const { recipient, from, content } = req.body
+    try {
+      const resp = await WhatsappService.sendMessage(from, recipient, content)
+      return res.json(resp)
+    } catch (e) {
+      logger.error({
+        message: 'Something went wrong with sending a message',
+        error: e,
+      })
+      return res.status(401).json({ message: e })
```
stanleynguyen, line 45: Shouldn't bee 401, should be a 500 if WA is down or something (catastrophic), or codes depending the specific errors thrown from WA

shared/src/clients/whatsapp-client.class/index.ts
```diff
@@ -0,0 +1,73 @@
...
+      method: 'POST', // default method will be post
+      host: this.baseUrl,
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${this.bearerToken}`,
+      },
+    }
+
+    return new Promise((resolve, reject) => {
+      const req = https.request({ ...defaultOptions, ...options }, (res) => {
```
stanleynguyen, line 28: this is a bit masochistic lol, can we use the existing axios installation?

shared/src/clients/whatsapp-client.class/index.ts
```diff
@@ -0,0 +1,73 @@
...
+      void resolution.then((res) => {
+        // graphAPI docs state that even for errors, it would return status 200 but have an error object
+        // so we check for error object instead of http status.
+        if (!res.error) {
+          resolve(res)
+        } else {
+          reject(res)
+        }
+      })
+    })
```
stanleynguyen, line 71: This could be replaced with `await`-ing line 52 `this.request` and then dealing with the awaited res object

backend/src/whatsapp/services/whatsapp.service.ts
```diff
@@ -0,0 +1,32 @@
...
+    recipient = PhoneNumberService.normalisePhoneNumber(
+      recipient,
+      config.get('defaultCountry')
+    ).substring(1)
+    // strip out the plus sign afterwards
+  } catch (err) {
+    throw new InvalidRecipientError('Invalid phone number')
+  }
+
+  const client = initializeBasicClient()
```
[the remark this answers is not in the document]
richard-yak: updated to use the same instance from the service level (doing it this way vs initializing directly in the client class as worker will need to instantiate their own instance)

PR #2002, opened 2023-04-17
feat(backend): add API key expiration field

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/DB-Migration-to-add-valid_util-and-notification_emails-a9e9460aa4894386b643a6eaef7d699c?pvs=4)

## Solution

Implement it and default existing ones to be valid until 21st Apr 2024 (1 year from expected release)

## Deployment Checklist

- [x] Migrate DB


richard-yak (approved):

backend/src/database/migrations/20230408151021-add-api-key-expiry-and-contact.js
```diff
@@ -0,0 +1,21 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('api_keys', 'valid_until', {
+      type: Sequelize.DataTypes.DATE,
+      allowNull: false,
+      defaultValue: new Date('2024-04-21T00:00:00.000Z'),
```
tan-zx, line 8: Do we want to use a default time that is more intuitive to users in Singapore?

tan-zx: E.g. 2024-04-20 11:59pm GMT+8

backend/package.json
```diff
@@ -53,6 +53,7 @@
     "libphonenumber-js": "1.10.6",
     "lodash": "4.17.21",
     "module-alias": "2.2.2",
+    "moment": "^2.29.4",
```
tan-zx, line 56: by convention, think we're freezing the dependencies (realise I didn't do that for `yamljs` below oops)

tan-zx (approved):

backend/src/database/migrations/20230408151021-add-api-key-expiry-and-contact.js
tan-zx: actually, now that I look at it `21 April 2024, 8am` seems pretty intuitive to me too

image

backend/package.json
stanleynguyen: Will freeze both

tan-zx: thanks boss

PR #2003, opened 2023-04-17
feat: display API key expiry date and notification addresses

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Display-API-key-expiration-date-acdf0b3cc6874010ae78a28598435044?pvs=4)

## Solution

Implement it duh

## Deployment Checklist

- [ ] N/A


richard-yak (approved):

backend/src/core/middlewares/api-key.middleware.ts
```diff
@@ -54,10 +55,33 @@ export const InitApiKeyMiddleware = (
     const userId = req.session?.user?.id
     const apiKey = await credentialService.generateApiKey(
       +userId,
-      req.body.label
+      req.body.label,
+      req.body.notification_addresses
     )
     return res.status(201).json(convertApiKeyModelToResponse(apiKey))
   }
+  const updateApiKey = async (
```
tan-zx, line 63: **[question]** so right now, this only updates notification addresses, but in the future, this endpoint will be used to update the other stuff too right (permissions? anything else?)

backend/src/core/models/user/api-key.ts
```diff
@@ -29,4 +29,8 @@ export class ApiKey extends Model<ApiKey> {
 
   @Column({ type: DataType.DATE, allowNull: false })
   validUntil: Date
+
+  // we're enforcing not-null only for new keys via application logic
+  @Column({ type: DataType.ARRAY(DataType.STRING), allowNull: true })
+  notificationAddresses: string[]
```
tan-zx, line 35: **[question]** in the future, if we expand to non-email-based methods of sending notifications (e.g. SMS), we can just add a new column right

tan-zx: **[nit]** suggest renaming this to `notificationEmailAddresses`

frontend/src/components/dashboard/settings/api-key/CreateUpdateModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+          id="apiKeyLabel"
+          type="text"
+          placeholder="The label with which you can identify your API key with"
+          value={keyLabel}
+          onChange={setKeyLabel}
+          disabled={isUpdateMode}
+        ></TextInput>
+        <h4>Contact Emails</h4>
+        <p>
+          Please provide the email contacts in the event we need to reach out to
```
tan-zx, line 132: **[copy]** Please enter email addresses that will receive updates regarding this API key (e.g. reminders to rotate expiring key, information about new features etc.). It is recommended to keep these email addresses updated so that critical updates will not be missed.

tan-zx (commented):
I haven't gone through the frontend code in detail. I will poke around deeper when I have a time to review the code 

Here are some nits/edge cases I noticed:


### Text in button

Somehow, while filling up the info to create a brand new API key, I get `Updating Key`. 
SCR-20230417-hr2

Probably a state management issue. Ideally, I think it should say `Generate Key` in this context.

### Repeated email addresses

OK this one is pretty lame, but do we want to guard against duplicate emails

image

backend/src/core/middlewares/api-key.middleware.ts
stanleynguyen: yes

backend/src/core/models/user/api-key.ts
stanleynguyen: Oh no, I'm thinking we will be using the same column, that's why I named it `notificationAddresses` only (who knows we might include physical addresses too. JKJK lmao)

tan-zx: in that case maybe `notificationContacts`? anyway small issue — just the db column harder to rename in the future, but no biggie

tan-zx (approved):
lgtm

PR #2005, opened 2023-04-18
feat(backend): add error code for api validation errors

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Update-response-for-validation-errors-4389a3c1631549eaa39cb6eff6e8c823?pvs=4)

## Solution

Custom-handle celebrate error and transform the default returned error to one that includes `code`

## Deployment Checklist

- [ ] N/A


richard-yak (approved):

tan-zx:
in terms of deployment, we should inform our users first before deploying right? let's add that to the deployment checklist and coordinate with @kendra-wong?

stanleynguyen:
> in terms of deploying this branch, we should inform our users first before deploying right? let's add that to the deployment checklist and coordinate with @kendra-wong?

Yup we might want to inform our users first because we're excluding the field `fields` (automatically added by celebrate) from the response

PR #2007, opened 2023-04-18
hotfix: priotise showing reschedule header for scheduled campaigns

richard-yak:
[Ticket](https://www.notion.so/opengov/Incorrect-waitTime-estimation-for-scheduled-campaign-89b5ff2d44bd4f7bb3433c81601d2588?pvs=4)

This solution simply moves the check for schedule sending up top, as scheduled messages do not directly 'join the sending queue', so the wait time logic is not applicable to scheduled campaigns. 

Another uglier solution considered was to change the wait time conditional to `campaign.status !== Status.Scheduled && stats.waitTime && stats.waitTime > 0`.


tan-zx (approved):
lgtm

PR #2009, opened 2023-04-19
announcement modal to gather interest for whatsapp

lisatjide:
We want to gather interest from agencies who wants to send messages through WhatsApp.


richard-yak (approved):
pending physical verification, then will deploy

PR #2010, opened 2023-04-21
feat: new attachment limits setup

tan-zx:
[Ticket](https://www.notion.so/opengov/Coordinate-change-in-attachment-number-and-attachment-size-13ba040a5f5942b38f8219867e96a6bc)

In this PR:
- Increase number of attachments to 10 (from 5)
- Set a global file attachment limit to 10MB
- Increase Beanstalk's nginx `client_max_body_size` to 11MB (this makes sense since max body size is 1MB + 10MB of attachments → in practice, unlikely to hit this precise limit, user will see our custom error message instead)

- [x] Test on staging
- [x] Update `openapi.yaml`

### DIFF
--- a/backend/.platform/nginx/conf.d/client_max_body_size.conf
+++ b/backend/.platform/nginx/conf.d/client_max_body_size.conf
@@ -1,1 +1,1 @@
-client_max_body_size    10M;
+client_max_body_size    11M;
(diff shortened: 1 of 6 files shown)


richard-yak (approved):
looks pretty straightforward, remember to add the config.ts changes to secrets manager in ur deployment checklist!

tan-zx:
> looks pretty straightforward, remember to add the config.ts changes to secrets manager in ur deployment checklist!

haha no need, we're just using the default values

stanleynguyen (approved):

PR #2012, opened 2023-04-24
feat: models to be used for whatsapp

richard-yak:
Lines mainly due to migration files

This PR includes
- The minimum required models to run the integration, with reference to pre-existing campaign data models
- A new schema `domain_credentials` which allows us to save credentials on domain level

Additional context
- You will notice that the relationship between campaign vs whatsapp template specifically differs from existing implementations. Existing implementation has template creation along with campaign creation, with a one-to-one relationship. Whatsapp templates are pre-created and pre-approved on the Whatsapp Business Manager, and will have a many-to-one relationship with campaigns, held via the foreign key table `campaign_whatsapp_templates`


richard-yak:
PR has been up for 3 weeks, merging as assumed reviewed

PR #2013, opened 2023-04-24
refactor(backend): refactor general errors declaration

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Update-responses-for-general-errors-d3d2bb802aa542f2850c83333bf78db0?pvs=4)

## Solution

Refactor it to use the shared REST API error types

## Deployment Checklist

- [ ] N/A


richard-yak (approved):
lgtm, will we be updating all the endpoints from try catch to using this style of error capturing?

PR #2015, opened 2023-04-25
build(ci): update deprecated GitHub Actions dependencies

tan-zx:
Get rid of warnings in GitHub Actions. 
For more info on changes made:
- `setup-node`, `einaregilsson/beanstalk-deploy` and `configure-aws-credential`: [here](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12)
- [`set-output`](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

Did not update:
- Lambda functions because they will be refactored away
- `db-migration.yml` workflow (not sure about the status, doesn't seem to be in use) @stanleynguyen 
- Warnings from ECS task definitions
  - If I understand the setup properly, our GitHub Actions are pulling the task definitions from ECS and then using to deploy the new image
  - However, in doing so, it is actually pulling fields that are strictly not necessary and are ignored by the CI, such as `compatibilities`, `taskDefinitionArn`, `requiresAttributes`, `revision`, `status`, `registeredAt`
  - In theory, we could modify the bash command to remove these fields, but it seems more effort than it's worth, so I'm gonna ignore this for now

- [x] Test on staging

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,9 +13,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
-          node-version: "16"
+          node-version-file: .nvmrc
       - name: Lint lock file
         run: npx lockfile-lint --type npm --path package-lock.json -o "https:" -o "file:" --allowed-hosts npm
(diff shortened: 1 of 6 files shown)


tan-zx:
ohh there are way more to update

image

stanleynguyen (approved):

PR #2016, opened 2023-04-30
refactor(email): refactor responses and remove unnecessary email campaign endpoints

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Update-responses-for-email-related-endpoints-errors-b302ecb6e25f4844969ef76089be287f?pvs=4)

## Solution

- Adjust responses
- Remove unnecessary endpoints

## Deployment Checklist

- [ ] N/A

### DIFF
--- a/backend/src/core/services/job.service.ts
+++ b/backend/src/core/services/job.service.ts
@@ -120,17 +120,6 @@ const sendCampaign = async ({
 }
 
 /**
- * @see ../backend/src/database/migrations/20210501121338-create-stop-jobs-function.js
- * @param campaignId
- */
-const stopCampaign = (campaignId: number): Promise<any> | undefined => {
-  return Campaign.sequelize?.query('SELECT stop_jobs(:campaignId);', {
-    replacements: { campaignId },
-    type: QueryTypes.SELECT,
-  })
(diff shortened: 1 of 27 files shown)


backend/src/core/services/job.service.ts
```diff
@@ -119,17 +119,6 @@ const sendCampaign = async ({
   return Promise.all(jobs)
 }
 
-/**
- * @see ../backend/src/database/migrations/20210501121338-create-stop-jobs-function.js
- * @param campaignId
- */
-const stopCampaign = (campaignId: number): Promise<any> | undefined => {
-  return Campaign.sequelize?.query('SELECT stop_jobs(:campaignId);', {
```
richard-yak, line 127: should we also create a migration to kill this function in our DB?

richard-yak (approved):

backend/src/core/services/job.service.ts
stanleynguyen: Let's keep it first for the record, we might want to use this one instead of deleting in the delete campaign endpoint. This can be refactored out if we want to later in the db refactor

PR #2017, opened 2023-05-01
refactor(backend): refactor sms api error responses

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Update-responses-for-sms-related-endpoints-errors-7599ca20255e44339fe64b64f2af1454?pvs=4)

## Solution

- update error responses

## Deployment Checklist

- [ ] N/A

### DIFF
--- a/backend/src/sms/middlewares/sms-template.middleware.ts
+++ b/backend/src/sms/middlewares/sms-template.middleware.ts
@@ -12,6 +12,7 @@ import { SmsTemplateService, SmsService } from '@sms/services'
 import { StoreTemplateOutput } from '@sms/interfaces'
 import { loggerWithLabel } from '@core/logger'
 import { ChannelType } from '@core/constants'
+import { ApiInvalidTemplateError } from '@core/errors/rest-api.errors'
 
 const logger = loggerWithLabel(module)
 /**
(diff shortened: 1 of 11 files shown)


backend/src/sms/middlewares/sms-template.middleware.ts
```diff
@@ -252,16 +253,11 @@ const pollCsvStatusHandler = async (
...
-    await UploadService.deleteS3TempKeys(+campaignId)
-    res.sendStatus(200)
-  } catch (e) {
-    next(e)
-  }
+  res: Response
+): Promise<Response> => {
+  const { campaignId } = req.params
+  await UploadService.deleteS3TempKeys(+campaignId)
+  return res.status(200).json({ id: campaignId })
```
richard-yak, line 260: i see that sometimes there's a return and sometimes there isn't. i understand it's the same but just making sure

richard-yak (approved):

backend/src/sms/middlewares/sms-template.middleware.ts
stanleynguyen: Because we're not catching the error here anymore (not needed) so I'm returning to keep the function signature simply to just `Response`. My convention is usually `Response` for the final handler function and `void` for middlewares in the middle

PR #2018, opened 2023-05-02
refactor(backend): update responses for telegram endpoints

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Update-responses-for-telegram-related-endpoints-errors-c9e353019a384daaae216751912e2cea?pvs=4)

## Solution

Update the responses

## Deployment Checklist

- [ ] N/A

### DIFF
--- a/backend/src/telegram/middlewares/telegram.middleware.ts
+++ b/backend/src/telegram/middlewares/telegram.middleware.ts
@@ -4,7 +4,13 @@ import { CredentialService } from '@core/services'
 import { TelegramService } from '@telegram/services'
 import { loggerWithLabel } from '@core/logger'
 import { formatDefaultCredentialName } from '@core/utils'
-import { ApiAuthorizationError } from '@core/errors/rest-api.errors'
+import {
+  ApiAuthorizationError,
+  ApiInvalidCredentialLabelError,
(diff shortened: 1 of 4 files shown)


backend/src/telegram/middlewares/telegram.middleware.ts
```diff
@@ -352,21 +345,16 @@ export const InitTelegramMiddleware = (
...
+    res: Response
+  ): Promise<Response> => {
+    const { campaignId } = req.params
+    const { credentialName } = res.locals
+    if (!credentialName) {
+      throw new Error('Credential does not exist')
     }
+
+    await TelegramService.setCampaignCredential(+campaignId, credentialName)
+    return res.json({ message: 'OK' })
```
richard-yak, line 357: [to clarify] is this supposed to be the campaignId? I see that some other endpoints return the campaignId

richard-yak (approved):

backend/src/telegram/middlewares/telegram.middleware.ts
stanleynguyen: No, the rest of the equivalent endpoints for other types were kept this way as well

PR #2020, opened 2023-05-05
change 'government' to 'Government' in email template

twoflyingforks:
## Problem

The standard "A Singapore Government Agency Website" banner at the top of usual Govt websites has Government with a capital G. 

## Solution

Changed "government" to "Government" for consistency. Looks more legitimate too. 

## Deployment Checklist

Nil


tan-zx:
image

do any of these look like copy changes we should make as well?

twoflyingforks:
Thanks! Looked through, think the other instance you flagged, is the only other one that we should consider changing too. PR updated - have pushed the changes to my fork.

The others - the phrases/sentences are entirely in lower case, so don't think changes are needed.

tan-zx (approved):

PR #2022, opened 2023-05-08
feat: updating create campaign modal

richard-yak:
[Ticket](https://www.notion.so/opengov/Update-new-create-campaign-modal-as-per-design-90d4b5e7b8504a4c8193186739d811f4?pvs=4)

Included in this release
1. New radio selection instead of buttons for create campaign
2. Splitting of e-mail and protected e-mail as per design
3. Updates to tests


e2e/tests/email-campaign.test.ts
```diff
@@ -1,10 +1,10 @@
-import test, { expect } from '@playwright/test';
-import { Page } from 'playwright';
+import test, {expect} from '@playwright/test';
```
stanleynguyen, line 1: Can you run `npm run lint` at project root to get rid of these formatting diffs?

frontend/src/components/dashboard/create/create-modal/CreateModal.tsx
```diff
@@ -99,114 +76,104 @@ const CreateModal = ({
...
+                type="radio"
+                aria-label={ChannelType.Email}
+                value={ChannelType.Email}
+                checked={selectedChannel === ChannelType.Email && !protect}
+                onChange={() => {
+                  setSelectedChannel(ChannelType.Email)
+                  setProtected(false)
+                }}
+              />
+              <p className={styles.subtext}>Email</p>
```
stanleynguyen, line 89: Should use `<label />` tag and `htmlFor` prop so that user can click on the label to choose the option as well

stanleynguyen (changes requested):
Just 2 comments

e2e/tests/email-campaign.test.ts
richard-yak: this command didn't work, but i had to run `npx prettier --single-quote --trailing-comma all --write e2e/*` to get the right formatting. perhaps something wrong with my prettier config that i had to do it manually

richard-yak:
holding, i just realised i have to make updates to duplicate modal as well

tan-zx (approved):
lgtm, just a couple of comments:
1. The design includes the `WhatsApp` radio button, but I'm guessing that's intentionally excluded until the feature is ready right
2. Could I suggest the following copy change?

Before:
> Choosing a channel is irreversible. If you would like to change channels halfway, you will have to restart. You may only edit your message or re-upload recipients list. Please proceed with caution.

After:
> Choosing a channel is irreversible for a given campaign. If you would like to change channels, you will have to create a new campaign. 

Reasoning:
- Really not that big a deal to re-create campaign, don't need "proceed with caution"
- Don't think it's necessary to talk about editing message or re-uploading recipients list at this stage

stanleynguyen (approved):

PR #2033, opened 2023-05-16
chore: enhance security as per CSA recommendations

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/Work-on-email-domain-security-693c8755a2d04cd9bb80b1495d5a0d9a?pvs=4)

## Solution

- Added `helmet` for backend & API server
- Enabled DNSSEC from Cloudflare
- Override some HTTP headers in Amplify

## Deployment Checklist

- N/A

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -28,7 +28,7 @@
         "express-session": "1.17.0",
         "express-winston": "4.2.0",
         "file-type": "^16.5.4",
-        "helmet": "3.23.3",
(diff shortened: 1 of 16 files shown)


tan-zx (approved):
lgtm

would it be helpful to break this ticket into multiple smaller tickets (just realized the number of things the reports are asking us to address is massive), so it's easier to see which recommendations these changes address!

PR #2039, opened 2023-05-24
chore: migrate backend to ECS deployed using Pulumi components

tan-zx:
[Ticket](https://www.notion.so/opengov/Set-up-IAC-for-new-backend-lambda-architecture-8b267cda5c6c4dd5bc2069b4a3ce0b1e?pvs=4)

## Overview

In this PR, our high level goal is to deploy our backend on ECS instead of Elastic Beanstalk. 

All new infra created to support the changes in this PR are done using Pulumi components (see infra repo [here](https://github.com/opengovsg/postmangovsg-infra)). This is part of a larger rearchitecting described [here](https://www.notion.so/opengov/RFC-Rearchitecting-Postman-s-Backend-and-Lambda-e986377446234fd3867f7ec65879303f?pvs=4)

## Decision log
- We have changed the port number of the backend to `8080` instead of `4000` (because this is the default port number for ECS and it was easier to make this application-level change than to make an infra-level change)
- Removed a bunch of EB specific config files
- We have written the `ecs-task-definition` in `yml` (because it supports comments, unlike `json`). This is then converted to `json` as part of the GitHub Actions CI/CD.
- Currently, Elastic Beanstalk is split between two environments, one for callback and one for everything else, so that if callbacks are too spiky and brings down the callback server, the API server will not be affected. Ever since we've throttled the callback rate on SNS for emails, the need for this split no longer exists. 
- Create new S3 buckets for CSVs and common attachments (and migrate existing ones over) and manage entirely via IAC. That way settings like auto deleting objects after X amount of time can be managed using code too. (Currently, these are hidden in the console settings.)

## Remaining to-dos:
- [x] Test on staging + ensure end-to-end tests pass @tan-zx
  - [ ] **[optional]** Move CIDR, security group ID, ARN to be "secrets" in Pulumi @tan-zx 
- [x] Move ECS staging to separate subdomain and branch (`staging2`), so that it doesn't conflict with existing staging users
  - [ ] To undo this change before merging
- [ ] Create new S3 buckets using IAC (to manage everything) @tan-zx 
  - [ ] Bucket replication — sync files to new bucket
- [ ] Datadog integration @blakegong 
- [ ] Manage Cloudflare certs (and DNS?) using IAC
- [ ] Modify GitHub Actions to preserve existing behaviour of reverting upon failure of e2e tests @blakegong 
- [ ] Load test to ensure that our ECS minimum instance can handle the agreed upon ICA load + callback (see [this](https://www.notion.so/opengov/Load-Testing-for-SGAC-Results-70e910bef5f14d6d815c9f08bd243578)) @tan-zx
  - (Need to switch `callback-staging` on Cloudflare to point to ECS load balancer, otherwise, callbacks will be hitting the Beanstalk instead)
- [ ] Write full deployment checklist @tan-zx 
- [ ] Set up proper DX on `postmangovsg-infra`

## Postpone to subsequent PRs
- [ ] Migrate existing existing Redis use cases (OTP, session, rate limit) to the new MemoryDB
- [ ] Eventually deploy ECS to three clusters: 
  1. "backend-for-frontend" (new subdomain: `bff` and `bff-staging`?)
  2. tx APIs (`api` and `api-staging`)
  3. callbacks (`callback` and `callback-staging`)
- [ ] Rest of rearchitecting work (using BullMQ etc.)

### Staging
- [ ] Destroy current staging infra using Pulumi on current AWS account
- [ ] Create a new AWS account to host staging infra (@blakegong not sure if we need to change our VPC peering component if this is cross-AWS account?)
- [ ] Create new set of staging infra
- [ ] Modify DNS on Cloudflare to point to load balancer of ECS
- [ ] Run end-to-end tests on staging

### Production
- [ ] Create new set of production infra on existing AWS account
- [ ] Modify DNS on Cloudflare to point to load balancer of ECS
- [ ] Merge in PR → end-to-end tests will run automatically. (Wonder if there's some way for us to run this first before changing the DNS hmm)
- [ ] Monitor for failures, if any, switch DNS back to existing Elastic Beanstalk

### After successful deployment
- [ ] Destroy existing Elastic Beanstalk environments on staging and production 2 weeks after DNS switch + no incidents
- [ ] Remove unused subdomains CNAME record on Cloudflare (`api-staging2`, `api-callback2`)


tan-zx:
NOTE: need to put on hold

This PR spins up a new VPC. Currently, we are using:
- VPC peering to connect to WA Business Client On Premises API. 
- Whitelisted IP to connect to Flamingo database

The WhatsApp functionality will break if we implement if this PR is deployed. Ideally, we should migrate to WhatsApp's Cloud API before continuing this rearchitecting.

PR #2040, opened 2023-05-30
feat(backend): endpoint to upload common attachments

stanleynguyen:
## Problem

[Ticket](https://www.notion.so/opengov/File-upload-endpoint-9d23558d490e4ff68cf66af8888a4401?pvs=4)

## Solution

Implement uploading to a new bucket

## Deployment Checklist

- [ ] Create production bucket
- [ ] Set env vars for common attachments
- [ ] Migrate db


backend/src/core/utils/attachment.ts
```diff
@@ -9,4 +18,7 @@ const ensureAttachmentsFieldIsArray = (
   return attachments
 }
 
-export { ensureAttachmentsFieldIsArray }
+export const getAttachmentHash = (content: Buffer): string => {
```
tan-zx, line 21: Think you mentioned this is unnecessary and can be refactored. Can put in a TODO here?

backend/src/core/utils/attachment.ts
```diff
@@ -1,6 +1,15 @@
+import crypto from 'crypto'
 import { UploadedFile } from 'express-fileupload'
 
-const ensureAttachmentsFieldIsArray = (
+export interface Attachment {
```
tan-zx, line 4: Note: find type exported by library if available

backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -57,7 +52,7 @@ export const InitEmailTransactionalMiddleware = (
     from: string
     recipient: string
     reply_to: string
-    attachments?: { data: Buffer; name: string; size: number }[]
+    attachments?: Attachment[]
```
tan-zx, line 55: Good catch

backend/src/core/config.ts
```diff
@@ -697,6 +701,19 @@ const config: Config<ConfigSchema> = convict({
       format: Number,
     },
   },
+  commonAttachments: {
+    maxFileSize: {
+      doc: 'Maximum file size of a common attachment',
+      default: 50 * 1024 * 1024, // 50 MB
```
tan-zx, line 707: Why so big

backend/src/core/middlewares/file-attachment.middleware.ts
```diff
@@ -1,40 +1,42 @@
...
+const getFileUploadHandler = (fileSize: number, fieldSize: number) =>
+  fileUpload({
+    limits: {
+      fileSize,
+      fieldSize,
+    },
+    abortOnLimit: true,
+    limitHandler: function (_req: Request, _res: Response, next: NextFunction) {
+      next(
+        new ApiAttachmentLimitError(
```
tan-zx, line 34: Should this be renamed since it's no longer specific to API

backend/src/database/migrations/20230526011651-create-common-attachments-table.js
```diff
@@ -0,0 +1,35 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.createTable('common_attachments', {
+      id: {
+        primaryKey: true,
+        allowNull: false,
+        type: Sequelize.DataTypes.UUID,
```
tan-zx, line 9: I'm guessing we're using UUID here so that we have an ID we can easily pass to S3. 

However I wonder if using it for primary key might be bad for indexing and if it might not be better to use ULID instead. 

(Will insert links later -- have thought through sth similar for tx emails previously)

richard-yak (approved):
looks good, looks like `common_attachments` table is purely for record purposes, no querying? 

// edit : oh ok i just saw query by id in another PR

backend/src/core/utils/attachment.ts
tan-zx: actually, I just went in and check, `UploadedFile` is the correct file type, can delete this `Attachment` interface

backend/src/database/migrations/20230526011651-create-common-attachments-table.js
stanleynguyen: For our usecase, none of the advantages of ULID is applicable (namely sorting using ID => we have created_at, no foreseeable need for partitioning), while the downside of it being not very widely supported yet (specifically not by sequelize) is applicable
=> Sticking with UUID v4

backend/src/core/utils/attachment.ts
stanleynguyen: Lemme just include it in this PR

stanleynguyen:
PENDING: Endpoint to stream embeds are still sitting behind authentication, they need to be public

PR #2043, opened 2023-05-30
feat(frontend): direct upload for image embedding

stanleynguyen:
## Problem

- [Ticket](https://www.notion.so/opengov/Rich-text-Editor-Integration-9dd07e59f741482a89d5472496637fd9?pvs=4)
- On top of that, I also added an enhancement so user can set the width to whatever they want instead of fixed values that we previously allow them to (25, 50, 75, and 100)

## Solution

- Implement dropzone inside the image embed popup

- Demo

https://github.com/opengovsg/postmangovsg/assets/12974927/1417327a-5f33-423b-9b34-65f5db269c69

The final email looks like this (the image wasn't working cos it's being served from `localhost:4000`)
Screenshot 2023-05-30 at 7 54 51 PM

## Deployment Checklist

- [ ] N/A


richard-yak (approved):

frontend/src/index.tsx
```diff
@@ -29,12 +28,10 @@ const container = document.getElementById('root') as HTMLElement
 const root = createRoot(container)
 root.render(
   <BrowserRouter>
-    <StrictMode>
```
tan-zx, line 32: I get that `<StrictMode>` was causing the embed image feature to not work locally, but I feel like removing it entirely from the entire application is not very good.

Have [read this](https://react.dev/reference/react/StrictMode) and seems like `StrictMode` could help catch problems in the future. Unfortunately, there is no way to opt out of Strict Mode inside a tree wrapped in `StrictMode`.

Perhaps a compromise solution might be to:
- Remove `StrictMode` for the purposes of this PR
- Add to our frontend tech debt backlog with a view towards re-adding `StrictMode` when we want to properly refactor our frontend

tan-zx (approved):
lgtm

tan-zx (approved):
lgtm

PR #2045, opened 2023-05-31
feat: Introducing phonebook service to modularize worker & backend usage

richard-yak:
in this release: 
- Introduce a new client service to connect with phonebook and make it reusable between worker & backend. 
- changes that are not invoked at the moment, simply to break down further PRs
- kill switch on worker side to make sure we can halt any phonebook related functionality even after job has been queued.

To note:
- There are some older configs used by phonebook v1. They will be removed once we have deprecated the older functionalities that reference these configs
- The older configs are duplicates of newer ones, but now we keep the naming consistent and renamed them to be more intuitive. 
- There is no kill switch in backend as we can control the display on the client side.

Before deployment
- [ ] Add these config values into aws secrets manager
- [ ] Update 1Password shared env file

### DIFF
--- a/backend/src/core/config.ts
+++ b/backend/src/core/config.ts
@@ -154,6 +154,12 @@ interface ConfigSchema {
     maxAttachmentNum: number
     maxCumulativeAttachmentsSize: number
   }
+
+  phonebook: {
+    endpointUrl: string
+    apiKey: string
+    version: string
+  }
 }
 
 convict.addFormats({
@@ -695,6 +701,26 @@ const config: Config<ConfigSchema> = convict({
       default: 10 * 1024 * 1024,
       env: 'FILE_ATTACHMENT_MAX_CUMULATIVE_SIZE',
       format: Number,
+    },
+  },
+  phonebook: {
+    endpointUrl: {
+      doc: 'Endpoint url of phonebook server',
+      default: 'http://localhost:8080',
+      env: 'PHONEBOOK_URL',
+      format: 'required-string',
+    },
+    apiKey: {
+      doc: 'API key to make requests to Phonebook',
+      default: 'API_KEY',
+      env: 'PHONEBOOK_API_KEY',
+      format: 'required-string',
+    },
+    version: {
+      doc: 'Version number of Phonebook',
+      default: '1',
+      env: 'PHONEBOOK_VERSION',
+      format: 'required-string',
     },
   },
 })
--- a/backend/src/core/config.ts
+++ b/backend/src/core/config.ts
@@ -158,7 +158,6 @@ interface ConfigSchema {
   phonebook: {
     endpointUrl: string
     apiKey: string
-    version: string
   }
 }
 
@@ -714,12 +713,6 @@ const config: Config<ConfigSchema> = convict({
       doc: 'API key to make requests to Phonebook',
       default: 'API_KEY',
       env: 'PHONEBOOK_API_KEY',
-      format: 'required-string',
-    },
-    version: {
-      doc: 'Version number of Phonebook',
-      default: '1',
-      env: 'PHONEBOOK_VERSION',
       format: 'required-string',
     },
   },
--- a/shared/src/clients/phonebook-client.class/index.ts
+++ b/shared/src/clients/phonebook-client.class/index.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosInstance, AxiosRequestConfig } from 'axios'
+import axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios'
 import * as http from 'http'
 
 export default class PhonebookClient {
@@ -20,7 +20,10 @@ export default class PhonebookClient {
     })
   }
 
-  private request(options: AxiosRequestConfig, body?: any): Promise<any> {
+  private request<TBody>(
+    options: AxiosRequestConfig<TBody>,
+    body?: TBody
+  ): Promise<AxiosResponse> {
     const defaultOptions: AxiosRequestConfig = {
       method: 'post', // default method will be post
     }
--- a/shared/src/clients/phonebook-client.class/index.ts
+++ b/shared/src/clients/phonebook-client.class/index.ts
@@ -34,7 +34,10 @@ export default class PhonebookClient {
     })
   }
 
-  public async getManagedLists(email: string, channel: string) {
+  public async getManagedLists(
+    email: string,
+    channel: string
+  ): Promise<{ id: number; name: string }[]> {
     try {
       const res = await this.request({
         method: 'get',
@@ -50,7 +53,9 @@ export default class PhonebookClient {
     }
   }
 
-  public async getManagedListById(listId: number) {
+  public async getManagedListById(
+    listId: number
+  ): Promise<{ s3Key: string; etag: string; filename: string }> {
     try {
       const res = await this.request({
         method: 'get',
--- a/shared/src/clients/phonebook-client.class/index.ts
+++ b/shared/src/clients/phonebook-client.class/index.ts
@@ -1,0 +1,62 @@
+import axios, { AxiosInstance, AxiosRequestConfig } from 'axios'
+
+export default class PhonebookClient {
+  private client: AxiosInstance
+  private baseUrl: string
+  private apiKey: string
+  private version: string
+
+  constructor(baseUrl: string, apiKey: string, version: string) {
+    this.baseUrl = baseUrl
+    this.apiKey = apiKey
+    this.version = version
+    this.client = axios.create({
+      baseURL: this.baseUrl,
+      timeout: 5000,
+      headers: {
+        'Content-Type': 'application/json',
+        'x-api-key': this.apiKey,
+      },
+    })
+  }
+
+  private request(options: AxiosRequestConfig, body?: any): Promise<any> {
+    options.url = `/api/v${this.version}/${options.url}`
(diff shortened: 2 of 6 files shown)


backend/src/core/config.ts
```diff
@@ -154,6 +154,12 @@ interface ConfigSchema {
     maxAttachmentNum: number
     maxCumulativeAttachmentsSize: number
   }
+
+  phonebook: {
```
hanu-ogp, line 158: kill switch not required here ?

hanu-ogp (approved):

backend/src/core/config.ts
richard-yak: Intention is that we can control backend's killswitch by hiding UI of frontend instead. (Returning empty list from agency list endpoint will disable the ui as well)

worker/src/core/config.ts
```diff
@@ -334,6 +341,31 @@ const config: Config<ConfigSchema> = convict({
...
+    },
+    apiKey: {
+      doc: 'API key to make requests to Phonebook',
+      default: 'API_KEY',
+      env: 'PHONEBOOK_API_KEY',
+      format: 'required-string',
+    },
+    version: {
+      doc: 'Version number of Phonebook',
+      default: '1',
```
stanleynguyen, line 364: [suggestion] I was thinking we can collapse this env var into the `endpointUrl` (i.e. so instead of `https://phonebook.com/` do `https://phonebook.com/v1`)

shared/src/clients/phonebook-client.class/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    } catch (err) {
+      throw new Error(err as any)
+    }
+  }
+
+  public async getManagedListById(listId: number) {
+    try {
+      const res = await this.request({
+        method: 'get',
+        url: `managed-list/${listId}/members/s3`,
```
stanleynguyen, line 55: Should add a leading slash

shared/src/clients/phonebook-client.class/index.ts
```diff
@@ -0,0 +1,62 @@
...
+  private client: AxiosInstance
+  private baseUrl: string
+  private apiKey: string
+  private version: string
+
+  constructor(baseUrl: string, apiKey: string, version: string) {
+    this.baseUrl = baseUrl
+    this.apiKey = apiKey
+    this.version = version
+    this.client = axios.create({
```
stanleynguyen, line 13: [suggestion] Can also add `keepAlive` functionalities to reuse the same connection instead of establishing new one to phonebook everytime https://stackoverflow.com/questions/69889711/does-axios-support-keepalive-functionality-same-as-the-native-fetch

shared/src/clients/phonebook-client.class/index.ts
```diff
@@ -0,0 +1,61 @@
...
+      timeout: 5000,
+      headers: {
+        'Content-Type': 'application/json',
+        'x-api-key': this.apiKey,
+      },
+      httpAgent: new http.Agent({ keepAlive: true }),
+    })
+  }
+
+  private request(options: AxiosRequestConfig, body?: any): Promise<any> {
```
tan-zx, line 23: **[change request]** can we type this instead of using `any`? code smell to use `any` in TS.

shared/src/clients/phonebook-client.class/index.ts
```diff
@@ -0,0 +1,61 @@
...
+        method: 'get',
+        url: `/managed-list`,
+        params: {
+          owner: email,
+          channel,
+        },
+      })
+      return res.data
+    } catch (err) {
+      throw new Error(err as any)
```
tan-zx, line 46: same

shared/src/clients/phonebook-client.class/index.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  public async getManagedListById(listId: number) {
+    try {
+      const res = await this.request({
+        method: 'get',
+        url: `/managed-list/${listId}/members/s3`,
+      })
+      return res.data
+    } catch (err) {
+      throw new Error(err as any)
```
tan-zx, line 58: same

shared/src/clients/phonebook-client.class/index.ts
richard-yak: this is intended as the body data for a request client could be anything. the type is enforced in the function that calls `request`, i.e. the functions below.

tan-zx: Hmm so it's not literally anything right, I'm gonna guess at the very least, it'll be a JSON with some compulsory fields (`method`, `url`, etc.)?

richard-yak: those mentioned compulsory fields are part of AxiosRequestConfig, the first param, not the body param.

It could be a JSON, it could also be a DTO that we define. Axios handles jsonification for us. 
The body is basically a passthrough param for the `Data`, of type `any`, that is in the axios client, refer to screenshot below.
Screenshot 2023-06-07 at 10 26 33 AM

worker/src/core/config.ts
richard-yak: refactored to use one config variable instead of 2

shared/src/clients/phonebook-client.class/index.ts
richard-yak: refactored to return more informative custom error message

shared/src/clients/phonebook-client.class/index.ts
richard-yak: refactored to return more informative custom error message

shared/src/clients/phonebook-client.class/index.ts
tan-zx: Oops sorry, have misread the code. 

Hmm iinw the `<D = any>` is to provide a default `any` to the generic, not to say that it should be of type `any` though. If we use `any`, we will lose all type-safety 

Seems like the solution here is to use generics. Let me know if any of the below is wrong:
- the second `body` optional argument's type is not known at compile time. this could be the DTO we define.
- the generic that `AxiosRequestConfig` accepts is to provide typing to the data returned

In that case, this should be rewritten as:

```
  private request<TBody, TData>(
    options: AxiosRequestConfig<TData>,
    body?: TBody
  ): Promise<TData> {
    const defaultOptions: AxiosRequestConfig = {
      method: 'post', // default method will be post
    }
    return this.client.request({
      ...defaultOptions,
      ...options,
      data: { ...body },
    })
  }
```

richard-yak: on point 2: 
> the generic that AxiosRequestConfig accepts is to provide typing to the data returned
the generic that `AxiosRequestConfig` accepts is to provide typing to the body, not the data.

That said, I am not 100% familiar with generics, but i do understand what you are trying to say. 

I have updated the code to not use `any`. let me know if there are further changes required.

tan-zx: Let me look at it again when I reach office.

tan-zx: > the generic that AxiosRequestConfig accepts is to provide typing to the body in the request, not the data returned.

Ah you're right. Your change to `options: AxiosRequestConfig<TBody>, body?: TBody` is correct. 

I think there is still a need for the second generic though, using `AxiosResponse` simply hides the fact that `any` is still being returned:
image

stanleynguyen (approved):

shared/src/clients/phonebook-client.class/index.ts
richard-yak: updated return type

tan-zx: I see you merged alr. It's actually better to use inferred return type (here, to use the second `TData` generic), rather than explicitly add a return type on a per function basis. Right now, your `res.data` is still `any` and you're merely asserting the type in the return (which could be false; [see this](https://www.youtube.com/watch?v=I6V2FkW1ozQ)). 

image

With the second generic and inferred return type, the user of `request` will be required to state the generic, the type will naturally flow through to `res.data`, and this has helped to catch the edge case where you might be sending `data: {}` if `body` is undefined.

image

```TS
request<TData, TBody = any>(
    options: AxiosRequestConfig<TBody>,
    body?: TBody
  ) {
    const defaultOptions: AxiosRequestConfig = {
      method: 'post', // default method will be post
    }
    const requestConfig = body
      ? { ...defaultOptions, ...options, data: { ...body } }
      : { ...defaultOptions, ...options }
    return this.client.request<TData, AxiosResponse<TData>, TBody>(
      requestConfig
    )
  }

  public async getManagedLists(email: string, channel: string) {
    try {
      const res = await this.request<{ id: number; name: string }[]>({
        method: 'get',
        url: `/managed-list`,
        params: {
          owner: email,
          channel,
        },
      })
      return res.data
    } catch (err) {
      throw new Error('Could not get managed lists')
    }
  }

  public async getManagedListById(listId: number) {
    try {
      const res = await this.request<{
        s3Key: string
        etag: string
        filename: string
      }>({
        method: 'get',
        url: `/managed-list/${listId}/members/s3`,
      })
      return res.data
    } catch (err) {
      throw new Error('Could not get managed list by id')
    }
  }
}

```

richard-yak: thanks for the in-depth explanation! appreciate that you take the time to write this down and i will be incorporating this into the separate PRs that are coming up as per the above comments.

PR #2046, opened 2023-06-01
feat: phonebook worker integration

richard-yak:
Included in this release:
- Using a similar implementation of the old contact pref link, retrieve the unique links and append them to the email / SMS.
- Using two different implementation from SMS and Emails. 

Post-deployment:

After this deployment is stable and with PM approval, remove all old phonebook contact preference functionality and related logic as cleanup.

Other considerations:

This is just one part of the deployment! There is a second part for backend and ui changes pending.

Deployment Checklist 

- [ ] Check env vars first!!!


shared/src/clients/phonebook-client.class/index.ts
```diff
@@ -59,4 +60,21 @@ export default class PhonebookClient {
...
   }
+
+  public async getUniqueLinksForUsers(
+    body: PhonebookChannelDto
+  ): Promise<UserChannel[]> {
+    try {
+      const res = await this.request(
+        {
+          method: 'post',
+          url: 'public-user/unique_links',
```
hanu-ogp, line 71: this reveals an inconsistency in phonebook api - follows both `-` and `_`.  Created as task https://linear.app/ogp/issue/PBOOK-35/inconsistencies-in-api-following-both-and-

shared/src/clients/phonebook-client.class/interfaces.ts
```diff
@@ -0,0 +1,10 @@
+export type PhonebookChannelDto = {
+  userChannels: UserChannel[]
+  postmanCampaignOwner: string
```
hanu-ogp, line 3: i dont think we need `postmanCampaignOwner` anymore as it was needed for V1

shared/src/clients/phonebook-client.class/index.ts
```diff
@@ -59,4 +60,21 @@ export default class PhonebookClient {
       throw new Error(err as any)
     }
   }
+
+  public async getUniqueLinksForUsers(
+    body: PhonebookChannelDto
+  ): Promise<UserChannel[]> {
+    try {
+      const res = await this.request(
```
hanu-ogp, line 68: Lets continue the practice of circuit breaker. We dont want campaigns to halt or postman to slowdown if phonebook isnt responding

shared/src/clients/phonebook-client.class/index.ts
richard-yak: pre-emptively changing to use unique-links

worker/src/core/services/phonebook.service.ts
```diff
@@ -0,0 +1,106 @@
...
+        const userChannel = userChannelMap.get(row.message.recipient)
+        return {
+          ...row.message,
+          showMasthead,
+          userUniqueLink: userChannel?.userUniqueLink || '',
+        }
+      })
+    })
+    .catch((err) => {
+      throw err
```
hanu-ogp, line 71: Note - in case of error from phonebook, the campaign should continue to be sent without the link. I believe its taken care of, just in case

hanu-ogp (approved):

worker/src/core/services/phonebook.service.ts
```diff
@@ -0,0 +1,128 @@
...
+      channelId: row.message.recipient,
+    }
+  })
+  const payload: PhonebookChannelDto = {
+    userChannels: channels,
+  }
+
+  return breaker
+    .fire(payload)
+    .then((res) => {
```
stanleynguyen, line 56: Can we write this in a normal async/await flow instead of `.then.catch` as this function has already been declared `async`

worker/src/core/services/phonebook.service.ts
```diff
@@ -0,0 +1,128 @@
...
+      channelId: message.recipient,
+    }
+  })
+  const payload: PhonebookChannelDto = {
+    userChannels: channels,
+  }
+
+  return breaker
+    .fire(payload)
+    .then((res) => {
```
stanleynguyen, line 104: Can we write this in a normal async/await flow instead of `.then.catch` as this function has already been declared `async`

worker/src/core/services/phonebook.service.ts
hanu-ogp: async await doesnt work. The breaker code works only with then

stanleynguyen: @richard-yak @hanu-ogp have we tried that out and confirmed that it doesnt work? Cos for this specific piece of code I don't see how it wouldn't. 
Can we tried it out and compare the behaviours? I'm really curious to why it would not work

hanu-ogp: yes i tried last time when i implemented circuit breaker and async await didnt work. We should merge this PR as this is a non blocking and cosmetic issue and @richard-yak Can you follow up post launch to see if code can be converted to async await

stanleynguyen (approved):
TODO: Experiment and tryna convert circuit breaker to use async/await

richard-yak:
This portion no longer has a change, will merge into the release branch

PR #2047, opened 2023-06-07
docs(openapi): add govsg endpoints

stanleynguyen:


stanleynguyen:
@tan-zx the diff a bit fked so a better approach to review is to checkout the cost and hosting the swagger in local

backend/openapi.yaml
```diff
@@ -1032,20 +1031,875 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/Error404Response'
+        '500':
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error500Response'
+
+  /campaign/{campaignId}/govsg:
```
jia1, line 1512: Qn: Why do we not attempt to reuse existing endpoints/schemas? At least for this endpoint, the response schemas only differ in one of the keys i.e. `email_templates` vs `govsg_templates`.

backend/openapi.yaml
```diff
@@ -1032,20 +1031,875 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error403Response'
+        '500':
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error500Response'
+  /campaign/{campaignId}/govsg/upload/start:
```
jia1, line 1586: Qn: Sorry if it sounds a bit RTFM but, how are `campaignId`s generated?

When I was trying the product out, the IDs look like incrementing numbers to me.

And as such, I don't know why we need different paths for different `campaignId` seeing that we could grab all the metadata we need with `campaignId` alone. Is it a legacy thing?

backend/openapi.yaml
stanleynguyen: In our code, we actually have a URL rewrite that fetch all the `campaign` types and then redirect to the corresponding `/<channel_type>` endpoint (for the other endpoints' purposes) hence this. We can keep in mind to refactor this but perhaps later

backend/openapi.yaml
stanleynguyen: There're a bunch of `Campaigns` endpoints that are for creating, updating campaigns etc. The IDs are generated from there.

tan-zx: > how are campaignIds generated

yeah it's auto-incremented primary key in db

> I don't know why we need different paths for different campaignId seeing that we could grab all the metadata we need with campaignId alone

not sure I fully understand this question, but there are a bunch of paths for campaigns that are triggered at different points of creating the campaign (i.e. is not just about grabbing metadata).

tan-zx:
@stanleynguyen can lmk how urgent it is to review this? I'm still working on other stuff + you're still committing changes right.

jia1:
Do GovSG endpoints need a complete upload endpoint like:

`/campaign/{campaignId}/telegram/upload/complete`?

jia1:
Is the store template endpoint like `/campaign/{campaignId}/telegram/template` **not** needed because WhatsApp template approval is manual?

jia1:
For single-recipient campaigns, having a separate endpoint that processes the recipient payload would be helpful,

but right now I have not checked if some kind of recipient data structure exists already. I remember seeing something like a `Lists` class?

backend/openapi.yaml
stanleynguyen: @jia1 that's right

backend/openapi.yaml
stanleynguyen: @jia1 , we can do that, no reason not too. Here I'm just sticking with the current convention for consistency. We can refactor this later, but the impact is rather minimal rn that it could be overlooked (namely just a few extra lines in the route files)

backend/openapi.yaml
```diff
@@ -1032,20 +1031,875 @@ paths:
...
+        '500':
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error500Response'
+    delete:
+      security:
+        - bearerAuth: []
+      description: 'Deletes error status from previous failed upload'
```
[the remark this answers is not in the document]
stanleynguyen: Correcto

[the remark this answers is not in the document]
stanleynguyen: U will see if when reading the frontend code

stanleynguyen:
> For single-recipient campaigns, having a separate endpoint that processes the recipient payload would be helpful,
> 
> but right now I have not checked if some kind of recipient data structure exists already. I remember seeing something like a `Lists` class?

My current idea for sending to a single-recipient is to just use the transactional endpoint. Pending the design to confirm that

stanleynguyen:
> Is the store template endpoint like `/campaign/{campaignId}/telegram/template` **not** needed because WhatsApp template approval is manual?

Correct correct, so we will manually insert templates into our db instead of having them coming from users

jia1 (approved):

jia1:
No other input from me. I can't find the button to resolve the threads though.

tan-zx (approved):
looks good to me

PR #2048, opened 2023-06-07
Richardyak/whatsapp/campaign

stanleynguyen:


backend/src/whatsapp/middlewares/whatsapp.middleware.ts
```diff
@@ -77,7 +77,8 @@ export const InitWhatsappMiddleware = (
     res: Response
   ): Promise<void | Response> => {
     const { campaignId } = req.params
-    return res.json({ campaignId })
+    const result = await WhatsappService.getCampaignDetails(+campaignId)
```
jia1, line 80: TIL [unary plus](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus)

PR #2051, opened 2023-06-08
feat(backend): add data models and migrations for govsg messages

stanleynguyen:
## Problem

Closes [insert issue #]

## Solution

## Deployment Checklist

- [ ] Run db migration


backend/src/database/migrations/20230608034909-create-govsg-tables.js
```diff
@@ -0,0 +1,287 @@
+'use strict'
+
+const { GovsgMessageStatus } = require('@core/constants')
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(
+      "ALTER TYPE enum_campaigns_type ADD VALUE 'WHATSAPP';"
```
jia1, line 8: Is it `GOVSG` or `WHATSAPP`?

stanleynguyen: nice catch, should be GOVSG

tan-zx (approved):
some nits, otherwise looks good to me.

curious whether we want to continue with the `GovsgMessage` naming now that we're only sending WhatsApp? ~~I wonder whether calling it WhatsApp is better since, if we ever have a SMS fallback, we will likely reuse the SMS code too.~~

UPDATE: discussed verbally
- We are using `GovsgMessage` so as to leave open the possibility of SMS callback. In that case, we will extend the data 
- Sending logic will still be channel-specific (i.e. WhatsApp, SMS etc.)

~~I can't think of any reason why a "hybrid" `GovsgMessage` data model is a bad idea, as long as we do the extension carefully~~ UPDATE: see [this](https://github.com/opengovsg/postmangovsg/pull/2051#issuecomment-1601138236)

backend/src/govsg/models/govsg-message-transactional.ts
```diff
@@ -0,0 +1,60 @@
...
+  template: GovsgTemplate
+
+  @Column(DataType.STRING)
+  recipient: string
+
+  @Column(DataType.JSONB)
+  params!: object
+
+  @Column(DataType.STRING)
+  serviceProviderMessageId?: string
```
tan-zx, line 32: what is this supposed to map to? is it:
- SMS vs WhatsApp; or
- the specific WhatsApp client (e.g. phone number 1 vs phone number 2)

backend/src/database/migrations/20230608034909-create-govsg-tables.js
```diff
@@ -0,0 +1,291 @@
...
+        type: Sequelize.DataTypes.STRING(255),
+        allowNull: true,
+      },
+      error_description: {
+        type: Sequelize.DataTypes.TEXT,
+        allowNull: true,
+      },
+      status: {
+        type: Sequelize.DataTypes.ENUM(Object.values(GovsgMessageStatus)),
+        allowNull: true,
```
tan-zx, line 244: `allowNull` can be false right? since the enums should be exhaustive alr

backend/src/govsg/models/govsg-template.ts
```diff
@@ -0,0 +1,47 @@
...
+    type: DataType.BIGINT,
+    autoIncrement: true,
+    primaryKey: true,
+  })
+  id: number
+
+  @BelongsToMany(() => Campaign, {
+    through: { model: () => CampaignGovsgTemplate },
+  })
+  campaign!: Campaign
```
tan-zx, line 23: same as above

backend/src/govsg/models/govsg-template.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  @Column({
+    type: DataType.TEXT,
+  })
+  body?: string
```
tan-zx, line 28: I think the convention is to use `body: string | null` rather `body?: string`, which might screw up TS. (can this field be null though?)

backend/src/govsg/models/govsg-template.ts
```diff
@@ -0,0 +1,47 @@
...
+    allowNull: true,
+  })
+  params?: Array<string>
+
+  @Column({
+    type: DataType.STRING,
+    unique: true,
+    allowNull: true,
+  })
+  whatsappTemplateLabel?: string | null
```
tan-zx, line 41: same as previous

backend/src/govsg/models/govsg-message-transactional.ts
```diff
@@ -0,0 +1,60 @@
...
+  underscored: true,
+  timestamps: true,
+})
+export class GovsgMessageTransactional extends Model<GovsgMessageTransactional> {
+  @ForeignKey(() => GovsgTemplate)
+  @Column(DataType.BIGINT)
+  templateId: number
+
+  @BelongsTo(() => GovsgTemplate)
+  template: GovsgTemplate
```
tan-zx, line 23: now that I am looking at the WhatsApp API more closely, there are two kinds of messages:
- template messages (which makes sense to have this relationship with GovsgTemplate
- text messages (which are custom text messages that you can send)

We are only supporting the first type for our MVP, but I wonder if:
- we should rename this table to indicate that it only covers the first type of message (e.g. `GovsgTemplateMessageTransactional`)
- or we should redesign or data model here to support both kinds (or would that be premature optimisation)

backend/src/govsg/models/govsg-template.ts
tan-zx: think it might be better to call it `whatsappTemplateName`.

This is how the API call looks like:
image

Components is where we populate the parameters e.g. like this:
```
 "components": [{
   "type": "body",
   "parameters": [{
                "type": "text",
                "text": "name"
            },
            {
            "type": "text",
            "text": "Hi there"
            }]
      }] 
```

Think for consistency, we'll put these in the `params` column? Hmm

backend/src/database/migrations/20230608034909-create-govsg-tables.js
```diff
@@ -0,0 +1,300 @@
...
+      },
+      body: {
+        type: Sequelize.DataTypes.TEXT,
+        allowNull: true,
+      },
+      params: {
+        type: Sequelize.DataTypes.JSONB,
+        allowNull: true,
+      },
+      whatsapp_template_label: {
```
tan-zx, line 98: maybe should be unique. see [discussion](https://opengovproducts.slack.com/archives/C05AQ9K7R2T/p1687503185358169?thread_ts=1687485697.075039&cid=C05AQ9K7R2T). 

Also see [this](https://github.com/opengovsg/postmangovsg/pull/2051/files#r1238600480)

backend/src/govsg/models/govsg-template.ts
tan-zx: maybe should [be unique](https://opengovproducts.slack.com/archives/C05AQ9K7R2T/p1687503185358169?thread_ts=1687485697.075039&cid=C05AQ9K7R2T)

tan-zx:
actually no need to change for the parts involving `GovsgMessageTransactional`, I'm changing those directly in my PR

backend/src/govsg/models/govsg-message-transactional.ts
stanleynguyen: Our `GovsgMessage` is already the first type and similar concept across the code base also refers to the same thing, hence adding `Template` would be redundant

backend/src/govsg/models/govsg-message.ts
```diff
@@ -0,0 +1,66 @@
...
+  campaignId: number
+
+  @BelongsTo(() => Campaign)
+  campaign: Campaign
+
+  @Column(DataType.STRING)
+  recipient: string
+
+  @Column(DataType.JSONB)
+  params!: object
```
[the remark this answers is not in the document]
stanleynguyen: Eh shit this is redundant, I was using it to test out something else only haha. Will remove

backend/src/govsg/models/govsg-message.ts
```diff
@@ -0,0 +1,66 @@
...
+import {
+  BelongsTo,
+  Column,
+  DataType,
+  ForeignKey,
+  Model,
+  Table,
+} from 'sequelize-typescript'
+
+export enum ModelGovsgMessageStatus {
```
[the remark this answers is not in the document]
stanleynguyen: Eh shit this is redundant, I was using it to test out something else only haha. Will remove

backend/src/govsg/models/govsg-message.ts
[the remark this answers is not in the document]
stanleynguyen: Hmm yea ok lemme remove

backend/src/govsg/models/govsg-template.ts
stanleynguyen: This field won't be null actl, lemme just remove

backend/src/govsg/models/govsg-template.ts
```diff
@@ -0,0 +1,47 @@
...
+  @Column({
+    type: DataType.TEXT,
+  })
+  body?: string
+
+  @Column({
+    type: DataType.ARRAY(DataType.STRING),
+    allowNull: true,
+  })
+  params?: Array<string>
```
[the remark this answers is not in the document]
stanleynguyen: Practically no it wont be, but logically this field can be null for template without params

backend/src/govsg/models/govsg-template.ts
stanleynguyen: Hmm my first time seeing the namespace, but I think that one could be in the config as we will probably only use one namespace, but can change to name.

No it doesn't necessarily need to be unique so no not gonna make it unique.

backend/src/database/migrations/20230608034909-create-govsg-tables.js
stanleynguyen: Can, but then I have to change the implementation for the db function a bit haha

stanleynguyen: But will do it cos it makes sense

backend/src/govsg/models/govsg-template.ts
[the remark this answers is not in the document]
tan-zx: yeah makes sense. in my code, I was checking for either null or array with length 0

backend/src/govsg/models/govsg-message-transactional.ts
tan-zx: no what I mean is, in the future, if we send transactional `text messages` (which is fully custom messages that don't need to follow any templates), then it might be a bit confusing what to name that table hmm

tan-zx: (this is in contrast to other channels like SMSes and emails, where the content of the transactional messages are alr fully custom)

PR #2052, opened 2023-06-08
test(playwright): add debug mode + improve writeup + fix small bugs

tan-zx:
small changes while trying to run Playwright tests against staging

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -39,4 +39,8 @@ Set the credentials for Gmail API call
 ## To run locally
 
 - From the `e2e` folder, run `npm i`.
-- From the root folder, run `npm run test:e2e`.
+- From the root folder
+  - To run the end-to-end tests directly: `npm run test`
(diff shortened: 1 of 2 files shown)


e2e/package.json
```diff
@@ -4,7 +4,8 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "test": "playwright test"
+    "test": "playwright test",
+    "test:debug": "playwright test --debug"
```
stanleynguyen, line 8: Actually dont need to add this leh. Usually if I want to add flag I will just `npm run test -- --debug`

richard-yak (approved):

e2e/package.json
tan-zx: true 🤔️ I'll remove this and add a link [to this](https://playwright.dev/docs/debug) in the README that more clearly illustrates the debugging options available

PR #2054, opened 2023-06-13
feat(backend): endpoints to formulate and send a govsg campaign

stanleynguyen:
## Problem

Endpoints for creating and sending a Govsg campaign

## Solution

## Deployment Checklist

- [ ] N/A


backend/src/govsg/middlewares/govsg.middleware.ts
```diff
@@ -0,0 +1,55 @@
...
+export const isGovsgCampaignOwnedByUser = async (
+  req: Request,
+  _res: Response,
+  next: NextFunction
+): Promise<void> => {
+  const campaignId = req.params.campaignId
+  const userId = req.session?.user?.id
+  const campaign = await GovsgService.findCampaign(+campaignId, +userId)
+  if (!campaign || !userId) {
+    throw new ApiAuthorizationError("User doesn't have access to this campaign")
```
tan-zx, line 20: [nit] do we want to specify the userId and and campaignId in this error

backend/src/govsg/middlewares/govsg-template.middleware.ts
```diff
@@ -0,0 +1,161 @@
...
+  const statusObj = {
+    is_csv_processing: isCsvProcessing,
+    csv_filename: filename,
+    temp_csv_filename: tempFilename,
+    csv_error: error,
+    num_recipients: undefined as any,
+    preview: undefined as any,
+  }
+  if (!isCsvProcessing) {
+    ;[statusObj.num_recipients, statusObj.preview] = await Promise.all([
```
tan-zx, line 135: weird semicolon, let me review more in-depth later

backend/src/govsg/services/govsg-template.service.ts
```diff
@@ -0,0 +1,64 @@
...
+import { GovsgMessage } from '@govsg/models/govsg-message'
+import { GovsgTemplate } from '@govsg/models/govsg-template'
+import { TemplateClient } from '@shared/templating'
+import { GovsgService } from '.'
+
+const templateCli = new TemplateClient({
+  xssOptions: { whiteList: { br: [] }, stripIgnoreTag: true },
+})
+
+export async function getFilledTemplate(
```
tan-zx, line 13: don't really understand this. what is meant by `FilledTemplate`? also why return `null` instead of throwing an error

(note to self: to review more in-depth later)

tan-zx (commented):
left some comments more for myself, will review more in-depth later

backend/src/govsg/middlewares/govsg-template.middleware.ts
stanleynguyen: this one is cos I use spread operator to reassign to existing vars

backend/src/govsg/services/govsg-template.service.ts
stanleynguyen: Getting the actual template with body and params from pivot table

PR #2056, opened 2023-06-16
feat: remove attachment per file size and max number limit

tan-zx:
takeaway from today's meeting with customer:
- they might need to send >10 attachments
- from our perspective, costs are based on total file size of attachments; as such, as long as cumulative file size applies, we can remove "2MB per file" and "10 attachments max" limit]

to-do:
- [x] fix backend tests
- [x] test on staging
  - [x] send single attachment larger than 2MB
  - [x] attachments cumulatively larger than 10MB still get blocked
  - [x] more than 10 attachments cumulatively smaller than 10MB can be sent

^@stanleynguyen dk whether there is value in making these part of the e2e tests...

### DIFF
--- a/backend/src/core/config.ts
+++ b/backend/src/core/config.ts
@@ -150,8 +150,6 @@ interface ConfigSchema {
     checkStalledInterval: number
   }
   file: {
-    maxAttachmentSize: number
-    maxAttachmentNum: number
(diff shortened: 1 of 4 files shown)


lisatjide (approved):
ltgm based only the specs of this change.

PR #2058, opened 2023-06-19
feat(frontend): bulk sending ui

stanleynguyen:
## Problem

UI Integration for sending bulk Govsg campaign
## Solution

## Deployment Checklist

- [ ] N/A


frontend/src/components/dashboard/create/email/EmailCredentials.tsx
```diff
@@ -24,7 +24,8 @@ const EmailCredentials = ({
   setActiveStep: Dispatch<SetStateAction<EmailProgress>>
 }) => {
   const { campaign, updateCampaign } = useContext(CampaignContext)
-  const { hasCredential: initialHasCredential, protect } = campaign
+  const { hasCredential: initialHasCredential, protect } =
+    campaign as EmailCampaign
```
tan-zx, line 28: to fix type error?

backend/src/govsg/models/campaign-govsg-template.ts
```diff
@@ -13,7 +13,11 @@ import { GovsgTemplate } from './govsg-template'
 //   but we're creating a pivot table to not have to modify campaign table,
 //   which, in tandem with the old set up of templates for other channels,
 //   will greatly confuse readers
-@Table({ tableName: 'campaign_govsg_template', underscored: true })
+@Table({
+  tableName: 'campaign_govsg_template',
+  underscored: true,
+  timestamps: false,
```
tan-zx, line 19: why don't we want these columns

stanleynguyen: This table is just a pivot table, not that impt for these timestamps me thinks

PR #2060, opened 2023-06-20
fix(frontend): switch back to embedding gogov link

stanleynguyen:


tan-zx (approved):
looks good to me but can test on staging first

PR #2063, opened 2023-06-23
feat(worker): setup govsg message sending worker

stanleynguyen:
## Problem

Closes [insert issue #]

## Solution

## Deployment Checklist

- [ ] Run DB migrations for worker functions
- [ ] Run seed for GOVSG_DEFAULT credentials

### DIFF
--- a/backend/src/core/services/stats.service.ts
+++ b/backend/src/core/services/stats.service.ts
@@ -96,20 +96,24 @@ const getStatsFromTable = async (
       where: { campaign_id: campaignId }, // not sure why it has to be `campaign_id` here
       attributes: [
         [fn('sum', cast({ status: GovsgMessageStatus.Error }, 'int')), 'error'],
-        [fn('sum', cast({ status: GovsgMessageStatus.Sent }, 'int')), 'sent'],
         [
           fn(
             'sum',
             cast(
               {
                 [Op.or]: [
+                  { status: GovsgMessageStatus.Sent },
                   { status: GovsgMessageStatus.Accepted },
-                  { status: GovsgMessageStatus.Unsent },
+                  { status: GovsgMessageStatus.Delivered },
                 ],
               },
               'int'
             )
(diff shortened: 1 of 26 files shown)


worker/src/core/loaders/message-worker/util/govsg.class.ts
```diff
@@ -0,0 +1,132 @@
...
+    logger.info({
+      message: 'Enqueued govsg messages',
+      workerId: this.workerId,
+      jobId,
+      action: 'enqueueMessages',
+    })
+  }
+
+  async getMessages(
+    jobId: number,
```
tan-zx, line 45: **[discussion with Stanley]**
- we will do batch query to flamingo db, put it as a field in the message object to specify which API client to use
- sendMessage function callsite includes this information 

As such:
- use query builder because Sequelize not available on workers
- two different queries, batch queries for efficiency on workers for campaign sending, single query for transactional sending

backend/src/database/migrations/20230621062632-create-enqueue-messages-govsg-function.js
tan-zx, line 1: for the migrations that create db functions, you said you made some optimizations? so it's not just straight up copying the previous worker's db functions?

do you mind describing them briefly (before I review in-depth)?

thanks!

backend/src/core/services/stats.service.ts
```diff
@@ -87,6 +90,51 @@ const getStatsFromTable = async (
   campaignId: number,
   model: any
 ): Promise<CampaignStatsCount> => {
+  if (model === GovsgOp) {
+    const [data] = await model.findAll({
+      raw: true,
+      where: { campaign_id: campaignId }, // not sure why it has to be `campaign_id` here
```
tan-zx, line 96: haha because the class wasn't extended correctly last time, should be fixed now

PR #2064, opened 2023-06-23
feat(SGC-56): transactional govsg message endpoint

tan-zx:
[Ticket](https://linear.app/ogp/issue/SGC-56/govsg-transactional-sending-endpoint)

Todos:
- [ ] Test on staging
  - [x] Remember to revert db migrations if there are changes in the data model (commit hash is `b2ece95c6174cdf134f1dfb778f88d25ee521675`)
  - [ ] Tested once, should test again

Deployment notes:
- [ ] Run database migration
- [ ] Add env vars
  - [ ] `WHATSAPP_NAMESPACE`
  - [ ] `WHATSAPP_LB_URL_1`
  - [ ] `WHATSAPP_LB_URL_2`
  - [ ] `WA_AUTH_TOKEN_ONE`
  - [ ] `WA_AUTH_TOKEN_TWO` (using auth token directly is just a temporary solution, to remove and extract from KMS in the future)
  - [ ]  `FLAMINGO_DB_URI`
- [ ] Need to grant permissions to `flamingo_iam` first (because new table is created). Specifically:
  - [ ] SSH to jumphost (`npm run connectjumphost:${staging|prod}`), connect to RDS via `psql -h postmangovsg-XXXX-aurora.cluster-XXX.ap-southeast-1.rds.amazonaws.com -p 5432 -U XXXX --password --dbname=XXX`
  - [ ] Run `GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO flamingo_iam;`
  - [ ] Run `GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA public TO flamingo_iam;`

For subsequent PRs:
- [ ] Proper params validation + populate API call using user-supplied params
- [x] Call Flamingo to determine which API client to call from (#2067)
- [ ] Convert into proper response for client, update api docs
- [ ] Handling callbacks **[huge; pending update]**
- [ ] Do auth properly by extracting token from KMS + running Lambda **[can do after launch?]**


backend/src/database/migrations/20230608034909-create-govsg-tables.js
```diff
@@ -229,7 +229,7 @@ module.exports = {
       },
       recipient: {
         type: Sequelize.DataTypes.STRING(255),
-        allowNull: true,
+        allowNull: false,
```
tan-zx, line 232: should be false because every tx message will have a recipient.

have edited the migration file directly.

tan-zx:
Works on staging

backend/src/database/migrations/20230623093632-add-userid-column-to-govsg-tx-message.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+module.exports = {
```
stanleynguyen, line 3: I think this one can merge with the prior migrations (from my other PR) as we haven't ran the migrations in prod, lemme do that

tan-zx: let me know when you're done, then I'll rebase

stanleynguyen: done

backend/src/core/routes/index.ts
```diff
@@ -247,6 +253,12 @@ export const InitV1Route = (app: Application): Router => {
     emailTransactionalRoutes
   )
 
+  router.use(
+    '/transactional/govsg',
+    authMiddleware.getAuthMiddleware([AuthType.Cookie, AuthType.ApiKey]),
```
stanleynguyen, line 258: Let's restrict this to `AuthType.ApiKey`

backend/src/core/services/whatsapp.service.ts
```diff
@@ -0,0 +1,8 @@
+import WhatsAppClient from '@shared/clients/whatsapp-client.class'
+import config from '@core/config'
+
+const whatsappClient = new WhatsAppClient(config.get('whatsapp'))
```
stanleynguyen, line 4: A neater implementation (imo) is to do as below here
```
export const whatsappClient = ...
```
then in `services/index.ts` do
```ts
export * as WhatsAppService from './whatsapp.service.ts'
```

backend/src/govsg/middlewares/govsg-transactional.middleware.ts
```diff
@@ -0,0 +1,155 @@
...
+          throw new ApiNotFoundError(
+            `Unable to find id ${govsgTransactionalId} in govsg_messages_transactional`
+          )
+        }
+        const messageId = await GovsgTransactionalService.sendMessage({
+          recipient: govsgTransactional.recipient,
+          templateName: whatsappTemplateLabel,
+          params: govsgTransactional.params,
+        })
+        govsgTransactional.set('status', GovsgMessageStatus.Accepted)
```
stanleynguyen, line 133: Shouldn't this be `SENT`? `ACCEPTED` is for after the saving step isn't it?

shared/src/clients/whatsapp-client.class/index.ts
```diff
@@ -0,0 +1,156 @@
...
+  private credentials: WhatsAppCredentials
+  private axiosInstance: AxiosInstance
+
+  constructor(credentials: WhatsAppCredentials) {
+    this.credentials = credentials
+    this.axiosInstance = axios.create({
+      responseType: 'json',
+      httpsAgent: new https.Agent({
+        // required for connecting to on-prem API client
+        rejectUnauthorized: false,
```
stanleynguyen, line 27: can add `keepAlive` here to reuse the connection

shared/src/clients/whatsapp-client.class/index.ts
```diff
@@ -0,0 +1,156 @@
...
+          components: [],
+        },
+      })
+      .catch((err: Error | AxiosError) => {
+        if (
+          axios.isAxiosError<TemplateMessageErrResponse>(err) &&
+          err.response
+        ) {
+          const { errors } = err.response.data
+          throw new Error(
```
stanleynguyen, line 142: Can we created error types to differentiate errors that are up for downstream users to handle?

backend/src/core/services/whatsapp.service.ts
tan-zx: ohh I was following how it's done for MailService. I won't make this change in the interest of consistency?
image
image

tan-zx: actually the re-exporting in the index.ts is redundant if I'm not wrong

backend/src/govsg/middlewares/govsg-transactional.middleware.ts
tan-zx: so after the saving step, it's `UNSENT`. then it's `ACCEPTED` after it has been accepted by the WA API. then WA will give us a callback to say that it has been sent (see [here](https://developers.facebook.com/docs/whatsapp/on-premises/webhooks/outbound#notification-types) for the list of values).

this is similar to how it's done for tx email API at the moment

shared/src/clients/whatsapp-client.class/index.ts
tan-zx: done

backend/src/govsg/routes/index.ts
```diff
@@ -1,2 +1,3 @@
 export { default as govsgCampaignRoutes } from './govsg-campaign.routes'
 export { default as govsgTemplateRoutes } from './govsg-template.routes'
+export * from './govsg-transactional.routes'
```
tan-zx, line 3: not a fan of default exports, but I see that it's being used inconsistently in other parts of the codebase too

PR #2066, opened 2023-06-25
feat(SGC-101): query flamingo db to select WA API Client

tan-zx:
[Ticket](https://linear.app/ogp/issue/SGC-101/query-flamingo-db-to-determine-wa-api-client-to-use) and [explanation on decision to not use Sequelize here](https://www.notion.so/opengov/RFC-for-WhatsApp-Integration-c04a169adc9d452b9e37aaef8bbd8e72?pvs=4#b60c798c355444dca5c35ec794f38582)

- [x] Test on local
- [x] Test on staging

Deployment notes:
- [x] Make necessary infra changes on Flamingo DB to allow connection (both staging and production)
  - NB for local dev, connect with OGP VPN turned one
- [ ] Add env var `FLAMINGO_DB_URI`


shared/tsconfig.json
```diff
@@ -31,6 +31,7 @@
     "esModuleInterop": true,
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
-    "composite": true
+    "composite": true,
+    "skipLibCheck": true
```
tan-zx, line 35: needed to solve a TS error with Drizzle ORM (suspect it's to do with diverging TS version?)

shared/src/clients/flamingo-db-client.class/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  private dbUri: string
+  constructor(dbUri: string) {
+    this.dbUri = dbUri
+  }
+
+  private async initializeDb(): Promise<NodePgDatabase> {
+    const client = new Client({
+      connectionString: this.dbUri,
+      ssl: {
+        rejectUnauthorized: false,
```
tan-zx, line 16: necessary setting to connect to Flamingo DB

shared/src/clients/flamingo-db-client.class/index.ts
```diff
@@ -0,0 +1,35 @@
...
+      },
+    })
+    await client.connect()
+    return drizzle(client)
+  }
+  public async getAssociatedApiClientId(
+    phoneNumber: string
+  ): Promise<number | null> {
+    const db = await this.initializeDb()
```
tan-zx, line 26: initializing db here to get around the fact that you cannot have an async constructor.

I'm pretty sure it's not a good idea to initialize and connect to the db on each method call (?) but I'm also not sure what's a better way to deal with this. Stopgap solution, should be fine at a small scale, should remove afterwards?

tan-zx: fixed I think

backend/src/govsg/services/govsg-transactional.service.ts
```diff
@@ -0,0 +1,76 @@
...
+    components: params,
+    apiClient,
+    language: WhatsAppLanguages.english,
+  }
+  // differential treatment based on local vs staging/prod
+  // because WA API Client is inaccessible from local
+  const isLocal = config.get('env') === 'development'
+  await WhatsAppService.whatsappClient
+    .validateSingleRecipient(messageToSend, isLocal)
+    .catch((err) => {
```
stanleynguyen, line 51: can we use try/catch here? I have a strong opinion of not mixing Promises and async/await

backend/src/govsg/services/govsg-transactional.service.ts
```diff
@@ -0,0 +1,76 @@
...
+        message: 'Error validating recipient',
+        action,
+        error: err,
+      })
+      throw new ApiInvalidRecipientError(err.message)
+    })
+
+  const messageId = await WhatsAppService.whatsappClient
+    .sendMessage(messageToSend, isLocal)
+    .catch((err) => {
```
stanleynguyen, line 62: similar point to above about try/catch vs Promises

backend/src/govsg/services/govsg-transactional.service.ts
tan-zx: > I have a strong opinion of not mixing Promises and async/await

why/wdym? I think this helps to reduce the indentation and makes it clearer where the error occurs tho

tan-zx:
closing this PR — to reopen a separate PR that doesn't use Drizzle ORM. keep things light and simple first

PR #2075, opened 2023-07-05
feat(backend): add base32 to OTP

stanleynguyen:
## Problem

GTA-17-001 report

## Solution

## Deployment Checklist

- [ ] N/A


tan-zx:
Think from the discussion thread, Tim suggested using [base32](https://en.wikipedia.org/wiki/Base32) — that way, we won't have clash of potentially confusing characters like `O` and `0`. Just a thought.

tan-zx (approved):

cure53:
:heavy_check_mark: LGTM

PR #2076, opened 2023-07-05
feat(backend): add permission check for campaign put and delete

stanleynguyen:
## Problem

GTA-17-004

## Solution

## Deployment Checklist

- [ ] N/A


tan-zx (approved):
looks good to me

cure53:
:heavy_check_mark: LGTM

PR #2070, opened 2023-06-29
feat(SGC-118): handle WhatsApp callback

tan-zx:
[Ticket](https://linear.app/ogp/issue/SGC-118/implement-callback-handling) and [RFC](https://www.notion.so/opengov/WhatsApp-On-Prem-API-Callback-related-research-f7bb4341658048ab8e525b0c32d3ea1a?pvs=25#1bc5fedf832b4e2ab85a82aeadddca9a)

- [x] Test on local (seems to work, accept somehow not updating to `SENT` hmm
- [x] Test on staging
  - [x] Add `WHATSAPP_CALLBACK_VERIFY_TOKEN` env var
  - [x] Change application settings on WA Biz Client to point to prod URL

Deployment notes:
- [ ] Run migration (required for performance, optional for testing)
- [ ] Add `WHATSAPP_CALLBACK_VERIFY_TOKEN` env var
- [ ] Change application settings on WA Biz Client to point to prod URL

PR #2077, opened 2023-07-05
fix: index message_id column of transactional SMS

tan-zx:
[Issue](https://opengovproducts.slack.com/archives/C015VQQ3HMG/p1688552733333679?thread_ts=1688542528.385169&cid=C015VQQ3HMG)

Solution: create index concurrently

Deployment notes:
- [x] Test on local
- [x] Test on staging

### DIFF
--- a/backend/src/database/migrations/20230705102914-index-sms-transactional-message-id-column.js
+++ b/backend/src/database/migrations/20230705102914-index-sms-transactional-message-id-column.js
@@ -1,0 +1,21 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, _) => {
+    await queryInterface.addIndex(
+      'sms_messages_transactional',
+      ['message_id'],
+      {
+        name: 'sms_messages_transactional_message_id_key',
+        unique: true,
+        concurrently: true,
+      }
+    )
+  },
+  down: async (queryInterface, _) => {
+    await queryInterface.removeIndex(
+      'sms_messages_transactional',
+      'sms_messages_transactional_message_id_key'
+    )
+  },
+}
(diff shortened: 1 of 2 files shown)


backend/src/database/migrations/20230705102914-index-sms-transactional-message-id-column.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = {
+up: async (queryInterface, _) => {
```
blakegong, line 4: /nit alignment (you guys have no linting huh 😂)

blakegong (approved):

backend/src/database/migrations/20230705102914-index-sms-transactional-message-id-column.js
tan-zx: lol weird, we do, but somehow it's not being applied. thanks, will fix.

tan-zx:
Previously have tried to add `@Index`, but it doesn't play well with how we set up our test database using `sequelize.sync()`:
- [commit adding `@Index`](https://github.com/opengovsg/postmangovsg/pull/2070/commits/672b6b79e454b7afbcb866843c55a6acc72dca81)
- [commit removing it](https://github.com/opengovsg/postmangovsg/pull/2070/commits/3698734cf37d9ea72bcbf3490e95ee5c83ed91fd)
- error countered:
image

backend/src/sms/models/sms-message-transactional.ts
```diff
@@ -44,6 +44,7 @@ export class SmsMessageTransactional extends Model<SmsMessageTransactional> {
   @Column({ type: DataType.TEXT, allowNull: false })
   body: string
 
+  // indexed for faster search
   @Column({ type: DataType.STRING, allowNull: true })
```
stanleynguyen, line 48: Can add `unique: true` here instead of commenting actually

tan-zx: by right should add `@Index` decorator, but [couldn't get it to work](https://github.com/opengovsg/postmangovsg/pull/2077#issuecomment-1621530115). same issue encountered when I was adding an index for WhatsApp's `message_id` column.

PR #2083, opened 2023-07-06
refactor: move govsg.class.ts

tan-zx:


tan-zx:
addressed here: https://github.com/opengovsg/postmangovsg/pull/2085

PR #2085, opened 2023-07-06
fix(callback): update status at GovsgOp table, use timestamp, enforce update order

tan-zx:
Fixes:
- Update `GovsgOps` table too for callbacks that come in so quickly before the message has been moved out of the ops table
- For Sent callback whose message ids are not found, return 400 error instead of 200. This will trigger a retry. We only do this for Sent because we update `messageId` right after sending.
- Use timestamp in callback, not `new Date()`
- Compare incoming status to current status and only enforce updates in the correct order

For future extension, to prevent excessive retries, we can use Redis and keep track of how many times we have seen a given message id. But unnecessary for now.


stanleynguyen (approved):

tan-zx:
Fixes:
- [ ] Use timestamp in callback, not `new Date()`
- [ ] Compare status and take the latest. Still need to update the timestamp of the corresponding column.

backend/src/govsg/models/govsg-op.ts
```diff
@@ -1,5 +1,69 @@
...
+  Model,
+  Table,
+} from 'sequelize-typescript'
+import { Campaign } from '@core/models'
+import { GovsgMessageStatus } from '@core/constants'
 
 @Table({ tableName: 'govsg_ops', underscored: true, timestamps: true })
-export class GovsgOp extends Model<GovsgMessage> {}
+export class GovsgOp extends Model<GovsgOp> {
+  @ForeignKey(() => Campaign)
```
tan-zx, line 14: identical to `GovsgMessage` — need to duplicate code for Sequelize ORM methods to work. Sequelize sucks.

stanleynguyen (approved):

backend/src/govsg/models/govsg-op.ts
tan-zx: wdym ah you want to commit directly?

PR #2091, opened 2023-07-08
refactor(SGC-141): remove contact validation, improve error handling

tan-zx:
Changes made:
- Contact validation has been deprecated on On-Prem API. So no point having those methods.
- Move `PhoneNumberService` to shared folder to reduce code duplication
- Better error handling so that `INVALID_RECIPIENTS` will be saved to db and show up properly on frontend

Deployment notes:
- [x] Test on staging


worker/src/core/loaders/message-worker/govsg.class.ts
```diff
@@ -245,7 +218,10 @@ class Govsg {
         action: 'sendMessage',
       })
     } catch (error: any) {
-      if ((error as { errorCode: string }).errorCode === 'invalid_recipient') {
+      if (
+        (error as { errorCode: string }).errorCode === 'invalid_recipient' ||
```
tan-zx, line 222: actually not super sure what `error as { errorCode: string }).errorCode === 'invalid_recipient'` is meant to catch and whether it's still necessary

backend/src/govsg/middlewares/govsg-transactional.middleware.ts
```diff
@@ -149,6 +150,18 @@ export const InitGovsgTransactionalMiddleware =
...
           templateName: whatsappTemplateLabel,
           params: req.body.normalisedParams as NormalisedParam[],
+        }).catch((err) => {
+          if (err instanceof ApiInvalidRecipientError) {
+            govsgTransactional.set(
+              'status',
+              GovsgMessageStatus.InvalidRecipient
+            )
+            govsgTransactional.set('errorCode', err.httpStatusCode.toString())
+            govsgTransactional.set('errorDescription', err.errorCode)
```
tan-zx, line 160: actually not super sure whether this is necessary, noticed that it's not done in the db function for campaign messages. but `INVALID_RECIPIENT` should be a type of error too imo.

worker/src/core/loaders/message-worker/govsg.class.ts
```diff
@@ -141,36 +142,8 @@ class Govsg {
...
-          }
-        )
-      }
-
-      const validMessages = validatedWhatsAppIds.filter(
-        (message) => message.status === 'valid'
-      )
 
-      if (validMessages.length === 0) {
+      if (messagesToSend.length === 0) {
```
stanleynguyen, line 146: If no more of whatsapp number validation then we don't need to keep this if condition already as nothing will be filtered

worker/src/core/loaders/message-worker/govsg.class.ts
stanleynguyen: Yup correct

stanleynguyen: We can actually change it to use the new type that u created also

worker/src/core/loaders/message-worker/govsg.class.ts
tan-zx: good point

worker/src/core/loaders/message-worker/govsg.class.ts
tan-zx: ok will do. I'm gonna refactor it I realize there are two identical `normalisePhonenumber` in the code base ugh

tan-zx: actually `error as { errorCode: string }).errorCode === 'invalid_recipient'` will never be caught here, since the normalisation is done at the earlier `getMessage` stage.

stanleynguyen (approved):

PR #2095, opened 2023-07-10
feat: script for sending feedback form

tan-zx:


backend/src/scripts/send-feedback-form.ts
```diff
@@ -0,0 +1,56 @@
...
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${apiKey}`,
+    },
+  })
+  const newFile = path.resolve(
+    __dirname,
+    '../../src/temp',
+    `${fileName}-response.csv`
+  )
+  for (const recipient of recipients) {
```
tan-zx, line 41: for reasons I don't understand, I can't get this to work using `.map` or `.forEach`

tan-zx: ohh this is why
> Using map or forEach would not allow for the use of await within the loop body, which would result in the loop moving on to the next iteration before the response from the API is received.

tan-zx:
@stanleynguyen actually I was thinking, since this doesn't use any of the models/loaders, it could be a standalone script for @phtay2011 to run

tan-zx:
have created a seprate [repo here](https://github.com/opengovsg/sgc-feedback-form)

backend/src/scripts/send-feedback-form.ts
stanleynguyen: `map` can, u need to map them to async function executions then do a `await Promise.all`

PR #2099, opened 2023-07-12
feat: API key expiry reminder email cron job

tan-zx:
[Ticket](https://www.notion.so/opengov/API-key-expiry-notification-email-e997af3cf3564323a5a246e5f804d592)

Note PR squashed into this PR: https://github.com/opengovsg/postmangovsg/pull/2108

Deployment notes
- [x] Add secrets for production 
  - [x] `POSTMAN_DB_URI`
  - [x] `POSTMAN_API_KEY`
  - [x] `CRONITOR_URL_SUFFIX`
  - [x] `CRONITOR_CODE_API_KEY_EXPIRY`
  - [x] `CRONITOR_CODE_REDACTION_DIGEST`
  - [x] `CRONITOR_CODE_UNSUB_DIGEST`
- [ ] Delete replaced Lambda functions via management console


tan-zx:
@stanleynguyen can help to review first while I work out the rest of the stuff

sst/.eslintrc.cjs
tan-zx, line 1: for reasons I don't understand, need to be `.cjs` for it to work properly

sst/.gitignore
```diff
@@ -0,0 +1,15 @@
...
+.sst
+.build
+
+# misc
+.DS_Store
+
+# local env files
+.env*.local
+
+# file for setting secrets via CLI
```
tan-zx, line 14: Note for writing README:

Example command is `res=$(cat secrets.txt) && npx sst secrets set POSTMAN_API_KEY $res`

sst/package.json
tan-zx, line 1: although there are multiple `package-json` files, there is only one `package-lock.json` and all dependencies live in a single `node_modules`. don't really fully understand SST's monorepo setup tbh

sst/packages/functions/src/api-key-expiry/main.ts
```diff
@@ -0,0 +1,93 @@
...
+import { sendEmail } from '@/core/util/email'
+
+import { IS_LOCAL, LOCAL_DB_URI } from '../env'
+
+import { reminderEmailMapper } from './helper'
+
+/*
+- This cron job runs every day at 12AM UTC, i.e. 8AM SGT
+- It picks up API keys that are expiring at exactly 28 days, 14 days, 3 days and 1 day from now
+- DATE_TRUNC("day", valid_until) function round down the valid_until timestamp to the nearest day
```
tan-zx, line 17: actually can just use a `WHERE XXX AND XXX` clause

sst/cdk.context.json
tan-zx, line 1: need to check into Git to ensure deployments work properly

sst/packages/core/src/database/client.ts
tan-zx, line 1: not very sure about the abstraction here, might refactor away in the future

.github/workflows/deploy-sst.yml
```diff
@@ -0,0 +1,59 @@
...
+  set_environment:
+    outputs:
+      current_env: ${{ steps.set_environment.outputs.current_env }}
+    runs-on: ubuntu-latest
+    steps:
+      - id: set-environment
+        run: |
+          echo "Running on branch ${{ github.ref }}"
+          if [ "${{ github.ref }}" = "refs/heads/master" ]; then
+            echo "::set-output name=current_env::production"
```
stanleynguyen, line 22: this is being deprecated. update with https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

.github/workflows/deploy-sst.yml
```diff
@@ -0,0 +1,59 @@
...
+          else
+            echo "::set-output name=current_env::staging"
```
stanleynguyen, line 24: Similar to above

sst/packages/core/src/util/email.ts
```diff
@@ -0,0 +1,31 @@
...
+  from?: string
+  tag?: string
+  // attachments?
+}
+export const sendEmail = async (email: Email) => {
+  const baseURL =
+    SST_STAGE === 'prod'
+      ? 'https://api.postman.gov.sg/v1'
+      : 'https://api-staging.postman.gov.sg/v1'
+  await axios
```
stanleynguyen, line 20: Can we use try/catch here instead?

sst/packages/functions/src/cron-jobs/api-key-expiry/main.ts
```diff
@@ -0,0 +1,93 @@
...
+
+/*
+- This function is triggered by a cron job that runs every day at 12AM UTC, i.e. 8AM SGT
+- It picks up API keys that are expiring 28 days, 14 days, 3 days and 1 day from now
+- DATE_TRUNC("day", valid_until) function round down the valid_until timestamp to the nearest day
+- The cron job will send emails to the API key owners to remind them to renew their API keys
+*/
+
+export async function handler() {
+  const dbUri = IS_LOCAL ? LOCAL_DB_URI : Config.POSTMAN_DB_URI
```
stanleynguyen, line 22: Deployment note: Let's make sure this DB_URI is of a read-replica as we only need that

sst/packages/functions/src/cron-jobs/api-key-expiry/main.ts
```diff
@@ -0,0 +1,93 @@
...
+import { reminderEmailMapper } from './helper'
+
+/*
+- This function is triggered by a cron job that runs every day at 12AM UTC, i.e. 8AM SGT
+- It picks up API keys that are expiring 28 days, 14 days, 3 days and 1 day from now
+- DATE_TRUNC("day", valid_until) function round down the valid_until timestamp to the nearest day
+- The cron job will send emails to the API key owners to remind them to renew their API keys
+*/
+
+export async function handler() {
```
stanleynguyen, line 21: Can we wrap this function body around a try/catch and log the error out before letting it bubble and fail cronitor?

stanleynguyen (commented):
Just a few comments left. Can we also look into pipe logs from these lambdas to datadog if it's not too mafan? Otherwise let's create a separate task and look into that later

sst/packages/functions/src/cron-jobs/api-key-expiry/main.ts
tan-zx: so for [the next PR](https://github.com/opengovsg/postmangovsg/pull/2108), we actually have to UPDATE the table. 

do you want to use two different `DB_URI`s or should we just stick to one

sst/packages/core/src/util/email.ts
tan-zx: hmm there's no difference right? since the only thing that happens in this function is sending the email

sst/packages/functions/src/cron-jobs/api-key-expiry/main.ts
tan-zx: good idea

tan-zx:
> Can we also look into pipe logs from these lambdas to datadog if it's not too mafan? Otherwise let's create a separate task and look into that later

let's create separate task for that hmm

tan-zx:
quite significant refactoring https://github.com/opengovsg/postmangovsg/pull/2108, might be good to review that directly instead

.github/workflows/deploy-sst.yml
```diff
@@ -0,0 +1,59 @@
...
+  set_environment:
+    outputs:
+      current_env: ${{ steps.set_environment.outputs.current_env }}
+    runs-on: ubuntu-latest
+    steps:
+      - id: set-environment
+        run: |
+          echo "Running on branch ${{ github.ref }}"
+          if [ "${{ github.ref }}" = "refs/heads/master" ]; then
+            echo "current_env=production"
```
stanleynguyen, line 22: missing `>> $GITHUB_OUTPUT`

sst/packages/core/src/util/email.ts
stanleynguyen: Materially no diff as there's literally only this function call here so it's rather a nit but I'd rather keep it consistent across the codebase to not use `then.catch` where possible

.github/workflows/deploy-sst.yml
tan-zx: good point will fix.

sst/packages/core/src/util/email.ts
tan-zx: > but I'd rather keep it consistent across the codebase to not use `then.catch` where possible

I'd like to push back against this though. I think `await fnCall().catch()` is a better pattern then `try... catch` the entire thing, because it forces you to think specifically about the I/O boundary where the error would occur.

sst/packages/functions/src/cron-jobs/api-key-expiry/cron.ts
```diff
@@ -0,0 +1,111 @@
...
+      return
+    }
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const cronitor = require('cronitor')(Config.CRONITOR_URL_SUFFIX) // common to all jobs on our shared Cronitor account
+    const invokeFunction = cronitor.wrap(
+      Config.CRONITOR_CODE_API_KEY_EXPIRY,
+      async function () {
+        await sendApiKeyExpiryEmail()
+      },
+    )
```
stanleynguyen, line 26:
```suggestion
    // eslint-disable-next-line @typescript-eslint/no-var-requires
    const cronitor = require('cronitor')(Config.CRONITOR_URL_SUFFIX) // common to all jobs on our shared Cronitor account
    const invokeFunction = process.env.IS_LOCAL === 'true' 
      ? async function() {
        console.log('Running cron locally')
        await sendApiKeyExpiryEmail()
      }
      : cronitor.wrap(
         Config.CRONITOR_CODE_API_KEY_EXPIRY,
          async function () {
            await sendApiKeyExpiryEmail()
          },
        )
```

tan-zx: got it thanks will test

stanleynguyen (approved):

PR #2111, opened 2023-07-20
fix(scripts): lowercase email to avoid duplicate

stanleynguyen:


tan-zx (approved):
looks good to me

tan-zx:
probably need to extend this script to allocate permission to the templates

PR #2093, opened 2023-07-10
feat: whitelist Govsg templates on per user basis

tan-zx:
Todo:
- [x] Migration script to grant permission to existing users 
- [x] Test on staging

Deployment notes
- [x] Run db migration
- [x] Grant permission to new table (run those Postgres commands)
- [x] Run migration script to ensure current users retain existing permissions
- [x] Remove script from this PR before merging into db
- [ ] Deploy to prod

PR #2112, opened 2023-07-20
feat: allow auto unsubscribe to Phonebook list

ganchinyao:
## Problem

The main goal of this PR is to replace the `unsubscribe` link of emails sent out to point to Phonebook domain and to handle the unsubscribe logic on Phonebook domain. Then, when MOP unsubscribed, they will be automatically removed from the Phonebook list without needing the agency to manually remove them. Only new email campaign using Phonebook will be affected. Email campaign not using Phonebook will default back to the current unsubscribe logic.

## Solution

1. To identify if a new email campaign is using Phonebook, we store campaign_id<>managed_list_id association whenever the agency user chooses a Phonebook list. Whenever the agency user uploads a CSV instead, we remove this association. This association is store in `managed_list_campaigns` table.
2. To check if a campaign is using Phonebook, check the `managed_list_campaigns` table to see if there is entry
3. Modify the worker code to first check if the campaign is using Phonebook. If it is, query the Phonebook API to get a custom unsubscribe link, and use this link instead.
4. After sending the campaign, if campaign is using Phonebook, FE will not show `Unsubscribers` and Unsubscribers info section in the report page. Also remove away `unsubscribeReason` in the delivery report if campaign is using Phonebook.
5. Each time when user enters the `Upload recipients` page, for both SMS and Email, we check if the current campaign has already selected a Phonebook list before (because user can select and navigate away). If the user has, we pre-populate the Phonebook Dropdown with that list, and use that list (current logic).

## Screenshots

| Item | Before | After |
| ----- | ------ | ----- |
| No unsubscribe information for campaign using Phonebook | Screenshot 2023-07-20 at 6 02 29 PM | Screenshot 2023-07-20 at 5 56 43 PM |
| Inform user of auto unsubscription | Screenshot 2023-07-20 at 5 59 42 PM | Screenshot 2023-07-20 at 5 59 51 PM |

## Test

Tested the following scenarios:

1. Use phonebook in Email campaign => Emails sent out will contain phonebook unsubscribe URL. No "Unsubscribe" information section in Report page.
2. Use CSV in Email campaign => Emails sent out will contain postman unsubscribe URL (current flow).
3. Toggle FE and worker feature flag off => Even if user uses phonebook in Email campaign => Will use postman unsubscribe URL (current flow)
4. When selecting a Phonebook list in `Upload Recipients` stage in Email/SMS and navigate away and back to the page => Phonebook list selected is populated by default

## Deployment Checklist

- [x] FE: Set .env variable: `REACT_APP_PHONEBOOK_ENABLE_AUTO_UNSUBSCRIBE`
- [x] Worker: Set config variable `PHONEBOOK_ENABLE_AUTO_UNSUBSCRIBE`


backend/src/core/models/managed-list-campaign.ts
```diff
@@ -0,0 +1,33 @@
...
+  @ForeignKey(() => Campaign)
+  @Column({
+    type: DataType.INTEGER,
+    allowNull: false,
+    primaryKey: true,
+  })
+  campaignId: number
+
+  @BelongsTo(() => Campaign)
+  campaign!: Campaign
```
tan-zx, line 26: the `!` is not necessary

backend/src/core/models/managed-list-campaign.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  @Column({
+    type: DataType.INTEGER,
+    allowNull: false,
+  })
+  managedListId!: number
```
tan-zx, line 32: `!` not necessary.

[nit] since this is an association table, would make sense to use a composite primary key across both `campaignId` and `managedListId`? see [this](https://github.com/opengovsg/postmangovsg/pull/2093/files#diff-ddbd590b554e5ebb98f8cdc53cfa3dc2a6ba6ae011a79a19c3c62b93671f4468) for an example

tan-zx (commented):
took a look at backend code, will review the rest later

backend/src/core/models/managed-list-campaign.ts
ganchinyao: Okay I will remove it

backend/src/core/models/managed-list-campaign.ts
ganchinyao: Yes `managedListId` is fetched from Phonebook API (existing).

>  how do we ensure that the user cannot arbitrarily associate their campaign with any manageListId?

That's a good point, thanks. I will add a BE check for the API by calling existing Phonebook API to check that the managedListId is indeed owned by the user.

As for the association table, yes one campaign can only be associated with at most one `managedListId`. One `managedListId` can be associated with multiple different campaigns. I think a composite primary key is not necessary since a primary key on campaignid would suffice. We could potentially implement this by the nullable column that you suggested. I tried to avoid that as I want to try to avoid making changes to the existing table as far as possible, lest it breaks something. Also, adding to a new table has it's advantages. When we search whether a campaign is using a `managedList`, the new table to search for is much smaller than the whole `campaigns` table, since it only consists of campaigns that are using `managedListId`.

tan-zx: > When we search whether a campaign is using a managedList, the new table to search for is much smaller than the whole campaigns table, since it only consists of campaigns that are using managedListId.

> I will add a BE check for the API by calling existing Phonebook API to check that the managedListId is indeed owned by the user.

why I think an "association table" is a bad idea:
- you don't really need an association table for a one-to-one relationship mapping
- lookup speed should not matter, since the campaign table is indexed by `id` alr
- if you implement the BE check described, presumably. you'd need to pass the `userId` or the user's email to Phonebook API. if you store this in a a separate table, you would need to do another join with the campaign table anyway.

> I tried to avoid that as I want to try to avoid making changes to the existing table as far as possible, lest it breaks something.

I am pretty sure adding a nullable column to the campaign table won't break anything :)

tan-zx (approved):

backend/src/core/models/managed-list-campaign.ts
ganchinyao: Have discussed this, we will stick to a new table for now

PR #2113, opened 2023-07-20
docs(govsg): touch up govsg transactional endpoint

stanleynguyen:


backend/openapi.yaml
```diff
@@ -2124,17 +2127,32 @@ paths:
...
-                whatsapp_template_label:
+                  description: >
+                    Mobile phone number of the recipient without special formatting 
+                    (only contains numerical characters and prefixed with plus 
+                    sign if country code is included).
+
+
+                    **Notes**:
+                      - Country code will need to be provided for non-SG mobile numbers.
+                      - If country code is not provided
```
tan-zx, line 2143: think it's incomplete here

backend/openapi.yaml
```diff
@@ -4421,13 +4439,36 @@ components:
...
       properties:
+        id:
+          type: number
         body:
           type: string
         params:
           type: array
           items:
             type: string
+        multilingual_support:
```
tan-zx, line 4455: don't really understand how this part works, but I'm guessing our own template actually abstracts across multiple WA's templates?

stanleynguyen: Yup so basically it's an array that contains information about the template in alternative languages, which only differ in: 
- `language_code` (e.g. `en_GB`, `zh_CN`, etc)
- `language`: This is mainly for frontend display purposes (e.g. `English (UK)`, `Chinese (CN)`, etc)
- `body`: template body in the alternative language

tan-zx (approved):

PR #2114, opened 2023-07-20
refactor: remove eb-env-update

tan-zx:
removing this lambda function

- not useful, only applies to Elastic Beanstalk backend (and doesn't apply to the ECS worker)
- not much effort saved

deployment note:
- [ ] Remove corresponding Cloudwatch rule from management console
- [ ] Remove Lambda from AWS console


stanleynguyen (approved):

stanleynguyen:
@tan-zx : 
- Add to checklist to remove the existing Lambda and all of its code versions from our AWS as well because we're paying to store those

PR #2115, opened 2023-07-24
feat: allow users to exclude params field

tan-zx:
[ticket](https://opengovproducts.slack.com/archives/CT8D1FESY/p1690190255044149)

Todo
- [x] Test on staging
- [x] Update guide

Deployment notes: N/A


stanleynguyen (approved):

stanleynguyen:
Lambda deployment failed as there're some active development to deprecate those functions

tan-zx:
> Lambda deployment failed as there're some active development to deprecate those functions

yup, should only affect the staging ones, production ones are unaffected

PR #2117, opened 2023-07-25
feat: officer verification by MOP

jia1:
## Problem

Closes [SGC-154](https://linear.app/ogp/issue/SGC-154/321-officer-verification-by-mop-p1)

## Solution

- On `sgc_notify_upcoming_call_1` delivered, run `sendPasscodeCreationMessage` and `storePrecreatedPasscode`.
- On button reply, send the correct passcode.
- In the frontend, show the Messages tab.

### Checklist

- [x] Persist passcode
- [x] Show the Messages tab + simple table
- [x] Create endpoint to get messages given campaign + include `passcode` column from `GovsgVerification`
- [x] Show messages from backend
- [x] Handle button replies

### Screenshots

image
image


backend/src/govsg/models/govsg-verification.ts
```diff
@@ -0,0 +1,29 @@
...
+  Column,
+  DataType,
+  ForeignKey,
+  Model,
+  Table,
+} from 'sequelize-typescript'
+import { GovsgMessage } from './govsg-message'
+
+@Table({ tableName: 'govsg_verification', underscored: true, timestamps: true })
+export class GovsgVerification extends Model<GovsgVerification> {
```
stanleynguyen, line 11: The passcode - govsg_message relation is 1-1 right? Do we actually need another model for this? Or would new fields on govsg_message model suffice?

jia1: Not exactly because `govsgMessageId` comes from the initial message with `params` and `passcodeCreationWamid` will be populated by the 2nd message with the quick-reply button.

As for allowing users to see non-send/delivery of the `passcodeCreationWamid`, I would want to think of this later.

jia1: And it won't be easy adding passcode as a column because:

- Not all messages will have `passcode` and `passcodeCreationWamid`.
- Having a separate relation is cleaner - in this new relation, we can establish everything as non-null because all values can be defined before creation. This is not the case when we add columns to `govsg_messages`. Let me make a commit to set `allowNull: false`.

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -59,7 +71,10 @@ const parseWebhook = async (
       message: 'Received status webhook from WhatsApp',
       action,
     })
-    await parseTemplateMessageWebhook(body as WhatsAppTemplateMessageWebhook)
+    await parseTemplateMessageWebhook(
```
KishenKumarrrrr, line 74: Do we want to remove the log above this on lines 69 - 73? (Sorry, I couldn't tag on the line itself)

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -59,7 +71,10 @@ const parseWebhook = async (
...
+      body as WhatsAppTemplateMessageWebhook,
+      clientId
+    )
```
KishenKumarrrrr, line 77: Sorry, can't tag on the line again. 

But on line 89: `if ('errors' in body) ...`, am I correct to say that `errors` cannot exist with the other types and thats why we don't have to check it before checking for valid types? Also, why don't we throw an error for this case?

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -88,8 +103,45 @@ const parseWebhook = async (
...
 }
 
+const sendPasscodeCreationMessage = async (
+  whatsappId: WhatsAppId,
+  clientId: WhatsAppApiClient
+): Promise<MessageId> => {
+  const templateMessageToSend: WhatsAppTemplateMessageToSend = {
+    recipient: whatsappId,
+    apiClient: clientId,
+    templateName: 'sgc_passcode_generation', // TODO: Un-hardcode this
```
KishenKumarrrrr, line 113: Could I ask why we expect this value to change?

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -88,8 +103,45 @@ const parseWebhook = async (
...
+    params: [],
+    language: WhatsAppLanguages.english,
+  }
+  const isLocal = config.get('env') === 'development'
```
KishenKumarrrrr, line 117: Curious, why do we need this?

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -88,8 +103,45 @@ const parseWebhook = async (
...
+      isLocal
+    )
+  return passcodeCreationWamid
+}
+
+const createPasscode = () => {
+  return randomInt(0, Math.pow(10, 4)).toString().padStart(4, '0')
+}
+
+const storePrecreatedPasscode = async (
```
KishenKumarrrrr, line 130: Could I ask why we consider this precreated? Doesn't this value get generated on demand? (And thus we can supply `passcodeCreationWamid` as a param)?

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -88,8 +103,45 @@ const parseWebhook = async (
...
+  govsgMessageId: GovsgMessage['id'],
```
KishenKumarrrrr, line 131: Just to confirm, `govsgMessageId` is the initial message sent to the user? If so, how do we intend to handle the case where we send multiple messages for the different languages?

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -226,6 +283,42 @@ const parseTemplateMessageWebhook = async (
       void govsgMessage?.update(fieldOpts, whereOpts)
       void govsgMessageTransactional?.update(fieldOpts, whereOpts)
       void govsgOp?.update(fieldOpts, whereOpts)
+      // TODO: Add logging statement here for traceability
```
KishenKumarrrrr, line 286: Did you mean to handle logging in this PR?

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -226,6 +283,42 @@ const parseTemplateMessageWebhook = async (
...
+      if (!govsgMessage && !govsgOp) {
```
KishenKumarrrrr, line 287: Do we need to throw an error?

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -226,6 +283,42 @@ const parseTemplateMessageWebhook = async (
...
+      }
+      const message = (govsgMessage ?? govsgOp) as GovsgMessage
+      // TODO: Add logging statement here for traceability
+      const experimentalData = await experimentService.getUserExperimentalData(
+        message.campaign.userId
+      )
+      // TODO: Add logging statement here for traceability
+      const canAccessGovsgV = `${ChannelType.Govsg}V` in experimentalData
+      if (!canAccessGovsgV) {
+        return
```
KishenKumarrrrr, line 298: Similar to above

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -259,12 +352,87 @@ const parseTemplateMessageWebhook = async (
   }
 }
 
+const sendPasscodeMessage = async (
```
KishenKumarrrrr, line 355: Can KIV for now but I think in the future we would need to perform a refactor to distinguish between whatsapp and sms functions?

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -259,12 +352,87 @@ const parseTemplateMessageWebhook = async (
...
   const { id: messageId, type } = body.messages[0]
+  if (type === WhatsappWebhookMessageType.button) {
+    const message = body.messages[0] as WhatsAppWebhookButtonMessage
+    if (message.button.text === 'Create passcode') {
+      const passcodeCreationWamid = message.context.id
+      const govsgVerification = await GovsgVerification.findOne({
+        where: { passcodeCreationWamid },
+        include: [GovsgMessage],
+      })
+      if (!govsgVerification) {
```
KishenKumarrrrr, line 405: Would this be considered an erroneous state? Do we need to explicitly throw an error?

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -259,12 +352,87 @@ const parseTemplateMessageWebhook = async (
...
+        // TODO: Update log statement
```
KishenKumarrrrr, line 406: Did you want to update this in this PR?

frontend/src/components/common/title-bar/TitleBarWithTabs.tsx
```diff
@@ -0,0 +1,35 @@
+import type { ReactNode } from 'react'
+
+import defaultStyles from './TitleBar.module.scss'
+
+const HStackStyle = {
```
KishenKumarrrrr, line 5: Is there a reason the naming convention is different from the variable below? (`subsequentElementStyle`)

frontend/src/components/common/title-bar/TitleBarWithTabs.tsx
```diff
@@ -0,0 +1,35 @@
...
+  overrideStyles,
+}: {
+  children?: ReactNode
+  title: string
+  overrideStyles?: { readonly [key: string]: string }
+}) => {
+  const styles = overrideStyles || defaultStyles
+  return (
+    <div className={styles.titleBar}>
+      {!!children && Array.isArray(children) && children.length === 2 && (
```
KishenKumarrrrr, line 24: Could I ask what `!!` does?

frontend/src/components/dashboard/create/govsg/GovsgMessages.tsx
```diff
@@ -0,0 +1,212 @@
...
+    const response = await axios.get(`/campaign/${campaignId}/govsg/messages`, {
+      params: {
+        ...options,
+      },
+    })
+    const { messages: govsgMessages, total_count: totalCount } = response.data
+    setGovsgMessageCount(totalCount)
+    setGovsgMessagesDisplayed(govsgMessages)
+    setSelectedPage(selectedPage)
+    if (!hasFetchedGovsgMessages) setHasFetchedGovsgMessages(true)
```
KishenKumarrrrr, line 53: Do we need to check this if the value of `hasFetchedGovsgMessages` will be `true` regardless?

frontend/src/components/dashboard/create/govsg/GovsgMessages.tsx
```diff
@@ -0,0 +1,212 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedPage])
+
+  const handlePageChange = (index: number) => {
+    void fetchGovsgMessages(index)
+  }
+
+  const columns = [
+    {
+      name: 'Recipient name'.toUpperCase(),
```
KishenKumarrrrr, line 67: Does`.toUpperCase()` transform the column name only? If not, why don't we just type the name in caps?

backend/src/govsg/middlewares/govsg-verification.middleware.ts
```diff
@@ -0,0 +1,52 @@
...
+  req: Request,
+  res: Response
+): Promise<Response | void> => {
+  const { campaignId } = req.params
+  const { offset, limit } = req.query
+  const { rows, count } = await GovsgMessage.findAndCountAll({
+    where: {
+      campaignId: +campaignId,
+    },
+    offset: +(offset as string),
```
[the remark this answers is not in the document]
jia1: Thanks, Kishen!

What sort of defensive checks are you looking at? Is it type checking only or do you have something else in mind? I would see a need for more sanitisation if `campaignId` were directly injected into an SQL statement but in this case, it's a programmatic query.

Let me know what you think!

backend/src/govsg/middlewares/govsg-verification.middleware.ts
```diff
@@ -0,0 +1,52 @@
...
+  })
+  const messages = rows.map((row) => {
+    const {
+      agency,
+      recipient,
+      officer_name,
+      recipient_name,
+      officer_designation,
+      ...remainingParams
+    } = row.params as any
```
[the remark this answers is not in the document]
jia1: Nice catch. I'll change it to: `Record<string, string>`.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
```diff
@@ -0,0 +1,52 @@
...
+    console.debug(
```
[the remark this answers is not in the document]
jia1: Hmm..., I meant it to avoid having to declare:

```
// eslint-disable-next-line @typescript-eslint/no-unused-vars
```

But the intent isn't obvious at all, so let me remove it and add the necessary `eslint-disable`.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
```diff
@@ -0,0 +1,52 @@
...
+        row.params
+      )}`
+    )
+    return {
+      ...row.get({ plain: true }),
+      name: recipient_name,
+      mobile: recipient,
+      data: JSON.stringify(remainingParams),
+      passcode: row.govsgVerification?.passcode ?? '****',
+      sent: row.sendAttemptedAt,
```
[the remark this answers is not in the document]
jia1: Sorry, I don't understand... need you to paraphrase your qn... ><

backend/src/govsg/models/govsg-message.ts
```diff
@@ -5,19 +5,32 @@ import {
...
 export class GovsgMessage extends Model<GovsgMessage> {
+  @Column({
+    type: DataType.BIGINT,
+    autoIncrement: true,
+    primaryKey: true,
+    allowNull: false,
+  })
+  id: number
+
   @ForeignKey(() => Campaign)
```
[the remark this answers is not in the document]
jia1: I left this alone for fear of potential backward incompatibility in production but I can investigate if this change is allowed...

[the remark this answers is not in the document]
jia1: Safe for now. Let me add the constraint.

backend/src/govsg/routes/govsg-campaign.routes.ts
```diff
@@ -125,4 +127,16 @@ router.get(
   GovsgStatsMiddleware.getDeliveredRecipients
 )
 
+router.get(
+  '/messages',
+  celebrate({
```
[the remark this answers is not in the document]
jia1: Thanks, Kishen!

I'll probably not take up on this offer as this entire file has been written in this style and I appreciate the locality of this information.

This is also currently inconsistent across the entire codebase. A rule-of-thumb I'm keen to take up would be: Extract only if too long.

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -10,20 +12,30 @@ import {
...
+  GovsgTemplate,
+} from '@govsg/models'
+import {
+  ChannelType,
   GovsgMessageStatus,
   govsgMessageStatusMapper,
   shouldUpdateStatus,
 } from '@core/constants'
-import { WhatsAppService } from '@core/services'
+import { WhatsAppService, experimentService } from '@core/services'
```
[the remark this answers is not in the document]
jia1: I'm actually also curious why it's inconsistent.

Refactoring one of them (I would what to de-capitalise `WhatsAppService`) would be nice but I would prefer to do this in a separate PR so as to not increase the diff anymore.

Hope I remember...

backend/src/govsg/services/govsg-callback.service.ts
jia1: Good qn. It's fine to have more logs actually... to improve traceability (I learnt this from the fintech team).

The logs could be improved to make troubleshooting even easier, but I might want to defer this cognitive task to a later PR.

Wdyt?

backend/src/govsg/services/govsg-callback.service.ts
jia1: I think GitHub doesn't allow you to tag multiple lines? I couldn't do it either iirc.

About `if ('errors' in body)`, we probably don't know much more than the docs [here](https://developers.facebook.com/docs/whatsapp/on-premises/guides/webhooks/#errors).

We don't throw errors because this is the expected way of receiving error notifications from WhatsApp. If we throw an error here, [Meta will resend the request](https://developers.facebook.com/docs/whatsapp/on-premises/guides/webhooks/#retry).

backend/src/govsg/services/govsg-callback.service.ts
jia1: Whenever we want to use a different template to enable passcode creation. This can happen on a whim, so it'll be ideal to not have to make a commit to change this value.

Also, when we update a template, it remains unavailable until re-approved. As for how long it takes for a template to be re-approved, it took 1 working day for the prototyping WABA and I am unsure how long it takes for the production WABA.

Will love to address this TODO in the next PR. Let me make a Linear task for this.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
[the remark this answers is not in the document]
KishenKumarrrrr: Yup, I was thinking of type checking. Since the query contains some typecasting and an implicit assumption that `campaignId` is a number, I was wondering if it would be better to explicitly check for these before executing the db query. 

Though I suppose either way we'd get an error thrown with the current implementation?

backend/src/govsg/middlewares/govsg-verification.middleware.ts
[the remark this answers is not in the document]
KishenKumarrrrr: Sorry I don't really understand. Why does `console.debug` avoid having to disable eslint?

backend/src/govsg/services/govsg-callback.service.ts
jia1: If you Cmd+Click `sendTemplateMessage`, you'll see that the headers change accordingly.

In dev env, requests have to go through a proxy because your local machine and WhatsApp server are in different network security zones.

In non-dev env, both servers are in the same security zone.

backend/src/govsg/services/govsg-callback.service.ts
jia1: It's pre-created in our perspective because `sgc_passcode_generation` is the template that comes with the "Create passcode" button. The user has not told us that he/she is creating a passcode yet, but we already have it pre-created.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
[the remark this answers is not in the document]
KishenKumarrrrr: oops, initially what I meant was why didnt we destructure `row.sendAttemptedAt` along with the other `row.params` on line 25. But I now realise that they are completely different things and hence there is no need to destructure. Can ignore my comment 😅

backend/src/govsg/routes/govsg-campaign.routes.ts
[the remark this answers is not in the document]
KishenKumarrrrr: Thats a fair rule-of-thumb. I think can just leave it then 👍

backend/src/govsg/services/govsg-callback.service.ts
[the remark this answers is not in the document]
KishenKumarrrrr: Sure, let me know if you're busy, I can help make the changes too.

backend/src/govsg/services/govsg-callback.service.ts
jia1: Yes.

To be frank, for the `multiple messages for the different languages`, I would like to think about it afterwards. Otherwise, there are too many concerns to address in this feature.

But as a start, there are a few things to consider:

- Restrict language to at most 1
- Restrict language to at most 2 and 1 of them must be English
- Associate `GovsgVerification` with only 1 of the messages in a certain order of languages

backend/src/govsg/services/govsg-callback.service.ts
jia1: Tbh I got a bit tired on Friday. Could I clean this up in a subsequent PR? This is because I want to think about what meaningful things can be logged. This line was originally something less useful like "Checking govsgMessage".

backend/src/govsg/services/govsg-callback.service.ts
jia1: I can throw, but I will need to handle it.

I suppose I should add a log statement here, if this scenario is really unexpected.

We shouldn't throw errors in response to the webhook request as our failure to process data is not their concern.

[Linear](https://linear.app/ogp/issue/SGC-171/add-logging-to-if-govsgmessage-andand-govsgop)

backend/src/govsg/services/govsg-callback.service.ts
jia1: This one is fine because it just means the user ain't got access to our new feature. It's a bit too bad for them.

backend/src/govsg/services/govsg-callback.service.ts
jia1: Agree to KIV.

backend/src/govsg/services/govsg-callback.service.ts
KishenKumarrrrr: Yup, I think its fine to leave cleaning up the logging in a different PR.

wrt having more logs, I think in this case, we might not want the log on lines 69-73 since `parseTemplateMessageWebhook` itself has logging for the different states it encounters (although the logging coverage is not complete)

Perhaps for now we can leave this particular log in. Once we have capacity to clean the logs, then we should remove it since the logging inside `parseTemplateMessageWebhook` is more specific.

backend/src/govsg/services/govsg-callback.service.ts
jia1: I suppose this could be erroneous because it means somehow someone managed to tap on a `Create passcode` button when `GovsgVerification` had no record of the message.

This can happen when:

- (Erroneous) We failed to create `GovsgVerification` after sending the passcode creation message.
- (Mischief...? But I guess it's still erroneous?) Someone sent the passcode creation message or a message containing a `Create passcode` button outside of Postman's ecosystem.

Let me change `logger.warn` to `logger.error`.

backend/src/govsg/services/govsg-callback.service.ts
jia1: Nope if possible

frontend/src/components/common/title-bar/TitleBarWithTabs.tsx
jia1: `HStack` is actually a borrowed term from chakra.

I'll rename to `horizontalStackStyle`.

frontend/src/components/common/title-bar/TitleBarWithTabs.tsx
jia1: [Here you go](https://www.joshwcomeau.com/operator-lookup?match=double-not)

frontend/src/components/dashboard/create/govsg/GovsgMessages.tsx
jia1: Hmm, I copied this from `Campaigns.tsx` rather blindly but sure, let's remove the check.

frontend/src/components/dashboard/create/govsg/GovsgMessages.tsx
jia1: Sure, I can do that.

backend/src/govsg/services/govsg-callback.service.ts
KishenKumarrrrr: I think the lines can't be tagged because it was not changed in this PR. 

To me, the docs suggest that error notifications are a specific type of message and so we won't have a case where `'errors' in body` is true and yet there is other content we need to parse. 

If we don't throw errors because it will result in meta resending the request, then why do we throw an error on line 103? In this case, wouldn't meta just resend this erroneous message indefinitely? 😨 

Also if we receive an error via webhook, how is this handled on the FE though? It currently looks to me like this error is not displayed anywhere and silently happens. (i.e, someone using the FE would never know that this error occured?)

backend/src/govsg/services/govsg-callback.service.ts
KishenKumarrrrr: Sure, we can discuss the implementation details for this when we embark on this task.

backend/src/govsg/services/govsg-callback.service.ts
KishenKumarrrrr: I see, thanks for explaining it to me 😄

backend/src/govsg/middlewares/govsg-verification.middleware.ts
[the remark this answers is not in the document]
jia1: The `console.debug` "uses" the variable and would prevent the need to disable eslint on the `no-unused-vars` rule.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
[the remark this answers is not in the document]
jia1: Yup, `+` catches that.

backend/src/govsg/services/govsg-callback.service.ts
KishenKumarrrrr: So what you mean is that we create the passcode as soon as the message is sent, instead of when the user clicks on the button?

backend/src/govsg/services/govsg-callback.service.ts
KishenKumarrrrr: Sure, I don't think the paths you suggested would radically change the implementation for this too, thanks!

backend/src/govsg/services/govsg-callback.service.ts
KishenKumarrrrr: Yup no problem. I think above we already discussed another PR for cleaning up logging stuff, we can just revisit all these whenever we can.

backend/src/govsg/services/govsg-callback.service.ts
KishenKumarrrrr: Yup, I think adding a log would be sufficient. Thanks!

frontend/src/components/common/title-bar/TitleBarWithTabs.tsx
KishenKumarrrrr: Thank you!

jia1:
@KishenKumarrrrr I removed the TODO that ask to add log statements. Too much cognitive load. I'll log when the need arises.

backend/src/govsg/services/govsg-callback.service.ts
jia1: I see, I'll remove it in the next PR. I don't have opinions on this so I don't mind just getting it done next.

backend/src/govsg/services/govsg-callback.service.ts
jia1: There's some subtle difference. `errors` is something Meta can send. Everything else is unexpected.

As for whether we should throw `UnexpectedWebhookError`:

For:

- We don't want them to keep resending
- Be alerted of potential attackers

Against:

- Breaking change from Meta and we want to handle the pending requests before they expire (3 days)

I'm not keen to make the decision now, but I can bring this discussion to Slack if that helps.

backend/src/govsg/services/govsg-callback.service.ts
jia1: Yes!

KishenKumarrrrr (approved):
Thanks for taking the time to address all my comments! LGTM 👍

PR #2120, opened 2023-07-26
feat(backend): sending endpoint for govsg txnal API

stanleynguyen:
## Problem

Touch up txnal api sending endpoint to new specs

## Solution

Implement it

## Deployment Checklist

- [ ] Run DB migration


backend/src/govsg/services/govsg-transactional.service.ts
```diff
@@ -47,34 +48,36 @@ async function sendMessage({
...
-          err.message + `. Recipient: ${recipient}`
-        )
-      }
-      logger.error({
-        message: 'Error sending message',
-        action,
-        error: err,
-      })
-      throw new ApiBadGatewayError(err.message)
+  try {
```
tan-zx, line 56: you really don't like the `await...catch` pattern haha

tan-zx (approved):

PR #2121, opened 2023-07-26
feat/improve-whitelist-script

KishenKumarrrrr:
## Problem

The logging in the existing whitelisting script is insufficient if we are adding a large number of rows. As such, it can be difficult to determine if our script ran correctly.

## Solution

I've added additional metrics to log as well as improved the formatting of the logs. 
The additional metrics added are:
- Number of rows in the raw data to be added
- Number of users that already have a postman account
- Emails of users that already had a postman account
- Number of new user accounts created
- Number of rows inserted into User Experimental table
- Number of rows inserted into Govsg Template Access table

### DIFF
--- a/backend/src/scripts/insert_experimental_govsg_user.ts
+++ b/backend/src/scripts/insert_experimental_govsg_user.ts
@@ -19,6 +19,9 @@ type RawOfficerData = {
   id?: number
 }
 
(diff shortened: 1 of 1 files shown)


jia1 (approved):

jia1:
LGTM

PR #2122, opened 2023-07-27
feat(backend): add govsg txn GET message by ID

stanleynguyen:
Transactional GovSG get message by ID endpoint

## Deployment Checklist

- [ ] N/A

### DIFF
--- a/backend/src/email/middlewares/email-transactional.middleware.ts
+++ b/backend/src/email/middlewares/email-transactional.middleware.ts
@@ -311,7 +311,7 @@ export const InitEmailTransactionalMiddleware = (
     }
   }
 
-  async function getById(req: Request, res: Response): Promise<void> {
+  async function getById(req: Request, res: Response): Promise<Response> {
     const { emailId } = req.params
     const message = await EmailMessageTransactional.findOne({
       where: { id: emailId, userId: req.session?.user?.id.toString() },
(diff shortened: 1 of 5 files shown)


backend/src/email/middlewares/email-transactional.middleware.ts
```diff
@@ -196,19 +196,16 @@ export const InitEmailTransactionalMiddleware = (
...
     if (!message) {
-      res
-        .status(404)
-        .json({ message: `Email message with ID ${emailId} not found.` })
-      return
+      throw new ApiNotFoundError(`Email message with ID ${emailId} not found.`)
     }
 
-    res.status(200).json(convertMessageModelToResponse(message))
+    return res.status(200).json(convertMessageModelToResponse(message))
```
tan-zx, line 208: no diff here either way right, just a stylistic choice

actually the return type is unnecessary also

tan-zx (approved):
looks good to me

backend/src/email/middlewares/email-transactional.middleware.ts
stanleynguyen: It's rather just a personal choice for establishing a convention for middlewares
- `Promise<Response>` for end-of-line middlewares
- `Promise<void>` for intermediary middlewares

PR #2136, opened 2023-07-31
feat: handle multiple updates in parseTemplateMessageWebhook and parseUserMessageWebhook

jia1:
Closes [SGC-161](https://linear.app/ogp/issue/SGC-161/must-we-handle-wa-webhook-request-body-messages[1]-onwards)

### DIFF
--- a/backend/src/govsg/services/govsg-callback.service.ts
+++ b/backend/src/govsg/services/govsg-callback.service.ts
@@ -18,6 +18,7 @@ import {
   WhatsAppWebhookButtonMessage,
   WhatsAppWebhookTextMessage,
   WhatsappWebhookMessageType,
+  HasTimestamp,
 } from '@shared/clients/whatsapp-client.class/types'
 import {
   MessageIdNotFoundWebhookError,
@@ -142,10 +143,10 @@ const storePrecreatedPasscode = async (
   } as GovsgVerification)
 }
 
-const timestampComparator: (
-  a: { timestamp: string },
-  b: { timestamp: string }
-) => number = (a, b) => {
+const timestampComparator: (a: HasTimestamp, b: HasTimestamp) => number = (
+  a,
+  b
+) => {
   const t1 = a.timestamp
   const t2 = b.timestamp
   if (t1 < t2) {
--- a/backend/src/govsg/services/govsg-callback.service.ts
+++ b/backend/src/govsg/services/govsg-callback.service.ts
@@ -4,6 +4,7 @@ import validator from 'validator'
 import { loggerWithLabel } from '@core/logger'
 import { Campaign } from '@core/models'
 import {
+  GenericMessage,
   UserMessageWebhook,
   WhatsAppApiClient,
   WhatsAppId,
@@ -406,10 +407,18 @@ const parseUserMessageWebhook = async (
   body: UserMessageWebhook,
   clientId: WhatsAppApiClient
 ): Promise<void> => {
-  const { wa_id: whatsappId } = body.contacts[0]
-  const { id: messageId, type } = body.messages[0]
+  await Promise.all(
+    body.messages.map((message) => parseGenericMessage(message, clientId))
+  )
+}
+
+const parseGenericMessage = async (
+  genericMessage: GenericMessage,
+  clientId: WhatsAppApiClient
+): Promise<void> => {
+  const { from: whatsappId, id: messageId, type } = genericMessage
   if (type === WhatsappWebhookMessageType.button) {
-    const message = body.messages[0] as WhatsAppWebhookButtonMessage
+    const message = genericMessage as WhatsAppWebhookButtonMessage
     if (message.button.text === 'Create passcode') {
       const passcodeCreationWamid = message.context.id
       const govsgVerification = await GovsgVerification.findOne({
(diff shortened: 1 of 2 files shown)


backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -245,7 +260,7 @@ const parseTemplateMessageWebhook = async (
         // not sure whether need to throw an error hmm probably not?
         return
       }
-      const { code, title, details, href } = body.errors[0]
+      const { code, title, details, href } = errors[0]
```
jia1, line 263: No known good way to handle `errors` ([docs](https://developers.facebook.com/docs/whatsapp/on-premises/webhooks/components#errors-object)).

shared/src/clients/whatsapp-client.class/types/webhook.ts
```diff
@@ -102,7 +102,7 @@ interface Contact {
   wa_id: WhatsAppId
 }
 
-interface Error {
+export interface Error {
```
tan-zx, line 105: strongly advise renaming these if you're going to export them. `Error` is a built-in type in JS

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -406,10 +421,18 @@ const parseUserMessageWebhook = async (
   body: UserMessageWebhook,
   clientId: WhatsAppApiClient
 ): Promise<void> => {
-  const { wa_id: whatsappId } = body.contacts[0]
-  const { id: messageId, type } = body.messages[0]
+  await Promise.all(
```
stanleynguyen, line 424: I'm wondering if the ordering of the items mean anything? Can we process them concurrently or must they be processed sequentially?

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -142,12 +145,24 @@ const storePrecreatedPasscode = async (
 const parseTemplateMessageWebhook = async (
   body: WhatsAppTemplateMessageWebhook,
   clientId: WhatsAppApiClient
+): Promise<void> => {
+  await Promise.all(
```
stanleynguyen, line 149: I think these are sequential events so if they come in bulk we should process them sequentially in the order that they come in. Or better yet, if there're event timestamps available we should follow those

jia1: I agree with you, Stanley. On it.

tan-zx: eh if I'm not wrong `body` always return a `statuses` array of length 1 (but no issue with handling it this way)

backend/src/govsg/services/govsg-callback.service.ts
tan-zx: I wrote the PR for this in a rush, but ideally we should use sth like `zod` at I/O boundaries, in this case to ensure that the shape of `errors` actually conform to the shape described in the doc. for future consideration.

backend/src/govsg/services/govsg-callback.service.ts
tan-zx: same as above, will always be array of length 1

backend/src/govsg/services/govsg-callback.service.ts
jia1: Noted on validating error shape.

This comment of mine was more of:

If there are 10 messages in `body.messages`, what about `body.errors`? None of the fields seem to help me map them to the individual message in `body.messages`.

backend/src/govsg/services/govsg-callback.service.ts
jia1: @tan-zx are you able to point me to the docs or comms from Meta that indicate this?

Why is their design so ambiguous!!!!

They have [a small section in their docs](https://developers.facebook.com/docs/graph-api/webhooks/getting-started#validate-payloads) which say batching may occur.

backend/src/govsg/services/govsg-callback.service.ts
jia1: Replied in the thread below!

backend/src/govsg/services/govsg-callback.service.ts
jia1: @stanleynguyen and yup done with sequential processing after sorting by timestamp.

tan-zx: > They have [a small section in their docs](https://developers.facebook.com/docs/graph-api/webhooks/getting-started#validate-payloads) which says batching may occur.

the link is definitely not for On Prem API (look at how they describe the configuration of webhooks)
image


> are you able to point me to the docs or comms from Meta that indicate this i.e. always array of length 1?
> Why is their design so ambiguous!!!!

it's documented somewhere on their site (I took 2 mins to try to find but couldn't, but I rmb it's there somewhere).

when I worked on it, I've documented [what their docs said vs what I actually received here](https://www.notion.so/opengov/WhatsApp-On-Prem-API-Webhook-f7bb4341658048ab8e525b0c32d3ea1a?pvs=4#f70cca05e7d74b4f847a6e1d72ea5617). as you can see, there are actually discrepancies between their docs and what their API does. I've given this feedback to the Meta team (you can find this in the support chat history).

backend/src/govsg/services/govsg-callback.service.ts
tan-zx: ohh my point was more general about mapping all I/O boundaries, not just error shape.

but ya to your point, `errors` is always length 1 also in practice.

backend/src/govsg/services/govsg-callback.service.ts
jia1: Thank you @tan-zx for sharing! I'll close this PR for now.

PR #2138, opened 2023-07-31
style: re-style messages table

jia1:
### Summary

Resolves non-functional changes in [SGC-168](https://linear.app/ogp/issue/SGC-168/additional-features-to-officer-verification)

### Before
Screenshot 2023-08-01 at 4 42 37 PM

### After
Screenshot 2023-08-01 at 4 38 45 PM


backend/src/govsg/middlewares/govsg-verification.middleware.ts
```diff
@@ -37,8 +37,8 @@ export const listMessages = async (
       name: recipient_name,
       mobile: recipient,
       data: JSON.stringify(remainingParams),
-      passcode: row.govsgVerification?.passcode ?? '****',
-      sent: row.sendAttemptedAt,
+      passcode: row.govsgVerification?.passcode ?? '',
```
KishenKumarrrrr, line 40: Curious, why did we decide to handle this in the FE instead?

frontend/src/components/common/StyledText/NegativeText.tsx
```diff
@@ -0,0 +1,11 @@
+import { ReactNode } from 'react'
+
+interface NegativeTextProps {
+  children?: ReactNode
+}
+
+const negativeTextStyle = { color: '#64707D' }
```
KishenKumarrrrr, line 7: Is the intention to make use of `$grey-700`? If so, would it be better to directly reference that variable?

frontend/src/components/common/StyledText/PlainBadge.tsx
```diff
@@ -0,0 +1,15 @@
+import { ReactNode } from 'react'
+
+interface PlainBadgeProps {
+  children?: ReactNode
+}
+
+const plainBadgeStyle = {
+  backgroundColor: '#F5F6FC',
```
KishenKumarrrrr, line 8: Similar to above

KishenKumarrrrr (commented):
Some small questions about the CSS, but other than that LGTM

backend/src/govsg/middlewares/govsg-verification.middleware.ts
jia1: Not sure if I am answering your qn:

Because I prefer to check `!value` than `value === 'Not created yet'`

Lmk if I'm not!

frontend/src/components/common/StyledText/NegativeText.tsx
jia1: I see. You saw that `grey-700` matches this colour? Thanks for the catch!

Let me change it tomorrow.

frontend/src/components/common/StyledText/PlainBadge.tsx
jia1: Got it!

frontend/src/components/common/StyledText/NegativeText.tsx
KishenKumarrrrr: Yup, if you search the hex code you should find that variable in `frontend/src/styles/base/_colours.scss`. 

But I think the intention is more important. Did you intend to make use of the app's colour scheme? 🤔

backend/src/govsg/middlewares/govsg-verification.middleware.ts
KishenKumarrrrr: My understanding of this code change is that previously we were resolving the display value entirely in this file (passcode = `****` if `row.govsgVerification?.passcode` is null) and then simply displaying the value of `passcode` on the FE. 

But now, we make it such that `passcode` could be an empty string and we check for this value in the FE instead (in `GovsgMessages.tsx`) before deciding what to display to the user.

jia1: Yup, your understanding is correct.

And now we want to show `Not created yet` instead of `****`. I rather do an empty string check.

Previously, I made the backend return `****` because I was simply lazy to do an empty string check. But now, it's unavoidable.

frontend/src/components/common/StyledText/NegativeText.tsx
jia1: My intention is to go for the easiest way out. But reusing `grey-700` works for me too. I've refactored accordingly.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
KishenKumarrrrr: Alright, I understand now. Thank you!

KishenKumarrrrr (approved):

PR #2143, opened 2023-08-03
feat: allow searching Govsg messages by recipient name or mobile number

jia1:
## Problem

Postman users need a way to search through Govsg messages by recipient name or recipient mobile number.

Closes [SGC-174](https://linear.app/ogp/issue/SGC-174/implement-search-by-recipient-name-or-mobile-number)

## Solution

- Allow the `listMessages` middleware that sits behind `/messages` (full path: `/campaign/${campaignId}/govsg/messages`) to accept a search query parameter.
- Run the search.
  - If the search term contains only 1 to 10 digits, search by mobile number. Else, search by recipient name.
  - I should explore if I can hand this over to the ORM through the use of an OR operation.

## Screenshots

Screenshot 2023-08-03 at 11 57 46 AM
Screenshot 2023-08-03 at 11 58 10 AM
Screenshot 2023-08-03 at 11 59 02 AM

## Other notes

We should pull `recipient_name` out into its own column in `govsg_messages` since we are searching by this field. I will checkout this branch and create the necessary migration there.

### DIFF
--- a/backend/src/govsg/middlewares/govsg-verification.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-verification.middleware.ts
@@ -10,7 +10,8 @@ const generateSearchOptions = (search: string) => {
   // TODO: Use an OR operation
   if (!search) {
     return {}
-  } else if (search.match(/^\d{1,10}$/)) {
+  }
+  if (search.match(/^\d{1,10}$/)) {
     return {
       recipient: {
         [Op.like]: `%${search}%`,
--- a/backend/src/govsg/middlewares/govsg-verification.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-verification.middleware.ts
@@ -4,16 +4,38 @@ import { GovsgMessage } from '@govsg/models'
 import { GovsgVerification } from '@govsg/models/govsg-verification'
 import { WhatsAppApiClient } from '@shared/clients/whatsapp-client.class/types'
 import { sendPasscodeCreationMessage } from '@govsg/services/govsg-verification-service'
+import { Op } from 'sequelize'
+
+const generateSearchOptions = (search: string) => {
+  if (!search) {
+    return {}
+  } else if (search.match(/\d{1,8}/)) {
+    return {
+      recipient: {
+        [Op.like]: `%${search}%`,
+      },
+    }
+  }
+  return {
+    params: {
+      recipient_name: {
+        [Op.like]: `%${search}%`,
+      },
+    },
+  }
+}
 
 export const listMessages = async (
   req: Request,
   res: Response
 ): Promise<Response | void> => {
   const { campaignId } = req.params
-  const { offset, limit } = req.query
+  const { offset, limit, search } = req.query
+  const searchOptions = generateSearchOptions(search as string)
   const { rows, count } = await GovsgMessage.findAndCountAll({
     where: {
       campaignId: +campaignId,
+      ...searchOptions,
     },
     offset: +(offset as string),
     limit: +(limit as string),
--- a/backend/src/govsg/middlewares/govsg-verification.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-verification.middleware.ts
@@ -19,7 +19,7 @@ const generateSearchOptions = (search: string) => {
   return {
     params: {
       recipient_name: {
-        [Op.like]: `%${search}%`,
+        [Op.iLike]: `%${search}%`,
       },
     },
   }
--- a/backend/src/govsg/middlewares/govsg-verification.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-verification.middleware.ts
@@ -7,9 +7,10 @@ import { sendPasscodeCreationMessage } from '@govsg/services/govsg-verification-
 import { Op } from 'sequelize'
 
 const generateSearchOptions = (search: string) => {
+  // TODO: Use an OR operation
   if (!search) {
     return {}
-  } else if (search.match(/\d{1,8}/)) {
+  } else if (search.match(/^\d{1,10}$/)) {
     return {
       recipient: {
         [Op.like]: `%${search}%`,
--- a/backend/src/govsg/routes/govsg-campaign.routes.ts
+++ b/backend/src/govsg/routes/govsg-campaign.routes.ts
@@ -1,4 +1,3 @@
-import { GovsgMessageStatus } from '@core/constants'
 import {
   CampaignMiddleware,
   JobMiddleware,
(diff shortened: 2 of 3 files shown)


jia1:
I forgot to debounce. Please let me do this in the next PR. Thank you.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
```diff
@@ -4,16 +4,39 @@ import { GovsgMessage } from '@govsg/models'
...
+  if (!search) {
+    return {}
+  } else if (search.match(/^\d{1,10}$/)) {
+    return {
+      recipient: {
+        [Op.like]: `%${search}%`,
+      },
+    }
+  }
+  return {
```
KishenKumarrrrr, line 20: nit: Should we complete the if-elseif-else structure?

backend/src/govsg/routes/govsg-campaign.routes.ts
```diff
@@ -133,7 +132,7 @@ router.get(
     [Segments.QUERY]: Joi.object({
       limit: Joi.number().integer().min(1).max(100).default(10),
       offset: Joi.number().integer().min(0).default(0),
-      status: Joi.string().valid(...Object.values(GovsgMessageStatus)),
+      search: Joi.string(),
```
KishenKumarrrrr, line 135: I have two questions regarding this:
1. Why did we previously validate `status` when it is not used in `listMessages`
2. Why do we not need it now?

frontend/src/components/dashboard/create/govsg/GovsgMessages.tsx
```diff
@@ -237,7 +254,18 @@ export const GovsgMessages = ({ campaignId }: GovsgMessagesProps) => {
         <TitleBar
           title={govsgMessageCount + ' Messages'}
           overrideStyles={overrideStylesTitleBar}
-        />
+        >
+          <TextInput
+            value={search}
+            type="text"
+            placeholder="Search for a message"
```
KishenKumarrrrr, line 261: Should we make this message more descriptive? Perhaps to inform the user that they can search by recipient name or phone number?

frontend/src/components/dashboard/create/govsg/GovsgMessages.tsx
```diff
@@ -85,6 +97,11 @@ export const GovsgMessages = ({ campaignId }: GovsgMessagesProps) => {
     void fetchGovsgMessages(index)
```
KishenKumarrrrr, line 97: Does this need to be changed? Consider the following scenario:
1. The user has performed a search for phone numbers starting with '9' and there are multiple pages returned
2. When the user navigates to the next page, this will cause the search params to be reset

We might need to ensure that all calls to `fetchGovsgMessages` contain the `search` variable.

KishenKumarrrrr (changes requested):
I have some questions regarding the use of `fetchGovsgMessages` and page changes in the FE.

backend/src/govsg/routes/govsg-campaign.routes.ts
jia1: (1) A copy mistake.
(2) Cus I realised...

frontend/src/components/dashboard/create/govsg/GovsgMessages.tsx
jia1: This is aligned with the Figma though...

Sure, I can change the description.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
```diff
@@ -4,16 +4,39 @@ import { GovsgMessage } from '@govsg/models'
 import { GovsgVerification } from '@govsg/models/govsg-verification'
 import { WhatsAppApiClient } from '@shared/clients/whatsapp-client.class/types'
 import { sendPasscodeCreationMessage } from '@govsg/services/govsg-verification-service'
+import { Op } from 'sequelize'
+
+const generateSearchOptions = (search: string) => {
```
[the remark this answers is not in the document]
jia1: Hmm, I'm currently not keen to take up on this because comments are code smell.

The fact that `generateSearchOptions` was extracted from the middleware into its own function serves as a self-documentation of what's going on.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
jia1: Thanks for catching the `else if`. I personally prefer `ifs` and early exit. Let me change the `else if` instead.

frontend/src/components/dashboard/create/govsg/GovsgMessages.tsx
jia1: I think so. I have not been successful with bulk campaigns so I can't verify this behaviour.

My qn to the team about "un-stuck-ing" campaigns wasn't being replied to, got to ping them again.

jia1: Reproduced locally.

Have done: `ensure that all calls to fetchGovsgMessages contain the search variable`.

jia1:
I'll resolve the threads which I've taken action on.

frontend/src/components/dashboard/create/govsg/GovsgMessages.tsx
KishenKumarrrrr: Oh... hmm maybe can just check with Yi Xin?

frontend/src/components/dashboard/create/govsg/GovsgMessages.tsx
```diff
@@ -77,12 +89,17 @@ export const GovsgMessages = ({ campaignId }: GovsgMessagesProps) => {
   }
 
   useEffect(() => {
-    void fetchGovsgMessages(selectedPage)
+    void fetchGovsgMessages(search, selectedPage)
```
KishenKumarrrrr, line 92: Sorry, I didn't catch this previously but can I ask why we need this `useEffect`?

When the user changes the page, the change is already handled in `handlePageChange` (which refetches data). Is there still a need to trigger `fetchGovsgMessages` again?

jia1: Without this, the first render will be empty (nobody clicked a page yet).

But, I have removed the `selectedPage` dependency now so that we don't run fetch twice per page change.

Thanks!

KishenKumarrrrr (approved):
Thanks for the hard work!

jia1:
Thanks Kishen for making time to spot bugs & useful improvements!

PR #2144, opened 2023-08-03
fix: TypeError: Cannot read properties of undefined (reading 'userId')

jia1:
https://open-government-products-re.sentry.io/issues/4354572595/?project=5272357&query=&referrer=project-issue-stream

### DIFF
--- a/backend/src/govsg/services/govsg-callback.service.ts
+++ b/backend/src/govsg/services/govsg-callback.service.ts
@@ -254,11 +254,9 @@ const parseTemplateMessageWebhook = async (
         return
       }
       const message = (govsgMessage ?? govsgOp) as GovsgMessage
-      const experimentalData = message.campaign
-        ? await experimentService.getUserExperimentalData(
-            message.campaign.userId
-          )
-        : {}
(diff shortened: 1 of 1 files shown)


backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -255,7 +255,7 @@ const parseTemplateMessageWebhook = async (
       }
       const message = (govsgMessage ?? govsgOp) as GovsgMessage
       const experimentalData = await experimentService.getUserExperimentalData(
-        message.campaign.userId
```
KishenKumarrrrr, line 258: Could I ask how you tested this?

KishenKumarrrrr (approved):
LGTM

backend/src/govsg/services/govsg-callback.service.ts
jia1: Sorry, I don't understand your qn...

If you mean how I found out, it's through Sentry.

If you mean how I ensured this works, I read the schema and got this to work.

PR #2145, opened 2023-08-03
refactor: ease search by recipient name

jia1:
## Note

Please review and merge #2143 first.

## Problem

In #2143, we search recipient name via the `params` JSONB.

Search is going to be a rather frequent operation and JSONB types aren't optimised for search. We foresee unnecessary load and potential bugs.

We should copy recipient name to a proper column of its own.

## Solution

- New migration: Add column, run mass update query
- Find all the spots where `GovsgMessage.*create` is called and ensure `recipientName` is defined.

## Deployment Checklist

- [ ] Run migration in staging
- [ ] Run migration in prod


KishenKumarrrrr (approved):
No issues with the changes, LGTM.

jia1:
I'll merge this on Monday.

PR #2146, opened 2023-08-03
refactor: make webhook server return 200 instead of 400

jia1:
This PR serves as a potential fix to our webhook taking up to several minutes to receive a `delivered` message status update.

Being too easy with returning HTTP 400 can lead to a snowball of retries from Meta's end, slowing down our own resources.

### DIFF
--- a/backend/src/govsg/services/govsg-callback.service.ts
+++ b/backend/src/govsg/services/govsg-callback.service.ts
@@ -65,7 +65,6 @@ const parseWebhook = async (
     return
   }
   if ('statuses' in body) {
-    // can delete this after we verified that it all works
     logger.info({
       message: 'Received status webhook from WhatsApp',
       action,
@@ -77,7 +76,6 @@ const parseWebhook = async (
     return
   }
   if ('messages' in body && 'contacts' in body) {
-    // can delete this after we verified that it all works
     logger.info({
       message: 'Received message webhook from WhatsApp',
       action,
--- a/backend/src/govsg/services/govsg-callback.service.ts
+++ b/backend/src/govsg/services/govsg-callback.service.ts
@@ -15,6 +15,7 @@ import {
   WhatsAppWebhookTextMessage,
   WhatsappWebhookMessageType,
 } from '@shared/clients/whatsapp-client.class/types'
+import { MessageIdNotFoundWebhookError } from '@shared/clients/whatsapp-client.class/errors'
 import {
   CampaignGovsgTemplate,
   GovsgMessage,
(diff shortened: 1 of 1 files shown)


backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -136,12 +132,14 @@ const parseTemplateMessageWebhook = async (
...
       },
     })
-    // throwing error here to return 400
+    // returning early
     // this is because callbacks could hit this endpoint before the messageId is updated in GovsgOp table
     // only do this for sending as this is unlikely to happen for other statuses
     // this will trigger a retry from WhatsApp, which will hit this endpoint again after messageId is updated
     if (whatsappStatus === WhatsAppMessageStatus.sent) {
-      throw new MessageIdNotFoundWebhookError('Message ID not found')
+      logger.error({
```
KishenKumarrrrr, line 140: > callbacks could hit this endpoint before the messageId is updated in GovsgOp table

In this scenario, it is acceptable to throw an error because we __need__ WhatsApp to retry sending the message right? Otherwise we would incorrectly resolve this webhook request -> update `messageId` in `GovsgOp` and then get stuck because the webhook request is never retried.

backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -58,15 +54,15 @@ const parseWebhook = async (
...
     logger.error({
       message: 'Unexpected webhook body',
       action,
       body,
     })
-    throw new UnexpectedWebhookError('Unexpected webhook body')
+    return
   }
   if ('statuses' in body) {
     // can delete this after we verified that it all works
```
KishenKumarrrrr, line 68: There are a few of these comments in the file, do we still need them?

KishenKumarrrrr (changes requested):
We might want to reconsider returning `200 OK` for the case I pointed out. Also, could you help to update the PR description to explain the need for this change please?

backend/src/govsg/services/govsg-callback.service.ts
jia1: Keep for traceability.

jia1: Removing such comments is out-of-scope of this PR but sure, I've made a commit to remove them.

backend/src/govsg/services/govsg-callback.service.ts
jia1: Nice catch, let me revert this.

jia1:
> We might want to reconsider returning `200 OK` for the case I pointed out. Also, could you help to update the PR description to explain the need for this change please?

Sorry, which case? I forgot.

KishenKumarrrrr:
> Sorry, which case? I forgot.

Sorry, I was referring to line 140 in `govsg-callback.service.ts` which you've already helped to address.

KishenKumarrrrr (approved):
Thanks for the changes.

PR #2147, opened 2023-08-03
refactor: add userClickedAt column to govsg_verification

KishenKumarrrrr:
## Problem

In this PR, we add a `user_clicked_at` column to the `govsg_verification` table. Doing so gives us an additional metric to track on our Grafana dashboard. 

## Solution

Each time our whatsapp webhook receives a button click event, it will update the `user_clicked_at` value for the correct entry in the `govsg_verification` table. 

## Deployment Checklist

- [ ] Run migrations

### DIFF
--- a/backend/src/govsg/services/govsg-callback.service.ts
+++ b/backend/src/govsg/services/govsg-callback.service.ts
@@ -338,11 +338,16 @@ const parseUserMessageWebhook = async (
   clientId: WhatsAppApiClient
 ): Promise<void> => {
   const { wa_id: whatsappId } = body.contacts[0]
-  const { id: messageId, type } = body.messages[0]
+  const { id: messageId, type, timestamp } = body.messages[0]
+  const timestampAsDate = new Date(parseInt(timestamp, 10) * 1000) // convert to milliseconds
   if (type === WhatsappWebhookMessageType.button) {
     const message = body.messages[0] as WhatsAppWebhookButtonMessage
     if (message.button.text === 'Create passcode') {
       const passcodeCreationWamid = message.context.id
+      await GovsgVerification.update(
+        { userClickedAt: timestampAsDate },
+        { where: { passcodeCreationWamid } }
+      )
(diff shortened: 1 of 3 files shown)


backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -338,11 +338,16 @@ const parseUserMessageWebhook = async (
...
 ): Promise<void> => {
   const { wa_id: whatsappId } = body.contacts[0]
-  const { id: messageId, type } = body.messages[0]
+  const { id: messageId, type, timestamp } = body.messages[0]
+  const timestampAsDate = new Date(parseInt(timestamp, 10) * 1000) // convert to milliseconds
   if (type === WhatsappWebhookMessageType.button) {
     const message = body.messages[0] as WhatsAppWebhookButtonMessage
     if (message.button.text === 'Create passcode') {
       const passcodeCreationWamid = message.context.id
+      await GovsgVerification.update(
```
jia1, line 347: Why not put this under `await GovsgVerification.findOne({` and make sure they are all done sequentially within 1 transaction?

jia1:
Thank you Kishen!

backend/src/govsg/services/govsg-callback.service.ts
KishenKumarrrrr: I tried to combine these into 1 query, but the problem I ran into was that `update` did not support `include: ...`.
As for why this is not a transaction, it is okay if the `update` is successful but the subsequent `findOne` fails because we are still correctly measuring the metric. (Since by this point the button click has already happened)

stanleynguyen: Yo guys, actually there's an update query that returns the object post-update also, u should use that instead of having 2 separate queries here

jia1: Got it, Kishen. Thank you for trying out my suggestion and letting me know what's up!!

jia1 (approved):
LGTM

PR #2101, opened 2023-07-13
feat: add cc functionality

KoalaTasha:
## Problem

Adding CC and BCC functionality for `/transactional/email/send `

## Solution

- store cc and bcc email addresses in a new table `email_messages_transactional_cc`
- update error_code for cc/bcc email addresses that have been blacklisted and exclude them from the cc/bcc list
- not storing the callback status of cc/bcc recipients
- ensure callback for `delivery`, `bounce` and `complaint` are correctly handled form main recipient (i.e callbacks meant for cc and bcc not processed for main recipient)
- `open` and `send` statuses cannot be differentiated for normal mail, cc and bcc recipients 

[RFC](https://www.notion.so/opengov/RFC-for-cc-functionality-on-Postman-7a99b01fea4a4a279aa897b984e9a539)
## Deployment Checklist

- run migration for new table in db

PR #2149, opened 2023-08-03
docs: update openapi.yaml

tan-zx:
should have been part of [this PR](https://github.com/opengovsg/postmangovsg/pull/2101)

### DIFF
--- a/backend/openapi.yaml
+++ b/backend/openapi.yaml
@@ -1256,18 +1256,23 @@ paths:
                   type: string
                 recipient:
                   type: string
+                cc:
+                  type: array
+                  items:
+                    type: string
+                bcc:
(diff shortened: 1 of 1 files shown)


backend/openapi.yaml
```diff
@@ -1256,18 +1256,23 @@ paths:
...
+                    type: string
+                bcc:
+                  type: array
+                  items:
+                    type: string
                 from:
                   type: string
                 reply_to:
                   type: string
-                  nullable: true
```
tan-zx, line 1263: these are just optional (which is the default in OpenAPI), not nullable. (same as the ones below)

stanleynguyen (approved):

PR #2150, opened 2023-08-04
chore: remove unused UnexpectedWebhookError and disable webhook retry for errors that don't need it

jia1:
### Summary of changes

- remove unused UnexpectedWebhookError
- make `isAuthenticated` middleware function in `GovsgCallbackMiddleware` return HTTP 200

### DIFF
--- a/backend/src/govsg/middlewares/govsg-callback.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-callback.middleware.ts
@@ -1,10 +1,7 @@
 import { loggerWithLabel } from '@core/logger'
 import { Request, Response, NextFunction } from 'express'
 import { GovsgCallbackService } from '@govsg/services/govsg-callback.service'
-import {
-  MessageIdNotFoundWebhookError,
-  UnexpectedWebhookError,
-} from '@shared/clients/whatsapp-client.class/errors'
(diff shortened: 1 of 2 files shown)


backend/src/govsg/middlewares/govsg-callback.middleware.ts
```diff
@@ -67,10 +64,6 @@ const parseWebhook = async (
     await GovsgCallbackService.parseWebhook(req.body, id)
     res.sendStatus(200)
   } catch (err) {
-    if (err instanceof UnexpectedWebhookError) {
-      res.sendStatus(500)
-      return
-    }
     if (err instanceof MessageIdNotFoundWebhookError) {
```
KishenKumarrrrr, line 67: Do you think we should discuss how we want our webhooks to behave? I'm not sure about the utility of returning non-200 statuses for cases such as invalid authentication or message id not being found if this will just trigger the same payload to get sent to us again. We can have a wider discussion with the other engineers on the team.

jia1: > I'm not sure about the utility of returning non-200 statuses for cases such as invalid authentication

Thanks for catching that I missed out the `isAuthenticated`. On it.

KishenKumarrrrr (approved):
Thanks for helping to clean up the webhook errors.

PR #2151, opened 2023-08-04
fix: retry delivered status updates without govsg message

jia1:
There is suspicion that "delivered" status updates without message ID are being received and subsequently dropped, leading to the absence of `sgc_passcode_generation`.

This PR deals with the above.

### DIFF
--- a/backend/src/govsg/services/govsg-callback.service.ts
+++ b/backend/src/govsg/services/govsg-callback.service.ts
@@ -101,10 +101,19 @@ const parseWebhook = async (
   return
 }
 
+const statusesWhichRequireMessageId = new Set([
+  WhatsAppMessageStatus.sent,
+  WhatsAppMessageStatus.delivered,
+])
+
 const parseTemplateMessageWebhook = async (
   body: WhatsAppTemplateMessageWebhook,
   clientId: WhatsAppApiClient
 ): Promise<void> => {
+  logger.info({
+    message: 'Logging status update(s) from WhatsApp...',
+    meta: body,
+  })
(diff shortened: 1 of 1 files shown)


backend/src/govsg/services/govsg-callback.service.ts
```diff
@@ -133,9 +142,9 @@ const parseTemplateMessageWebhook = async (
     })
     // throwing error here to return 400
     // this is because callbacks could hit this endpoint before the messageId is updated in GovsgOp table
-    // only do this for sending as this is unlikely to happen for other statuses
+    // only do this for sent and delivered as this is unlikely to happen for other statuses
     // this will trigger a retry from WhatsApp, which will hit this endpoint again after messageId is updated
-    if (whatsappStatus === WhatsAppMessageStatus.sent) {
+    if (statusesWhichRequireMessageId.has(whatsappStatus)) {
```
KishenKumarrrrr, line 147: Am I correct to say that this change is required because its possible that whatsapp directly sends a "delivered" status update? i.e, there's no guarantee that we will receive "sent" first before "delivered" for a particular messageId?

KishenKumarrrrr (approved):
No issues with the changes, LGTM.

backend/src/govsg/services/govsg-callback.service.ts
jia1: Yes. Your statements are correct. Here's [one of the docs](https://developers.facebook.com/docs/whatsapp/conversation-types#message-status) which support this.

PR #2152, opened 2023-08-06
feat: add whitelisting scripts

KishenKumarrrrr:
## Problem

In this PR, I add new whitelisting scripts to facilitate having different groups for WhatsApp templates. 

Closes [SGC-180](https://linear.app/ogp/issue/SGC-180/wa-templates-will-be-grouped-into-3-categories)

## Solution

Instead of handling the implementation for grouping templates at the DB level, we've decided that we can do it from the whitelisting scripts. Doing this gives us the added benefit of not having to change any of the query parameters/data model in the system. 

I've added 4 new whitelisting scripts. A detailed description of each script can be found in their respective files. 

1. `get_template_access.ts`
2. `get_all_templates.ts`
3. `update_whitelist.ts`
4. `whitelist_users.ts`

## Deployment Checklist

**New dev dependencies**:

- `@json2csv/plainjs` : used to create csv files

**No further deployment steps need to be taken**

### DIFF
--- a/backend/src/scripts/whitelisting/get_all_templates.ts
+++ b/backend/src/scripts/whitelisting/get_all_templates.ts
@@ -34,7 +34,7 @@ void (async function main() {
       templateId: t.id,
       templateLabel: t.whatsappTemplateLabel,
       templateName: t.name,
-      multilingualSupport: t.multilingualSupport,
+      multilingualSupport: t.multilingualSupport.map((mts) => mts.languageCode),
     }))
     console.log(csvData)
     const parser = new Parser(opts)
--- a/backend/src/scripts/whitelisting/get_all_templates.ts
+++ b/backend/src/scripts/whitelisting/get_all_templates.ts
@@ -1,0 +1,49 @@
+/* eslint-disable no-console */
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+require('dotenv').config()
+require('module-alias/register')
+
+import sequelizeLoader from '@core/loaders/sequelize.loader'
+import { GovsgTemplate } from '@govsg/models'
+import { Parser } from '@json2csv/plainjs'
+import { string as stringFormatter } from '@json2csv/formatters'
+import * as fs from 'fs'
+
+/**
+ * Returns a csv file of all templates in the database.
+ * The fields of the csv file returned are as follows:
+ * templateId, templateLabel, templateName, multilingualSupport
+ */
+void (async function main() {
+  try {
+    await sequelizeLoader()
+    const templates = await GovsgTemplate.findAll()
+
+    const opts = {
+      fields: [
+        'templateId',
+        'templateLabel',
+        'templateName',
+        'multilingualSupport',
+      ],
+      formatters: {
+        string: stringFormatter({ quote: '' }),
+      },
+    }
+    const csvData = templates.map((t) => ({
+      templateId: t.id,
+      templateLabel: t.whatsappTemplateLabel,
+      templateName: t.name,
+      multilingualSupport: t.multilingualSupport,
+    }))
+    console.log(csvData)
+    const parser = new Parser(opts)
+    const csv = parser.parse(csvData)
+    fs.writeFileSync('temp/output/templates.csv', csv)
+    console.log('Done ✅')
+  } catch (err) {
+    console.log(err)
+  }
+
+  process.exit(0)
+})()
--- a/backend/src/scripts/whitelisting/update_whitelist.ts
+++ b/backend/src/scripts/whitelisting/update_whitelist.ts
@@ -1,0 +1,113 @@
+/* eslint-disable no-console */
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+require('dotenv').config()
+require('module-alias/register')
+
+import sequelizeLoader from '@core/loaders/sequelize.loader'
+import { User } from '@core/models'
+import { Op } from 'sequelize'
+import { GovsgTemplate, GovsgTemplatesAccess } from '@govsg/models'
+import {
+  ALL_TEMPLATE_ACCESS,
+  MCI_TEMPLATE_ACCESS,
+  OGP_TEMPLATE_ACCESS,
+  SECTION_DIVIDER,
+} from './constants'
+
+/**
+ * Updates the user template access for the following default groups:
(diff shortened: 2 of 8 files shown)


backend/src/scripts/whitelisting/get_all_templates.ts
```diff
@@ -0,0 +1,49 @@
...
+  try {
+    await sequelizeLoader()
+    const templates = await GovsgTemplate.findAll()
+
+    const opts = {
+      fields: [
+        'templateId',
+        'templateLabel',
+        'templateName',
+        'multilingualSupport',
```
jia1, line 27: Have you tested this script with example values of `multilingualSupport`? Its type is `GovsgTemplateLanguageMetadata`.

backend/src/scripts/whitelisting/get_all_templates.ts
```diff
@@ -0,0 +1,49 @@
...
+    const csvData = templates.map((t) => ({
+      templateId: t.id,
+      templateLabel: t.whatsappTemplateLabel,
+      templateName: t.name,
+      multilingualSupport: t.multilingualSupport,
+    }))
+    console.log(csvData)
+    const parser = new Parser(opts)
+    const csv = parser.parse(csvData)
+    fs.writeFileSync('temp/output/templates.csv', csv)
```
jia1, line 42: Does this create a file in `backend/scripts/whitelisting/get_all_templates.ts` or project root? Are such files being git ignored?

backend/src/scripts/whitelisting/update_whitelist.ts
```diff
@@ -0,0 +1,113 @@
...
+      OGP_TEMPLATE_ACCESS.map(
+        (t) =>
+          ({
+            userId: u.id,
+            templateId: templateLabelToId.get(t),
+          } as unknown as GovsgTemplatesAccess)
+      )
+    ),
+    {
+      ignoreDuplicates: true,
```
jia1, line 62: Just confirming: This helps to prevent re-creation of records with the same primary key (`templateId` and `userId`).

backend/src/scripts/whitelisting/whitelist_users.ts
```diff
@@ -0,0 +1,246 @@
...
+  id?: number
+}
+
+/**
+ * Receives a csv and whitelists users to use the govsg feature.
+ * In addition to giving users default access for the different template groups (OGP, MCI, ALL),
+ * The user can be given adhoc access by passing in a list of template labels separated by ';' in the 5th column of the csv.
+ * Example: template1;template2;template3
+ */
+void (async function main() {
```
jia1, line 35: Just to confirm my understanding: This script should be used when we specify the 5th column in the whitelist csv.

As for `update_whitelist.ts`, it's typically used when:

- The 5th column is not specified and now we want to give the right access.
- For past users who never went through the `whitelist_users.ts`.
- And we don't want to update the csv whitelist in retrospect.

jia1 (approved):

backend/src/scripts/whitelisting/get_all_templates.ts
KishenKumarrrrr: I've not tested it against any example values yet. I tried to reference prod and staging db but the column has no fields. 

I wrongly assumed that the field's type was `string[]`. I think it would make sense for me to return an array of `GovsgTemplateLanguageMetadata.languageCode` in the csv?

backend/src/scripts/whitelisting/get_all_templates.ts
KishenKumarrrrr: Files will be created in `/backend/temp` that is already being git ignored.

backend/src/scripts/whitelisting/update_whitelist.ts
```diff
@@ -0,0 +1,113 @@
...
+  SECTION_DIVIDER,
+} from './constants'
+
+/**
+ * Updates the user template access for the following default groups:
+ * 1. OGP accessible templates
+ * 2. MCI accessible templates
+ * 3. Templates accessible to all users
+ */
+void (async function main() {
```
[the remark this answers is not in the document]
KishenKumarrrrr: My assessment was that using a transaction here isn't necessary. Reason for this is that the different writes are not strongly dependent on each other. (Its okay if we make a partial update)

Suppose the script fails halfway and we have only updated access to a subset of templates. This state is still acceptable and the script allows us to "patch" the results by just rerunning it.

backend/src/scripts/whitelisting/update_whitelist.ts
KishenKumarrrrr: Yup thats correct. This makes it easy to update the whitelist as we don't need to keep track of the values we have already added to the table.

backend/src/scripts/whitelisting/update_whitelist.ts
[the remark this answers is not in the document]
jia1: Got it. Thank you!

backend/src/scripts/whitelisting/get_all_templates.ts
jia1: Ok, that's great!

backend/src/scripts/whitelisting/whitelist_users.ts
KishenKumarrrrr: `whitelist_users.ts` is used for whitelisting new users. New users will automatically be given access to default templates (templates available to all/ogp officers/mci officers). The 5th column is used to give access to any additional non-default templates.

`update_whitelist.ts` on the other hand is used in the following scenarios:
1. To give user's added before we implemented template groups access to default templates
2. If we update the list of default templates a group has access to, this script helps to update the permissions in the db for all relevant users.

`update_whitelist.ts` does not take a csv file as input.

backend/src/scripts/whitelisting/get_all_templates.ts
jia1: Shucks, no examples around.

I suppose `multilingualSupport` is something to be updated by ourselves since template information is still hardcoded into db.

Yep, either `language*` works for me. But it's also more critical to have `languageCode` populated in the db.

Let's go with your suggestion.

jia1:
LGTM!

PR #2153, opened 2023-08-07
feat: Dedupe recipients for Gov.sg campaigns

jia1:
## Problem

If we allow each recipient in a campaign to receive multiple messages as a result of multi-lingual capabilities, there will be increased complexity in ensuring all such messages have the same OTP.

As such, we want to ensure each campaign contains unique recipients only. Each recipient can only receive 1 message in whichever language specified for the recipient.

Closes [SGC-181](https://linear.app/ogp/issue/SGC-181/de-dupe-recipient-column-in-postman-for-govsg-channel-only)

## Solution

- Update `GovsgService.uploadCompleteOnChunk` to additionally check for duplicate recipients and throw an error if so.
- Update the warning message in the frontend so that users are informed that `We do not allow duplicate recipients in Gov.sg campaigns`.

### Before
image

### After
Screenshot 2023-08-07 at 10 21 47 AM

### DIFF
--- a/backend/src/govsg/services/govsg.service.ts
+++ b/backend/src/govsg/services/govsg.service.ts
@@ -115,6 +115,7 @@ export function uploadCompleteOnChunk({
   campaignId: number
 }): (data: CSVParams[]) => Promise<void> {
   return async function (data: CSVParams[]): Promise<void> {
+    const recipients = new Set<string>([])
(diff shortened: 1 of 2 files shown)


KishenKumarrrrr (approved):
LGTM 👍

PR #2154, opened 2023-08-07
fix: remove initial flicker when showing Gov.sg messages

jia1:
## Problem

Before this PR, there would be an initial flicker from `renderNoMatch()` -> `renderGovsgMessage()` because we had no loading spinner for this component.

Now, we have it.

Closes [SGC-184](https://linear.app/ogp/issue/SGC-184/fix-flicker-between-empty-dashboard-and-list-of-messages)

Note: For now, it is ok to rapidly switch from `renderNoMatch()` -> `renderGovsgMessage()` during search. This is because we don't want the search bar to become a `LoadingSpinner` on key down.


KishenKumarrrrr (approved):
LGTM 👍

PR #2155, opened 2023-08-07
feat: enable multi-lingual templates for single-send

jia1:
## Problem

Right now, we only send messages in English. We want to allow officers to send messages in other languages (i.e. Chinese, Malay, Tamil).

This PR affects single-send only. 

Closes [SGC-182](https://linear.app/ogp/issue/SGC-182/[wa-templates]-multi-language-support-p2)

## Solution

- Note: Only users with access to the GOVSGV feature can enjoy this feature.
- When picking a message template, show language previews if the template has a non-empty `multilingual_support` column.
  - Now, we show either 0 or all 4 language chips. This [task](https://linear.app/ogp/issue/SGC-188/show-only-language-chips-which-the-template-has) will get this cleaned up.
- When specifying template arguments, allow users to select 1 out of 4 languages. English is selected by default.
  - On selection change, re-render the message preview below.
- Propagate language selection from frontend to backend and backend to worker.
- Remove hardcoded language values from worker.
- Note: After a campaign is sent, the template shown at the bottom is still in English only. Not sure whether/how to fix that yet ([Linear](https://linear.app/ogp/issue/SGC-190/after-campaign-is-sent-the-template-is-still-always-in-english)), but it'll be out-of-scope of this PR.

## Screenshots

Screenshot 2023-08-10 at 11 31 02 AM
Screenshot 2023-08-10 at 11 31 15 AM
Screenshot 2023-08-10 at 11 31 31 AM
Screenshot 2023-08-10 at 11 31 51 AM
Screenshot 2023-08-10 at 11 32 08 AM
Screenshot 2023-08-10 at 11 32 18 AM

## Other notes

- Upgraded `jest-canvas-mock` so that [this problem](https://stackoverflow.com/questions/75414699/jest-typeerror-cannot-redefine-property-window) does not surface when running frontend integration tests locally.


backend/src/database/migrations/20230807072256-make-get-messages-to-send-govsg-function-return-language-code.js
```diff
@@ -0,0 +1,105 @@
...
+          SELECT id FROM govsg_ops WHERE
+          campaign_id in (SELECT * FROM campaignId)
+          AND send_attempted_at IS NULL
+          LIMIT lim
+          FOR UPDATE SKIP LOCKED
+        ),
+        messages AS (
+          UPDATE govsg_ops SET send_attempted_at=clock_timestamp(), updated_at = clock_timestamp() WHERE
+          id in (SELECT * from selected_ids)
+          RETURNING id, recipient, params, campaign_id, language_code
```
jia1, line 30: Additionally return `language_code` so that we can pass this into `sendMessage` eventually.

backend/src/database/migrations/20230807072256-make-get-messages-to-send-govsg-function-return-language-code.js
```diff
@@ -0,0 +1,105 @@
...
+        ),
+        templates AS (
+          SELECT govsg_templates.whatsapp_template_label as whatsapp_template_label, campaign_govsg_template.campaign_id  as campaign_id, govsg_templates.params as param_order
+          from govsg_templates JOIN campaign_govsg_template ON govsg_templates.id = campaign_govsg_template.govsg_template_id 
+          where govsg_templates.id in (SELECT govsg_template_id from campaign_govsg_template where campaign_id in (SELECT * FROM campaignId))
+        )
+        SELECT json_build_object('id', m.id, 'recipient', m.recipient, 'params', m.params, 'paramOrder', t.param_order, 'whatsappTemplateLabel', t.whatsapp_template_label, 'campaignId', m.campaign_id, 'language', m.language_code)
```
jia1, line 37: Build json with `language` field since this is how the language code is referred to in the worker..

backend/src/database/migrations/20230807072256-make-get-messages-to-send-govsg-function-return-language-code.js
```diff
@@ -0,0 +1,105 @@
...
+        IF NOT FOUND THEN
+          UPDATE job_queue SET status = 'SENT', updated_at = clock_timestamp() where id = jid AND status = 'SENDING';
+        END IF;
+      `,
+      [],
+      { force: true }
+    )
+  },
+
+  down: async (queryInterface, _) => {
```
jia1, line 55: Reverting this migration should mean rolling back the function instead of dropping it.

backend/src/govsg/middlewares/govsg-template.middleware.ts
```diff
@@ -16,20 +16,37 @@ import { CampaignService, StatsService, UploadService } from '@core/services'
...
 
+const convertMultilingualSupportToResponse = (
+  multilingualSupport: Array<GovsgTemplateLanguageMetadata>
+) => {
+  return multilingualSupport.map((languageSupport) => ({
+    body: languageSupport.body,
+    language: languageSupport.language,
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    code: languageSupport.languageCode ?? languageSupport['language_code'],
```
jia1, line 36: I was expecting camel case keys but I got snake case keys here. I don't intend to investigate this further unless time permits.

backend/src/govsg/middlewares/govsg-template.middleware.ts
```diff
@@ -16,20 +16,37 @@ import { CampaignService, StatsService, UploadService } from '@core/services'
...
+
 function convertGovsgTemplateToResponse(template: GovsgTemplate) {
   return {
     id: template.id,
     body: template.body,
     params: template.params,
     param_metadata: template.paramMetadata,
     name: template.name,
-    multilingual_support: template.multilingualSupport,
+    languages: convertMultilingualSupportToResponse(
```
jia1, line 47: Too much confusion between camel case and snake case on fields and nested fields. Troubleshooting this was extremely painful. Chose to avoid this problem altogether.

backend/src/govsg/models/govsg-template.ts
```diff
@@ -71,5 +71,5 @@ export class GovsgTemplate extends Model<GovsgTemplate> {
     allowNull: true,
     defaultValue: [],
   })
-  multilingualSupport: GovsgTemplateLanguageMetadata[]
+  multilingualSupport: Array<GovsgTemplateLanguageMetadata>
```
jia1, line 74: I prefer using [] but I got to be consistent with the rest of the codebase as much as I see.

backend/src/govsg/services/govsg.service.ts
```diff
@@ -22,7 +22,13 @@ export async function getCampaignDetails(
       where: { campaignId },
       include: {
         model: GovsgTemplate,
-        attributes: ['id', 'body', 'params', 'param_metadata'],
+        attributes: [
+          'id',
+          'body',
+          'params',
+          'param_metadata',
+          'multilingualSupport',
```
jia1, line 30: A bit disturbing that I have it camel case here. I'll check.

backend/src/govsg/services/govsg.service.ts
```diff
@@ -32,6 +38,7 @@ export async function getCampaignDetails(
       ? {
           ...(pivot.govsgTemplate.toJSON() as any), // any to get around the snake vs camel casing difference between TS type and actual db table field
           for_single_recipient: pivot.forSingleRecipient,
+          multilingual_support: pivot.govsgTemplate.multilingualSupport,
```
jia1, line 41: TODO: Tidy this up

backend/src/core/interfaces/campaign.interface.ts
```diff
@@ -44,6 +47,7 @@ export interface CampaignDetails {
     params: Array<string> | null
     for_single_recipient: boolean
     param_metadata: Record<string, GovsgTemplateParamMetadata>
+    multilingual_support: GovsgTemplateLanguageMetadata[]
```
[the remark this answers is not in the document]
jia1: Fixed.

backend/src/govsg/services/govsg.service.ts
jia1: `multilingualSupport` is correct. `param_metadata` came from `master` and I don't intend to risk breaking anything by making a commit on this.

backend/src/govsg/services/govsg.service.ts
jia1: Fixed.

worker/src/core/loaders/message-worker/govsg.class.ts
```diff
@@ -72,6 +72,7 @@ class Govsg {
         body: string
         campaignId: number
         whatsappTemplateLabel: string
+        language: string
```
jia1, line 75: Note: This is actually language code but the downstream function looks for `language` field. I don't intend to refactor anything.

frontend/src/components/dashboard/create/govsg/GovsgPickTemplate.tsx
```diff
@@ -10,6 +11,8 @@ import {
 
 import { useParams } from 'react-router-dom'
 
+import { getLocalisedTemplateBody } from '../../../../utils/templateLocalisation'
```
jia1, line 14: Adding `utils/*` to the `tsconfig.paths.json` file worked locally but not when running the integration tests (see previously failed pipelines).

frontend/package.json
```diff
@@ -100,7 +100,7 @@
     "eslint-plugin-import": "2.25.4",
     "eslint-plugin-prettier": "4.0.0",
     "eslint-plugin-react": "7.28.0",
-    "jest-canvas-mock": "2.3.1",
+    "jest-canvas-mock": "2.5.2",
```
jia1, line 103: Need to upgrade this or I'll encounter [this problem](https://stackoverflow.com/questions/75414699/jest-typeerror-cannot-redefine-property-window) when running frontend tests locally.

frontend/src/classes/GovsgCampaign.ts
```diff
@@ -36,6 +40,7 @@ export class GovsgCampaign extends Campaign {
     this.numRecipients = input['num_recipients']
     this.forSingleRecipient = input['govsg_templates']?.for_single_recipient
     this.paramMetadata = input['govsg_templates']?.param_metadata
+    this.languages = input['govsg_templates']?.languages ?? []
```
KishenKumarrrrr, line 43: Am I correct to say that don't have to provide a default language because we already set a default value when the API is called?

frontend/src/components/common/ChipGroup/LanguageChipGroup.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const LanguageChipGroup = ({
+  selected,
+  setSelection,
+}: LanguageChipGroupProps) => {
+  return (
+    <div style={languageChipGroupStyle}>
+      {Object.keys(WhatsAppLanguages).map((language) => {
+        const value =
+          WhatsAppLanguages[language as keyof typeof WhatsAppLanguages]
```
KishenKumarrrrr, line 25: If we use `Object.entries()`, would we be able to avoid this access pattern?

frontend/src/components/dashboard/create/govsg/GovsgSingleRecipient.tsx
```diff
@@ -101,6 +132,34 @@ const GovsgSingleRecipient = ({
...
         </div>
+        {canAccessGovsgV && (
+          <div>
+            <p>
+              <label htmlFor="language">Language</label>{' '}
+            </p>
+            {Object.keys(WhatsAppLanguages).map((language) => {
+              const label = capitalize(language)
+              const languageCode =
+                WhatsAppLanguages[language as keyof typeof WhatsAppLanguages]
```
KishenKumarrrrr, line 143: Similar comment to above.

frontend/src/components/dashboard/create/govsg/RadioChoice.tsx
```diff
@@ -26,7 +26,10 @@ function RadioChoice({
         <label htmlFor={id}>{label}</label>
       </div>
       {rest.checked && children && (
```
KishenKumarrrrr, line 28: If `children = []`, will there be an error here? Since `[]` is truthy but we can't do `children[0]`

frontend/src/components/dashboard/create/govsg/RadioChoice.tsx
```diff
@@ -26,7 +26,10 @@ function RadioChoice({
...
-        <div className={styles.body}>{children}</div>
+        <div className={styles.body}>{children[0]}</div>
+      )}
+      {rest.checked && children && children.length > 1 && (
+        <div className={styles.footer}>{children.slice(1)}</div>
```
KishenKumarrrrr, line 32: Sorry but could you help explain both conditions, I dont really understand? (Lines 28 and 31)

KishenKumarrrrr (approved):
Good job with the implementation, its a difficult feature to implement.
I have some small questions but otherwise LGTM.

backend/src/govsg/middlewares/govsg-template.middleware.ts
jia1: Back when I did this PR, I thought I was hallucinating about the cases so I added `??` as a "safety" net.

Now I know `languageSupport.languageCode` has always been `null`, I made the correction to just `languageSupport['language_code']` in one of the smaller subsequent PRs.

frontend/src/classes/GovsgCampaign.ts
jia1: Which API?

But yes, there is a default value for `language_code` in `GovsgMessage`. As a result, I also specified a default value in the frontend (there was this line that says `fieldsToRender ...`).

frontend/src/components/dashboard/create/govsg/GovsgSingleRecipient.tsx
jia1: Have replied above too!

frontend/src/components/dashboard/create/govsg/RadioChoice.tsx
jia1: Thank you for the catch, Kishen.

I'll add a length check in a subsequent PR. Hope that's ok!

frontend/src/components/dashboard/create/govsg/RadioChoice.tsx
jia1: My goal is to have the first child be part of body and for the rest to be part of footer.

iirc when I checked the usages of this component, there is either 1 or 2 children.

frontend/src/classes/GovsgCampaign.ts
KishenKumarrrrr: Im referring to the `/send-single` route. I saw that a default value is added in the input checks backend/src/govsg/routes/govsg-campaign.routes.ts

jia1: I see. Yes, you are correct.

And we also have this in the frontend:

```javascript
  const [data, setData] = useState<Record<string, string>>(
    fieldsToRender.reduce((cul, f) => ({ [f.id]: '', ...cul }), {
      recipient: '',
      languageCode: WhatsAppLanguages.english,
    })
  )
```

frontend/src/components/dashboard/create/govsg/RadioChoice.tsx
jia1: I'll resolve this first.

jia1:
> Good job with the implementation, its a difficult feature to implement. I have some small questions but otherwise LGTM.

Thank you, Kishen!

frontend/src/components/common/ChipGroup/LanguageChipGroup.tsx
jia1: Update: This became a non-issue due to #2161

frontend/src/components/dashboard/create/govsg/RadioChoice.tsx
jia1: Added to #2162

PR #2156, opened 2023-08-08
Revert "fix: TypeError: Cannot read properties of undefined (reading 'userId')"

jia1:
Reverts opengovsg/postmangovsg#2144

The "delivered" status update webhook requests have been "missing" since last Thursday/Friday, leading to absence of officer verification message(s).

This commit was found to have caused the issue. It is currently unknown why but I propose to revert this first.

Impact of reverting: We'll see Sentry errors on this type error, but it's acceptable for now.

### DIFF
--- a/backend/src/govsg/services/govsg-callback.service.ts
+++ b/backend/src/govsg/services/govsg-callback.service.ts
@@ -259,7 +259,7 @@ const parseTemplateMessageWebhook = async (
       }
       const message = (govsgMessage ?? govsgOp) as GovsgMessage
       const experimentalData = await experimentService.getUserExperimentalData(
-        message.campaign?.userId ?? -1
+        message.campaign.userId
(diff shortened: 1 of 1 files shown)


KishenKumarrrrr:
> This commit was found to have caused the issue. It is currently unknown why but I propose to revert this first.

Should we be reverting if we don't know why this causes the problem yet? Or are you confident that this is really the issue

jia1:
> > This commit was found to have caused the issue. It is currently unknown why but I propose to revert this first.
> 
> Should we be reverting if we don't know why this causes the problem yet? Or are you confident that this is really the issue

I have just verified that this is the only commit in the entire history causing the problem :D

KishenKumarrrrr (approved):
KIV: We'll need to figure out why this causes the issue.

PR #2158, opened 2023-08-10
fix: fetch correct client

jia1:
## Problem

The Resend button uses client 2 all the time.

Closes [SGC-191](https://linear.app/ogp/issue/SGC-191/resend-causes-two-govsg-threads-in-prod)

## Solution

Remove whitespaces and prepend +countrycode to phone numbers when using the `getApiClientId` query.

Don't merge today. Merge on Friday.

### DIFF
--- a/backend/src/govsg/middlewares/govsg-verification.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-verification.middleware.ts
@@ -101,6 +101,8 @@ export const resendPasscodeCreationMessage = async (
   }
   const { recipient } = govsgMessage
   // flamingoDb has phone numbers prepended with +countrycode but phone numbers in govsgMessage don't always have that.
+  // In addition, phone numbers in govsgMessage are sometimes observed to have whitespaces inside them. Such is not observed
+  // in flamingoDb.
   const recipientWithCountryCode =
     removeWhitespacesAndPrependCountryCode(recipient)
   const apiClientIdMap = await whatsappService.flamingoDbClient.getApiClientId([
--- a/backend/src/govsg/middlewares/govsg-verification.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-verification.middleware.ts
@@ -4,8 +4,8 @@ import { GovsgMessage } from '@govsg/models'
 import { GovsgVerification } from '@govsg/models/govsg-verification'
 import { WhatsAppApiClient } from '@shared/clients/whatsapp-client.class/types'
 import { sendPasscodeCreationMessage } from '@govsg/services/govsg-verification-service'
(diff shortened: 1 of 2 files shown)


backend/src/govsg/middlewares/govsg-verification.middleware.ts
```diff
@@ -99,8 +100,13 @@ export const resendPasscodeCreationMessage = async (
     })
   }
   const { recipient } = govsgMessage
+  // flamingoDb has phone numbers prepended with +countrycode but phone numbers in govsgMessage don't always have that.
+  // In addition, phone numbers in govsgMessage are sometimes observed to have whitespaces inside them. Such is not observed
+  // in flamingoDb.
```
ihwk1996, line 105: Is this comment clearer?

```suggestion
  // flamingoDb numbers are prepended with +countrycode and have no whitespaces.
  // recipient numbers of govsgMessage may not fulfil the above thus have to be preprocessed first.
```

ihwk1996 (approved):

backend/src/govsg/middlewares/govsg-verification.middleware.ts
jia1: Let me take it!

Thank you Ivan!

jia1:
I heard from Stanley he does do normalisation so I found the existing normalisation function and will be using that.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
```diff
@@ -99,8 +100,11 @@ export const resendPasscodeCreationMessage = async (
     })
   }
   const { recipient } = govsgMessage
+  // flamingoDb numbers are prepended with +countrycode and have no whitespaces.
+  // recipient numbers of govsgMessage may not fulfil the above thus have to be preprocessed first.
+  const normalisedRecipient = PhoneNumberService.normalisePhoneNumber(recipient)
```
KishenKumarrrrr, line 105: Do we need to add this normalisation in the other instances where the `getApiClient` function is used (Looks like `govsg.class.ts` and `govsg-transactional.service.ts`)? I assume that this normalisation is always required?

jia1: Those are created by the original coreteam and they are aware of the need for normalisation. If you do a quick search of `normalisePhoneNumber`, it should be already called in those places.

I never encountered this because I was operating from the webhook-side where the client ID is already given by WhatsApp.

KishenKumarrrrr: I agree, best not to touch the flamingo function. Thanks for your reply.

KishenKumarrrrr (approved):

PR #2159, opened 2023-08-10
style: set cursor to pointer when hovering over tabs in CampaignHeaderTabs

jia1:
## Problem

When hovering over tabs in CampaignHeaderTabs, the cursor was not set, so it becomes unclear if the tabs are clickable.

Closes [SGC-187](https://linear.app/ogp/issue/SGC-187/use-cursor-pointer-for-report-messages-tabs)

## Solution

CSS


KishenKumarrrrr (approved):
LGTM

PR #2160, opened 2023-08-10
feat: enable multi-lingual support for bulk send

jia1:
## Notes

- Merge #2155 first. Remember to switch base to `master`.
- Only users with access to the GOVSGV feature can enjoy this feature.

## Summary of changes

Enable multi-lingual templates for bulk send too.

Closes [SGC-189](https://linear.app/ogp/issue/SGC-189/support-multi-lingual-messages-in-bulk-send)

## Checklist

- [x] Add `language` column to sample CSV
- [x] Inject language into `GovsgMessage`
- [x] Localise message preview
- [x] Transform language into language code for upstream processing

## Screenshots

Screenshot 2023-08-10 at 4 54 59 PM

### DIFF
--- a/backend/src/govsg/middlewares/govsg-template.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-template.middleware.ts
@@ -33,7 +33,7 @@ const convertMultilingualSupportToResponse = (
     language: languageSupport.language,
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
-    code: languageSupport.languageCode ?? languageSupport['language_code'],
+    code: languageSupport.language_code,
   }))
 }
 
--- a/backend/src/govsg/services/govsg-template.service.ts
+++ b/backend/src/govsg/services/govsg-template.service.ts
@@ -29,12 +29,17 @@ export async function getFilledTemplate(
 
 const getLocalisedTemplateBody = (
   template: GovsgTemplate | null,
-  languageCode: string
+  language: string
 ) => {
+  console.log(
+    'getLocalisedTemplateBody',
+    language,
+    template?.multilingualSupport
+  )
+  const languageInLowerCase = language.toLowerCase()
   return (template?.multilingualSupport.find(
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
-    (languageSupport) => languageSupport.language_code === languageCode
+    (languageSupport) =>
+      languageSupport.language.toLowerCase() === languageInLowerCase
   )?.body ?? template?.body) as string
 }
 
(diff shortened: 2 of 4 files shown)


backend/src/govsg/middlewares/govsg-template.middleware.ts
```diff
@@ -33,7 +33,7 @@ const convertMultilingualSupportToResponse = (
     language: languageSupport.language,
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
-    code: languageSupport.languageCode ?? languageSupport['language_code'],
+    code: languageSupport.language_code,
```
jia1, line 36: Sequelize does not seem to care about nested fields, so they remain in snake case. Thus removing the redundant `xxx ??`

backend/src/govsg/services/govsg-template.service.ts
```diff
@@ -27,6 +27,22 @@ export async function getFilledTemplate(
   return pivot.govsgTemplate
 }
 
+const getLocalisedTemplateBody = (
```
jia1, line 30: Will only think of moving this to a separate file if more functions are using this. Right now, only 1 usage.

backend/src/govsg/services/govsg.service.ts
```diff
@@ -114,6 +115,18 @@ export function uploadCompleteOnPreview({
   }
 }
 
+const getLanguageCode = (language: string) => {
+  const languageInLowerCase = language.toLowerCase()
+  const whatsAppLanguages = Object.keys(WhatsAppLanguages)
+  const key = whatsAppLanguages.find(
+    (whatsAppLanguage) => whatsAppLanguage.toLowerCase() === languageInLowerCase
```
jia1, line 122: If we look at the current situation, `toLowerCase()` is redundant here but I want to be resilient against careless refactors.

frontend/src/components/dashboard/create/govsg/GovsgRecipients.tsx
```diff
@@ -184,7 +184,7 @@ const GovsgRecipients = ({
           <FileInput isProcessing={isUploading} onFileSelected={uploadFile} />
           <p>or</p>
           <SampleCsv
-            params={params}
+            params={['language', ...params]}
```
jia1, line 187: I didn't want to put `language` as a fixed header inside `SampleCsv` because `SampleCsv` is used beyond Gov.sg.

KishenKumarrrrr (approved):
You fixed some of my comments from #2155 here, can just ignore those.

backend/src/govsg/services/govsg.service.ts
```diff
@@ -114,6 +115,21 @@ export function uploadCompleteOnPreview({
   }
 }
 
+const getLanguageCode = (language: string | undefined) => {
+  if (!language) {
+    return WhatsAppLanguages.english
+  }
+  const languageInLowerCase = language.toLowerCase()
```
KishenKumarrrrr, line 122: minor nit: but actually I think this is small enough to just be done inline on line 125

jia1: I just want to be a bit clearer than making this function call feel like an afterthought. Would probably leave this in until it becomes a pain.

PR #2161, opened 2023-08-10
feat: smart lang chips

jia1:
## Problem

Before this PR, we either show 0 or 4 language chips and radio options.

We should show them based on the actual languages supported by the template.

Closes [SGC-188](https://linear.app/ogp/issue/SGC-188/show-only-language-chips-which-the-template-has)


KishenKumarrrrr (approved):
LGTM

PR #2162, opened 2023-08-11
fix: bug fixes for multi-lingual feature

jia1:
### refactor: remove redundant length check

This length check was originally needed because we show either 0 or 4 lang chips. Now, we check the `languages` array to know what to render.

### fix: update capitalisation due to smart lang chips

The values passed around have changed a bit due to this refactor. This commit handles the aftermath that was previously undetected.

### fix: make getCampaignDetails return code instead of language_code

With the current set of PRs, coming back to an existing Draft campaign and selecting the language will lead to a crash because I did not make `getCampaignDetails` return `code` instead of `language_code`.

### fix: add missing length check

`rest.checked && children` alone is not enough to say there is at least 1 child in the array. This is because `[]` is truthy. Added a length check. Kudos to @KishenKumarrrrr for catching this.


KishenKumarrrrr (approved):
LGTM

PR #2163, opened 2023-08-11
chore: make import-in-the-middle ^1.4.2

jia1:
https://security.snyk.io/vuln/SNYK-JS-IMPORTINTHEMIDDLE-5826054

### Notes

- `import-in-the-middle` is a dep of `dd-trace`
- `dd-trace` for the other `package.json` are already updated but not this. So I'm updating it in this PR.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -44,7 +44,7 @@
     "connect-redis": "4.0.4",
     "convict": "6.2.4",
     "cors": "2.8.5",
-    "dd-trace": "2.13.0",
+    "dd-trace": "2.45.1",
     "dotenv": "8.2.0",
     "escape-html": "1.0.3",
     "express": "4.18.2",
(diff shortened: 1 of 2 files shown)


backend/package.json
```diff
@@ -44,7 +44,7 @@
     "connect-redis": "4.0.4",
     "convict": "6.2.4",
     "cors": "2.8.5",
-    "dd-trace": "2.13.0",
```
KishenKumarrrrr, line 47: `import-in-the-middle` is not directly upgraded?

Also, how do we test if the upgrade works? Pass CI == OK?

jia1: Not directly upgraded because it only exists in `package-lock.json` as a dep under `dd-trace`.

I'm taking a leap of faith here because the upgrade was a minor version upgrade and it was done for the other folders but not this.

PR #2164, opened 2023-08-11
Revert "fix: bug fixes for multi-lingual feature"

jia1:
Reverts opengovsg/postmangovsg#2162


jia1:
Not needed anymore. The reason things did not work as expected in production was because I forgot to run pending db migrations.

PR #2165, opened 2023-08-11
fix: make message campaign defined

jia1:


KishenKumarrrrr (approved):
Thanks for making the changes.

PR #2169, opened 2023-08-14
refactor: rename and hide language param

jia1:
## Problem

Two problems:

### Inconsistency in the param field name for language code

In bulk send, the param field name for language code is `language`. Refer to the sample csv file. But in single send, the param field name for language code is `languageCode`.

Not ideal because the govsg messages frontend table will show `language` for bulk send and `languageCode` for single send. Plus, in the db, `params` has both snake case (e.g. `officer_name` ) and camel case (e.g. `languageCode`) field names.

This PR standardises the param field name for language code to just `language`.

### Should hide the language from params in the govsg messages table 

Why:

- Leaving the code as-is would give a poor UX. I put in a small amount of time to see if I could expand the height of the table row but I have not been successful yet.
    - Poor UX: Table row is too squeezy for showing the language, see screenshot
- Original screen does not have this. We may want to render this in a column of its own. But given there are other more important tasks at hand, we hide this param first.
    - Poor UX: For single send, the language value is pre-populated by the radio button but for bulk, the language value is the raw value provided by the user.

#### Single send
Screenshot 2023-08-14 at 9 54 10 AM

#### Bulk send
Screenshot 2023-08-14 at 9 54 30 AM

#### Both
Screenshot 2023-08-14 at 9 54 44 AM

### DIFF
--- a/backend/src/govsg/middlewares/govsg-verification.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-verification.middleware.ts
@@ -58,6 +58,8 @@ export const listMessages = async (
       recipient_name,
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
       officer_designation,
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      language,
       ...remainingParams
     } = row.params as Record<string, string>
     return {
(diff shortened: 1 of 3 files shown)


KishenKumarrrrr (commented):
No issues with the changes made to rename `languageCode` to `language`. However could I check how we are performing the hiding of the language information in the frontend? It is not very clear to me based on the changes made as they only cover the renaming of the field.

jia1:
> No issues with the changes made to rename `languageCode` to `language`. However could I check how we are performing the hiding of the language information in the frontend? It is not very clear to me based on the changes made as they only cover the renaming of the field.

If you look at the 2nd commit, I separated `language` from `params` through de-structuring. Then, I return params with `remainingParams` as the value.

KishenKumarrrrr (approved):
> If you look at the 2nd commit, I separated language from params through de-structuring. Then, I return params with remainingParams as the value.

Brilliant stuff, sorry I missed it.

PR #2172, opened 2023-08-14
feat: allow user-specified default params in sample csv

jia1:
## Problem

Closes [SGC-197](https://linear.app/ogp/issue/SGC-197/set-default-language-in-bulk-send-csv-to-english)

## Solution

Create an optional `defaultParams` prop for `SampleCsv`.

### DIFF
--- a/frontend/src/components/common/sample-csv/SampleCsv.tsx
+++ b/frontend/src/components/common/sample-csv/SampleCsv.tsx
@@ -1,5 +1,5 @@
 import download from 'downloadjs'
-import { without, times, constant } from 'lodash'
+import { without } from 'lodash'
 
 import TextButton from '../text-button'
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/sample-csv/SampleCsv.tsx
```diff
@@ -39,11 +41,19 @@ const SampleCsv = ({
         ...without(variableHeaders, ...fixedHeaders),
       ]
 
+      const variableRow = Array(headers.length - 1).fill('abc')
```
KishenKumarrrrr, line 44: Am I correct to say that using `abc` is acceptable here because we don't have default parameters for all the different templates' fields?

KishenKumarrrrr (approved):
No issues with the changes, just a small question about the use of `abc`

frontend/src/components/common/sample-csv/SampleCsv.tsx
jia1: While I'm not able to know what the original committer was thinking, I would believe this is better than seemingly realistic values. Otherwise, might forget to change. As for abc vs some other placeholder value, I prefer something more eye-catching like TEST. But, out of scope of this task/PR.

jia1: I might have misunderstood your qn, I'll just add on to see if this lands:

I only want to add default params for `language` and nothing else. Since the original behaviour uses `abc`, I'll do the same.

PR #2173, opened 2023-08-14
feat: un-hardcode WA template label

jia1:
## Problem

Production will be using a different template label for pre-call notification. We need a way to configure the label for each environment.

## Solution

Create the `WA_PRECALL_TEMPLATE_LABEL` env var and use it.

## Deployment Checklist

- [x] Env vars for staging and production are already set
- [x] Update env var for dev in 1Password
- [x] Verify the if-check works in staging

## Post-deployment Checklist

- [ ] Update prod db to use the new template label


KishenKumarrrrr (approved):
LGTM, can I check if this will affect the current whitelisting procedure?

jia1:
> LGTM, can I check if this will affect the current whitelisting procedure?

The template label will become `sgc_notify_upcoming_call_2` instead of `sgc_notify_upcoming_call_1`. No new db records.

But if the whitelisting script / whitelist contains reference to `sgc_notify_upcoming_call_1`, then need to update it to `sgc_notify_upcoming_call_2`.

PR #2174, opened 2023-08-14
feat: un-remove language from govsg messages table and sort message data

jia1:


ihwk1996 (approved):
Lgtm. The sort is a good idea!

PR #2176, opened 2023-08-14
feat: disable resend until delivered

jia1:
## Problem

We want to disable to Resend button when the pre-call notification message has not reached the recipient's phone. We also add a tooltip to explain why the button was disabled.

Closes [SGC-194](https://linear.app/ogp/issue/SGC-194/disable-resend-button-when-msg-not-delivered-with-tooltip-explaining)

### Before
Screenshot 2023-08-14 at 6 37 16 PM

### After
Screenshot 2023-08-14 at 6 35 13 PM


KishenKumarrrrr (approved):
LGTM

PR #2177, opened 2023-08-15
feat: debounce search govsg messages

jia1:
We should debounce search so that we don't over-search (e.g. user might still be typing).

For everything else i.e. first render, handle page change, we don't debounce because there is no change to search input.


KishenKumarrrrr (approved):
LGTM

PR #2179, opened 2023-08-15
fix: handle JSON attachment edge case

tan-zx:
response to [incident](https://opengovproducts.slack.com/archives/C015VQQ3HMG/p1692068456160179)

tested on local, safe to deploy

### DIFF
--- a/backend/src/core/middlewares/file-attachment.middleware.ts
+++ b/backend/src/core/middlewares/file-attachment.middleware.ts
@@ -4,6 +4,7 @@ import config from '@core/config'
 import { ensureAttachmentsFieldIsArray } from '@core/utils/attachment'
 import { isDefaultFromAddress } from '@core/utils/from-address'
 import {
+  ApiAttachmentFormatError,
   ApiAttachmentLimitError,
   ApiAuthorizationError,
   ApiNotFoundError,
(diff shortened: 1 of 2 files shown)


backend/src/core/middlewares/file-attachment.middleware.ts
```diff
@@ -58,6 +59,10 @@ async function checkAttachmentValidity(
   _res: Response,
   next: NextFunction
 ): Promise<void> {
+  if (!req.files?.attachments && req.body.attachments) {
```
jia1, line 62: If both `req.files?.attachments` and `req.body.attachments` are defined, then we are still ok to continue running this function?

jia1 (approved):

backend/src/core/middlewares/file-attachment.middleware.ts
tan-zx: Hmm good point. In practice, that should never happen because `req.files?.attachments` can only be defined if they use multi-part upload and `req.body.attachments` can. only be defined if they're using JSON. 

But a blanket ban on all JSON `attachments` is more consistent with what our `openapi.yaml` says. Let me change it then! Thanks!

tan-zx: Hmm no this doesn't play well with [the tests](https://github.com/opengovsg/postmangovsg/actions/runs/5863380143/job/15896728265?pr=2179), however. Let me revert the change!

jia1: Thank you, Zi Xiang!

tan-zx: haha you've merged alr! looking at the code now, I think it's clearer if the validation is done here: 


but no matter!

PR #2180, opened 2023-08-15
chore: log resolution or rejection from delivered status update

jia1:
## Problem

Helps to troubleshoot [SGC-201](https://linear.app/ogp/issue/SGC-201/delivered-status-updates-are-not-reflected-in-db)


jia1:
We don't think this fixes things. Closing.

PR #2183, opened 2023-08-17
feat: show passcode asap

jia1:
## Problem

Closes [SGC-209](https://linear.app/ogp/issue/SGC-209/show-passcode-immediately)

## Notes

- Remember to run pending migrations in the various envs

### DIFF
--- a/backend/src/govsg/services/govsg-verification-service.ts
+++ b/backend/src/govsg/services/govsg-verification-service.ts
@@ -1,6 +1,7 @@
 import config from '@core/config'
 import { whatsappService } from '@core/services'
 import { GovsgMessage, GovsgVerification } from '@govsg/models'
+import { createPasscode } from '@govsg/utils/passcode'
 import {
   WhatsAppId,
   WhatsAppApiClient,
@@ -8,8 +9,6 @@ import {
   WhatsAppTemplateMessageToSend,
   WhatsAppLanguages,
 } from '@shared/clients/whatsapp-client.class/types'
-
-import { randomInt } from 'node:crypto'
 
 export const sendPasscodeCreationMessage = async (
   whatsappId: WhatsAppId,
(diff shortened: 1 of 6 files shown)


backend/src/govsg/services/govsg.service.ts
```diff
@@ -170,20 +172,35 @@ export function uploadCompleteOnChunk({
...
+            })
+          }
+        },
+        benchmark: true,
+        returning: true,
+      }
+    )
+    await Promise.all(
+      govsgMessages.map((message) =>
+        GovsgVerification.create(
```
KishenKumarrrrr, line 195: Could I ask why we can't make use of `bulkCreate` here?

KishenKumarrrrr (approved):
Just some small comments, but otherwise LGTM

backend/src/govsg/services/govsg.service.ts
jia1: Cus I was silly. Using `bulkCreate` now!

backend/src/govsg/services/govsg-verification-service.ts
```diff
@@ -31,20 +30,23 @@ export const sendPasscodeCreationMessage = async (
...
-}
-
 export const storePrecreatedPasscode = async (
   govsgMessageId: GovsgMessage['id'],
   passcodeCreationWamid: MessageId
 ): Promise<GovsgVerification> => {
-  const passcode = createPasscode()
-  return await GovsgVerification.create({
-    govsgMessageId,
+  const govsgVerification = await GovsgVerification.findOne({
```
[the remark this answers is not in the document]
jia1: To the best of my knowledge, `upsert` requires the knowledge of the entire row's data / identifier. There's no unique constraint on the field I'm searching with.

PR #2185, opened 2023-08-18
(PR 2) fix: create passcode only if using precall template

jia1:
## Problem

We should run passcode functionality and show passcode only if the message template is a pre-call template.

Closes [SGC-213](https://linear.app/ogp/issue/SGC-213/add-missed-call-and-house-visit-templates)

## Solution

- [x] Add if-condition before every creation of `GovsgVerification`
- [x] Change absent passcode text to `N/A`
- [x] Hide passcode column if the message template used is not a pre-call template
- [x] Temporarily disable Resend button for non-pre-call template messages

### Other improvements

- [x] Change `tr height` to `tr min-height` so that the message data column can expand vertically and we don't break other existing tables relying on `tr height`. Out-of-scope for this PR: Handling horizontal overflow

## Screenshots
Screenshot 2023-08-18 at 10 17 14 AM

### DIFF
--- a/backend/src/govsg/middlewares/govsg-verification.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-verification.middleware.ts
@@ -70,18 +70,10 @@ export const listMessages = async (
       officer: officer_name,
     }
   })
-  if (count > 0) {
-    const row = rows[0]
-    return res.json({
-      messages,
-      total_count: count,
-      has_passcode: !!row.govsgVerification?.passcode,
-    })
-  }
   return res.json({
     messages,
     total_count: count,
-    has_passcode: false,
+    has_passcode: !!rows[0]?.govsgVerification?.passcode,
   })
 }
 
--- a/backend/src/govsg/middlewares/govsg-verification.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-verification.middleware.ts
@@ -65,7 +65,7 @@ export const listMessages = async (
       name: recipient_name,
       mobile: recipient,
       data: JSON.stringify(remainingParams),
-      passcode: row.govsgVerification.passcode,
+      passcode: row.govsgVerification?.passcode,
       sent: row.sentAt,
       officer: officer_name,
     }
--- a/backend/src/govsg/middlewares/govsg-verification.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-verification.middleware.ts
@@ -70,9 +70,18 @@ export const listMessages = async (
       officer: officer_name,
     }
   })
+  if (count > 0) {
+    const row = rows[0]
+    return res.json({
+      messages,
+      total_count: count,
+      has_passcode: !!row.govsgVerification?.passcode,
+    })
+  }
   return res.json({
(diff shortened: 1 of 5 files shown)


backend/src/govsg/middlewares/govsg-verification.middleware.ts
```diff
@@ -65,14 +65,23 @@ export const listMessages = async (
...
+    const row = rows[0]
+    return res.json({
+      messages,
+      total_count: count,
+      has_passcode: !!row.govsgVerification?.passcode,
+    })
+  }
   return res.json({
     messages,
     total_count: count,
```
KishenKumarrrrr, line 83: In this case, `messages` is an empty array and `count` is 0? Would it be clearer to explicitly state the values and add a comment explaining the case?

backend/src/govsg/services/govsg.service.ts
```diff
@@ -139,6 +140,20 @@ const getLanguageCode = (language: string | undefined) => {
   return WhatsAppLanguages[key as keyof typeof WhatsAppLanguages]
 }
 
+const isUsingPrecallTemplate = async (govsgMessage: GovsgMessage) => {
```
KishenKumarrrrr, line 143: nit: Is this function name accurate? Having a passcode goes beyond just precall templates right? Perhaps `isTemplateWithPasscode` is better?

KishenKumarrrrr (commented):
I have an additional question. If passcode is now optional, do we need to modify the `GovsgVerification` model?

jia1:
> I have an additional question. If passcode is now optional, do we need to modify the `GovsgVerification` model?

Since some templates don't need passcode, then we don't even create `GovsgVerification` for those messages. This PR addresses that. Let me know if this doesn't answer your qn!

backend/src/govsg/services/govsg.service.ts
jia1: For current use cases, the name is accurate. But sure, let me rename that.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
jia1: Yep. Got it, let me refactor this...

jia1: Done!

backend/src/govsg/services/govsg.service.ts
jia1: Done!

backend/src/govsg/middlewares/govsg-verification.middleware.ts
KishenKumarrrrr: Did you make the change? 🤔

jia1: I refactored and that block is not needed anymore. See 2nd latest commit.

KishenKumarrrrr: My bad, thanks for making the change.

PR #2187, opened 2023-08-18
(PR 3) feat: implement click to reveal passcode

jia1:
## Problem

We want to track officer's usage of passcode. We do this by prompting them to Click to reveal passcode.

Closes [SGC-214](https://linear.app/ogp/issue/SGC-214/make-passcode-click-to-reveal-and-log-click-events)

And [SGC-215](https://linear.app/ogp/issue/SGC-215/remove-officer-verification-button) (see #2189)

## Solution

- Leveraging the `details` HTML element ([docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details))

## Notes

cursor not set to pointer when hovering over Click to reveal. This will be done in a separate PR.

### DIFF
--- a/backend/src/govsg/middlewares/govsg-verification.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-verification.middleware.ts
@@ -128,3 +128,18 @@ export const resendPasscodeCreationMessage = async (
     passcode_creation_wamid: passcodeCreationWamid,
   })
 }
+
+export const trackPasscodeReveal = async (
+  req: Request,
+  res: Response
+): Promise<Response | void> => {
+  const { govsg_message_id: govsgMessageId } = req.body
+  const userClickedAt = new Date()
+  await GovsgVerification.update(
+    { userClickedAt },
+    { where: { govsgMessageId } }
+  )
+  return res.json({
+    user_clicked_at: userClickedAt,
+  })
+}
--- a/backend/src/govsg/middlewares/govsg-verification.middleware.ts
+++ b/backend/src/govsg/middlewares/govsg-verification.middleware.ts
@@ -70,6 +70,14 @@ export const listMessages = async (
       officer: officer_name,
     }
   })
+  if (count > 0) {
+    const row = rows[0]
+    return res.json({
+      messages,
+      total_count: count,
+      has_passcode: !!row.govsgVerification?.passcode,
+    })
+  }
   return res.json({
     messages,
     total_count: count,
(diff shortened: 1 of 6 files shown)


backend/src/govsg/middlewares/govsg-verification.middleware.ts
```diff
@@ -128,3 +128,25 @@ export const resendPasscodeCreationMessage = async (
...
+): Promise<Response | void> => {
+  const { govsg_message_id: govsgMessageId } = req.body
+  const userClickedAt = new Date()
+  await GovsgVerification.update(
+    { userClickedAt },
+    {
+      where: {
+        govsgMessageId,
+        userClickedAt: {
+          [Op.eq]: null,
```
KishenKumarrrrr, line 144: Could I ask why this is necessary? From my understanding, we incur a DB call for each invocation of this function and even if we perform the update each time, it still counts as one click in our metrics.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
```diff
@@ -128,3 +128,25 @@ export const resendPasscodeCreationMessage = async (
     passcode_creation_wamid: passcodeCreationWamid,
   })
 }
+
+export const trackPasscodeReveal = async (
+  req: Request,
+  res: Response
+): Promise<Response | void> => {
+  const { govsg_message_id: govsgMessageId } = req.body
+  const userClickedAt = new Date()
```
KishenKumarrrrr, line 137: In `govsg-callback.service`, we [previous converted the date to milliseconds](https://github.com/opengovsg/postmangovsg/blob/9f78ca3ab2fad6084ff9e5a0ffad47aca1456eba/backend/src/govsg/services/govsg-callback.service.ts#L389). Any reason we don't do the same here?

KishenKumarrrrr (approved):
I've got some questions about `govsg-verification.middleware` but other than that LGTM.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
jia1: Just a personal preference of wanting the value to be the timestamp of the first click ever.

backend/src/govsg/middlewares/govsg-verification.middleware.ts
jia1: Because Meta returns seconds from epoch (sporadically mentioned in docs) and Date constructor returns milliseconds from epoch ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)).

This means we don't need any conversion here.

PR #2189, opened 2023-08-18
(PR 4) feat: replace passcode button with passcode in first message

jia1:
## Problem

Closes [SGC-215](https://linear.app/ogp/issue/SGC-215/remove-officer-verification-button)

## Checklist

- [x] Single send
- [x] Bulk send

## Screen recordings

https://drive.google.com/drive/folders/1JkFGh_SSYBMIm9F_YtXxdCCBxiCDl7xp?usp=sharing

## Notes

- cursor not set to pointer when hovering over Click to reveal passcode. Will address this in a separate PR.
- The template name should not mention `OTP` at all. Should be something along the lines of "with officer verification". Will rename when populating templates to production.

### DIFF
--- a/backend/src/govsg/services/govsg.service.ts
+++ b/backend/src/govsg/services/govsg.service.ts
@@ -111,11 +111,18 @@ export function uploadCompleteOnPreview({
   campaignId: number
 }): (data: CSVParams[]) => Promise<void> {
   return async function (data: CSVParams[]): Promise<void> {
+    const paramsWithoutPasscode = template.params!.filter(
+      (param) => param !== 'passcode'
+    ) // TODO: Un-hardcode this
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    UploadService.checkTemplateKeysMatch(data, template.params!)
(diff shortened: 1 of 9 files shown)


backend/src/govsg/services/govsg.service.ts
```diff
@@ -140,8 +147,7 @@ const getLanguageCode = (language: string | undefined) => {
   return WhatsAppLanguages[key as keyof typeof WhatsAppLanguages]
 }
 
-const isUsingPrecallTemplate = async (govsgMessage: GovsgMessage) => {
-  const campaignId = govsgMessage.campaignId
+const isUsingPrecallTemplate = async (campaignId: number) => {
```
KishenKumarrrrr, line 150: Is the name of this function outdated?

KishenKumarrrrr (approved):
LGTM

backend/src/govsg/services/govsg.service.ts
jia1: iirc I renamed this function in some other branch. I'll have to resolve the conflict later.

PR #2186, opened 2023-08-18
Revert "feat: show passcode asap"

jia1:
Reverts opengovsg/postmangovsg#2183

PR #2191, opened 2023-08-18
(PR 1) Revert Revert feat: show passcode asap

jia1:
## Deployment notes

Can run migration before merging. Update: Done!

## Summary of changes

1. Reverts opengovsg/postmangovsg#2186
1. Plus fix foreign key constraint violation when re-picking template and 2nd+ upload of CSV (see screenshot below)
1. AND #2185

### Before
Screenshot 2023-08-18 at 2 55 08 PM

## How to verify db migration in db

```sql
SELECT update_rule, delete_rule
FROM information_schema.referential_constraints
WHERE constraint_name = 'govsg_verification_govsg_message_id_fkey';
```

### DIFF
--- a/backend/src/govsg/models/govsg-verification.ts
+++ b/backend/src/govsg/models/govsg-verification.ts
@@ -25,7 +25,7 @@ export class GovsgVerification extends Model<GovsgVerification> {
   @Column({ type: DataType.BIGINT, allowNull: false })
   govsgMessageId: number
 
-  @Column({ type: DataType.STRING, allowNull: false })
+  @Column({ type: DataType.STRING })
   passcodeCreationWamid: string
 
   @Column({ type: DataType.STRING, allowNull: false })
--- a/backend/src/govsg/services/govsg-verification-service.ts
+++ b/backend/src/govsg/services/govsg-verification-service.ts
@@ -1,6 +1,7 @@
 import config from '@core/config'
 import { whatsappService } from '@core/services'
 import { GovsgMessage, GovsgVerification } from '@govsg/models'
+import { createPasscode } from '@govsg/utils/passcode'
 import {
   WhatsAppId,
   WhatsAppApiClient,
@@ -8,8 +9,6 @@ import {
   WhatsAppTemplateMessageToSend,
   WhatsAppLanguages,
 } from '@shared/clients/whatsapp-client.class/types'
-
-import { randomInt } from 'node:crypto'
 
 export const sendPasscodeCreationMessage = async (
   whatsappId: WhatsAppId,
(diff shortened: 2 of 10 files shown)


backend/src/govsg/models/govsg-verification.ts
```diff
@@ -25,7 +25,7 @@ export class GovsgVerification extends Model<GovsgVerification> {
   @Column({ type: DataType.BIGINT, allowNull: false })
   govsgMessageId: number
 
-  @Column({ type: DataType.STRING, allowNull: false })
+  @Column({ type: DataType.STRING })
```
KishenKumarrrrr, line 28: nit: Do you think we should make the default behaviour (`allowNull: true`) explicit? I think it would improve readability

backend/src/govsg/services/govsg-verification-service.ts
```diff
@@ -31,20 +30,23 @@ export const sendPasscodeCreationMessage = async (
...
-}
-
 export const storePrecreatedPasscode = async (
   govsgMessageId: GovsgMessage['id'],
   passcodeCreationWamid: MessageId
 ): Promise<GovsgVerification> => {
-  const passcode = createPasscode()
-  return await GovsgVerification.create({
-    govsgMessageId,
+  const govsgVerification = await GovsgVerification.findOne({
```
KishenKumarrrrr, line 37: Just checking on this, `upsert` did not work?

backend/src/govsg/services/govsg.service.ts
```diff
@@ -212,12 +228,22 @@ export async function processSingleRecipientCampaign(
...
+    const govsgMessage = await GovsgMessage.create(
+      {
+        campaignId,
+        recipient: data.recipient,
+        languageCode,
+        params: data,
+      } as GovsgMessage,
+      { transaction, returning: true }
+    )
+    await GovsgVerification.create(
```
KishenKumarrrrr, line 240: nit: It looks to me like both the `create` calls are related to each other. Do you think it would be beneficial to make use of a [nested create](https://sequelize.org/docs/v6/advanced-association-concepts/creating-with-associations/)?

(Sorry if I didnt pick this up the previous time I reviewed this PR.)

KishenKumarrrrr (commented):
Some small comments but LGTM otherwise. Let me know once the migrations have been run and I'll approve the PR.

backend/src/govsg/models/govsg-verification.ts
jia1: Probably not, otherwise we'll have to add all missing `allowNull: true` to be consistent.

backend/src/govsg/services/govsg-verification-service.ts
jia1: I probably mentioned this before - upsert requires uniqueness / data from the entire row. So the work to make upsert work is significant. I feel it is not worth the effort to refactor the code by tightening constraints.

On the other hand, we could assess if we want tighter constraints in our future schema.

backend/src/govsg/services/govsg.service.ts
jia1: I can try this out in the next PR if that's ok.

KishenKumarrrrr (approved):

PR #2192, opened 2023-08-21
feat: remove passcode button

jia1:
I think I merged this #2189 too late and the changes aren't in `master`.

Here, I rebase on top of `master` and fix any remaining issues (e.g. removing remnants of passcode button code).

### DIFF
--- a/backend/src/govsg/services/govsg.service.ts
+++ b/backend/src/govsg/services/govsg.service.ts
@@ -111,11 +111,18 @@ export function uploadCompleteOnPreview({
   campaignId: number
 }): (data: CSVParams[]) => Promise<void> {
   return async function (data: CSVParams[]): Promise<void> {
+    const paramsWithoutPasscode = template.params!.filter(
+      (param) => param !== 'passcode'
+    ) // TODO: Un-hardcode this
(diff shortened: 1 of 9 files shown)


backend/src/govsg/services/govsg.service.ts
```diff
@@ -111,11 +111,18 @@ export function uploadCompleteOnPreview({
   campaignId: number
 }): (data: CSVParams[]) => Promise<void> {
   return async function (data: CSVParams[]): Promise<void> {
+    const paramsWithoutPasscode = template.params!.filter(
```
KishenKumarrrrr, line 114: Do we need to address this warning?

KishenKumarrrrr (approved):
There are some warnings from the CI. I think they are acceptable but we might want to add code comments explaining why its like that + disable the checks for that line

backend/src/govsg/services/govsg.service.ts
jia1: By warning do you mean `!`?

The value has to be valid or we don't want to continue ahead.

`!` is not my preference but I would deem this out-of-scope. We should really have caught this during code review but we weren't involved back then.

PR #2196, opened 2023-08-22
feat: reject invalid language

jia1:
## Problem

We want to reject invalid language values, instead of falling back to English.

Closes [SGC-221](https://linear.app/ogp/issue/SGC-221/reject-misspelt-language)

## Notes

This PR is backward-compatible.

If the language column is entirely absent (like how it is supposed to be right now), we fall back to English.

On the other hand, if a language is specified, then it has to be valid.


ihwk1996 (approved):
Lgtm

ihwk1996 (approved):
Lgtm

PR #2197, opened 2023-08-22
fix: prevent undefined params in govsg recipients frontend

jia1:
https://open-government-products-re.sentry.io/issues/4411696399/?alert_rule_id=1124293&alert_type=issue&project=5261024&referrer=slack

### DIFF
--- a/frontend/src/components/dashboard/create/govsg/GovsgRecipients.tsx
+++ b/frontend/src/components/dashboard/create/govsg/GovsgRecipients.tsx
@@ -1,4 +1,5 @@
 import { i18n } from '@lingui/core'
+import { filter } from 'lodash'
 import {
   Dispatch,
   SetStateAction,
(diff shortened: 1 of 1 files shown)


ihwk1996 (commented):
Can u use _.filter()? I think it's safe for null/undefined

jia1:
> Can u use _.filter()? I think it's safe for null/undefined

I tried reading the Lodash docs but can't get an idea of what `_.filter` returns if the arg is `null/undefined`. And the docs don't say whether it accepts `null/undefined`.

If you have a quick ans for ^ that'll be great. In the meantime, I'll check on my end.

The return value of `params` must be an array no matter what or the spread operator below will fail.

stanleynguyen (approved):

ihwk1996 (approved):
lgtm

PR #2198, opened 2023-08-22
fix: add key attribute to details element

jia1:
## Problem

- (BUG) The passcode badge component is being reused i.e. reveal first row = first rows in subsequent pages are also revealed already.

## Solution

Add `key` attribute to the `details` element in `PasscodeBadge` ([docs](https://legacy.reactjs.org/docs/lists-and-keys.html#keys)).


ihwk1996 (approved):
Lgtm

jia1:
Thank you, Ivan.

PR #2199, opened 2023-08-22
feat(worker): implement alert for campaign with errors

stanleynguyen:
## Problem

Alert on GovSG campaign with errored messages

## Solution

Check if there're errored messages for a campaign at the finalize step in logger worker

## Deployment Checklist

- [ ] Add `SGC_CAMPAIGN_ALERT_WEBHOOK` env var

### DIFF
--- a/worker/src/core/loaders/message-worker/index.ts
+++ b/worker/src/core/loaders/message-worker/index.ts
@@ -16,6 +16,7 @@ import { NotificationService } from '@core/services/notification.service'
 import { TemplateClient, XSS_EMAIL_OPTION } from '@shared/templating'
 import MailClient, { MailToSend } from '@shared/clients/mail-client.class'
 import Govsg from './govsg.class'
+import axios from 'axios'
 
 require('module-alias/register') // to resolve aliased paths like @core, @sms, @email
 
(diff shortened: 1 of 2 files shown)


jia1 (approved):

worker/src/core/loaders/message-worker/index.ts
```diff
@@ -169,6 +170,45 @@ const finalize = tracer.wrap(
...
+        }
+
+        const messagesWithErr = rawMessages as Array<{
+          id: number
+          error_code: string
+          error_description: string
+        }>
+        if (
+          messagesWithErr.length === 0 ||
+          !config.get('sgcCampaignAlertChannelWebhookUrl')
```
jia1, line 189: Not blocking but:

Should we be throwing when `!config.get('sgcCampaignAlertChannelWebhookUrl')`? We want to be alerted, right?

worker/src/core/loaders/message-worker/index.ts
```diff
@@ -169,6 +170,45 @@ const finalize = tracer.wrap(
...
+          text += `| ${m.id} | ${m.error_code} | ${m.error_description} |\n`
+        })
+        text += '***********'
+        await axios
+          .post(config.get('sgcCampaignAlertChannelWebhookUrl'), {
+            text: text,
+          })
+          .catch((e) => {
+            logger.error({
+              message: 'Failed to report errors to #sgc-campaign-alerts',
```
jia1, line 204: This sounds like a never-ending cycle but:

Do we get alerted through other means when this happens?

stanleynguyen: we could set up an alert for this alert lol, but the log agent could go down as well 😂

jia1: Ok, this won't end well...

Let's keep it as-is.

PR #2200, opened 2023-08-22
fix: use snake case when reading language code from db json

jia1:
Language code is stored as `language_code` in the JSON column. It doesn't get transformed into camel-case (cus nested). We should thus use snake case when reading language code.


backend/src/govsg/middlewares/govsg-transactional.middleware.ts
```diff
@@ -89,7 +89,9 @@ export const InitGovsgTransactionalMiddleware =
       }
       const isLanguageAvailable =
         govsgTemplate.multilingualSupport.findIndex(
-          (v) => v.languageCode === languageCode
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
+          (v) => v.language_code === languageCode
```
stanleynguyen, line 94: I thought we store them as camelCase? https://github.com/opengovsg/postmangovsg/blob/f330bbca586491df2328f5686de3776cceedd895/backend/src/govsg/models/govsg-template.ts#L20

jia1: Nope, it's `language_code`.

stanleynguyen: @jia1 I think the one that needs fixing is the template data in our db, not the code

jia1: If we change the template data, we'll need to change them all and fix the breakage to non-transactional flow. It's more work.

jia1: I'll change the type of `GovsgTemplateLanguageMetadata`

KishenKumarrrrr (approved):
LGTM

PR #2201, opened 2023-08-22
feat(sst): add cron monitor for SGC whatsapp webhook

stanleynguyen:
## Problem

We want to monitor if there're webhooks coming back for all of our Gov.sg messages. No messages should stay in the `ACCEPTED` state for too long as there will always be `SENT` status webhooks.

## Solution

Implement a cron to check for the 2nd-immediate previous 10-minute window to see if there're any `ACCEPTED` messages.

## Deployment Checklist

- [x] create cronitor monitor and set `CRONITOR_CODE_GOVSG_WEBHOOK_MONITOR` for SST parameter store
- [x] set `SGC_ALERT_WEBHOOK` for SST parameter store

### DIFF
--- a/sst/packages/functions/src/cron-jobs/govsg-webhook-monitor/cron.ts
+++ b/sst/packages/functions/src/cron-jobs/govsg-webhook-monitor/cron.ts
@@ -1,0 +1,56 @@
+import axios from 'axios'
+import { sql } from 'drizzle-orm'
+import { Config } from 'sst/node/config'
+
+import PostmanDbClient from '@/core/database/client'
+import { govsgMessages } from '@/core/models'
+
+export async function handler() {
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const cronitor = require('cronitor')(Config.CRONITOR_URL_SUFFIX) // common to all jobs on our shared Cronitor account
+    const invokeFunction =
+      process.env.IS_LOCAL === 'true'
+        ? async function () {
+            console.log('Running cron locally')
+            await checkMessagesNotGettingWebhooks()
+          }
+        : cronitor.wrap(
+            Config.CRONITOR_CODE_GOVSG_WEBHOOK_MONITOR,
+            async function () {
+              await checkMessagesNotGettingWebhooks()
+            },
+          )
+    await invokeFunction()
+  } catch (error) {
+    console.log(error)
+    throw error
+  }
+}
+
+async function checkMessagesNotGettingWebhooks() {
+  const db = new PostmanDbClient(Config.POSTMAN_DB_URI).getClient()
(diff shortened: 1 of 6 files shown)


sst/packages/functions/src/cron-jobs/govsg-webhook-monitor/cron.ts
```diff
@@ -0,0 +1,56 @@
...
+
+async function checkMessagesNotGettingWebhooks() {
+  const db = new PostmanDbClient(Config.POSTMAN_DB_URI).getClient()
+  const noWebhookMessages = await db
+    .select({})
+    .from(govsgMessages)
+    .where(
+      // the reason we want to query for the previous 10-minute block instead of
+      // the immediate one is to prevent messages from campaigns currently being sent
+      // out from being flagged (This is assuming a campaign won't take more than
```
jia1, line 39: Is this assumption an estimation derived from Grafana?

sst/packages/functions/src/cron-jobs/govsg-webhook-monitor/cron.ts
```diff
@@ -0,0 +1,56 @@
...
+      // 10 minutes to receive all the `SENT` webhooks)
+      sql`status = 'ACCEPTED' AND updated_at < NOW() - INTERVAL '10 minutes' AND updated_at >= NOW() - INTERVAL '20 minutes'`,
+    )
+  if (noWebhookMessages.length === 0) {
+    return
+  }
+  try {
+    await axios.post(Config.SGC_ALERT_WEBHOOK, {
+      text: `<!subteam^S03JWLMTBTK|postmangineers>\n***NEW ALERT***\nThere're ${noWebhookMessages.length} messages that didn't get a webhook.\n******`,
```
jia1, line 48: Can we add more info like govsg message `id`?

jia1 (approved):

sst/packages/functions/src/cron-jobs/govsg-webhook-monitor/cron.ts
stanleynguyen: no so the average webhooks for sent actually comes back much faster (about a few seconds), but I'm just taking 10 minutes just to be safe, and I think our SLA to detect this kind is 30 mins so shud be ok

sst/packages/functions/src/cron-jobs/govsg-webhook-monitor/cron.ts
stanleynguyen: If this happens, I think the IDs could be quite long depending on whether there're a lot of messages getting sent out. So I dont think that will scale well. What we could do is to include a standard incident response in the play book for this kind

PR #2205, opened 2023-08-28
fix: bugs in whitelisting script

KishenKumarrrrr:
## Problem

- `update_whitelist.ts` : The previous script added all users regardless of whether or not they are part of the experimental feature.
- `insert_experimental_govsg_user.ts` : The previous script did not check to see whether or not the user should be deleted from the DB. 

## Solution

- `update_whitelist.ts` : When querying users, we do so from the `GovsgTemplatesAccess` table instead of `Users` table.
- `insert_experimental_govsg_user.ts` : Perform an extra check on the `to remove` column of the CSV file (5th column). If this value is `'Y'`, we do not add the user to the DB.

## Deployment Checklist

NA

### DIFF
--- a/backend/src/scripts/whitelisting/constants.ts
+++ b/backend/src/scripts/whitelisting/constants.ts
@@ -1,8 +1,12 @@
 export const SECTION_DIVIDER =
   '\n=============================================================\n'
 
-export const OGP_TEMPLATE_ACCESS = []
+export const OGP_TEMPLATE_ACCESS = ['sgc_trial1_feedback_message']
 
 export const MCI_TEMPLATE_ACCESS = []
 
-export const ALL_TEMPLATE_ACCESS = []
+export const ALL_TEMPLATE_ACCESS = [
+  'sgc_notify_upcoming_house_visit_1',
+  'sgc_notify_missed_call_1',
+  'sgc_notify_upcoming_call_2',
+]
--- a/backend/src/scripts/insert_experimental_govsg_user.ts
+++ b/backend/src/scripts/insert_experimental_govsg_user.ts
@@ -29,7 +29,7 @@ void (async function main() {
       'temp/Agency Whitelist - Whitelist.csv'
     )
   )
-    .filter((r) => r[4] === 'N')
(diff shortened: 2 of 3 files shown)


jia1:
I'm a bit confused over why we query from `GovsgTemplatesAccess`:

If we want to care if a user is part of the experimental feature, we should be querying the `user_experimental`

jia1:
I also noticed the script is rather complex for what it's aiming for. I would normally expect a db call for each set of users:

First, get all GOVSG users. Give them access to the templates.

Then, get all MCI GOVSG folks. Give them access to the MCI templates.

Finally, get all OGP GOVSG folks. Give them access to the OGP templates.

This is so we don't need to do additional steps like:

```
  const ogpOfficers = agencyUsersToUpdate.filter(
    (u) => u.user.email.split('@')[1] === 'open.gov.sg'
  )
```

and:

```
const allUsers = [...agencyUsersToUpdate, ...usersToUpdate]
```

which are easily missed out and require tracing. I understand that there's some performance tradeoff but it could be worth it if we want to make this easily readable for everyone running the script.

jia1 (approved):

backend/src/scripts/whitelisting/constants.ts
```diff
@@ -1,8 +1,12 @@
...
-export const OGP_TEMPLATE_ACCESS = []
+export const OGP_TEMPLATE_ACCESS = ['sgc_trial1_feedback_message']
 
 export const MCI_TEMPLATE_ACCESS = []
 
-export const ALL_TEMPLATE_ACCESS = []
+export const ALL_TEMPLATE_ACCESS = [
+  'sgc_notify_upcoming_house_visit_1',
+  'sgc_notify_missed_call_1',
+  'sgc_notify_upcoming_call_2',
```
jia1, line 11: Just noticed: We gonna use `sgc_notify_upcoming_call_3`.

joshualiangxy (changes requested):

KishenKumarrrrr:
We no longer need to do this.

PR #2212, opened 2023-09-05
Fix/hash collisions

KishenKumarrrrr:
## Context

The existing hashing implementation for Twilio secrets is problematic because `bcrypt` only uses the [first 72 bytes of a string to generate the resultant hash](https://github.com/kelektiv/node.bcrypt.js#security-issues-and-concerns). This resulted in a bug in the Postman application where different sets of Twilio secrets produced the exact same hash, causing the secrets to be overwritten due to hash collisions. 

## Approach

After the initial hash is generated from the user-submitted Twilio secrets, the hash is never subsequently used again to verify the authenticity of the credentials. The hash is only used to identify and fetch the particular set of credentials from AWS Secrets Manager. 

Since the hash is only used for identification of the secrets, we can circumvent the hash collision problem by generating a `uuid` for the Twilio secrets instead.

**Bug Fixes**:

- `backend/src/sms/middlewares/sms.middleware.ts`: `credentialName` is now generated via `uuid` instead of `bcrypt`

## Tests

No new test were written for this change. Manual testing was performed in staging to verify the changes and old tests were updated. 

Test updated: 
- `backend/src/sms/routes/tests/sms-campaign.routes.test.ts`
- `backend/src/sms/routes/tests/sms-settings.routes.test.ts`

### DIFF
--- a/backend/src/__mocks__/uuid/index.ts
+++ b/backend/src/__mocks__/uuid/index.ts
@@ -1,8 +1,8 @@
 // Defining these in global scope so tests can mock their implementation
-export const HASHED_CREDS = 'HASHED_CREDS'
+export const MOCK_UUID = 'MOCKED_UUID'
 
 const v4 = jest.fn()
 
-v4.mockReturnValue(HASHED_CREDS)
+v4.mockReturnValue(MOCK_UUID)
 
 export { v4 }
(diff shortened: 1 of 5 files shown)


ihwk1996 (approved):
lgtm

PR #2214, opened 2023-09-07
fix: use normalised recipient in templateMessageToSend

jia1:
We previously fixed normalising recipient before querying flamingo for which WA client id to use.

But, resend failed because recipient in templateMessageToSend is not normalised.


ihwk1996 (approved):
lgtm

jia1:
For now, we are not merging until we have an idea of the root cause.

jia1:
Found root cause: https://developers.facebook.com/docs/whatsapp/on-premises/reference/contacts/ Cmd+F "predictable"

PR #2227, opened 2023-09-22
feat: remove click to reveal

jia1:
As requested by Paul and users

## Before
click-to-reveal-before

## After
click-to-reveal-after

## Won't do

- Removing the `user_clicked_at` column


KishenKumarrrrr (approved):
LGTM

PR #2231, opened 2023-09-27
Fix/sgid bug

KishenKumarrrrr:
## Problem

Users have issues logging in via SGID. We are not sure of the root cause for now. As a stop gap, we will add some cleaning of the data returned by SGID and log the data returned to aid in our troubleshooting.

## Deployment Checklist

NA


ihwk1996 (approved):
lgtm

PR #2233, opened 2023-10-20
refactor: update SGID scope and POCDEX profile field names

KishenKumarrrrr:
## Problem

The SGID team made some breaking changes to the POCDEX login flow, as such in this PR, I have updated our BE and FE code to adhere to the newest changes. 

## Solution

- POCDEX scope has changed from `pocdex.public_officer_employments` to `pocdex.public_officer_details`
- The structure of data from the POCDEX scope has changed as well. The fields are now snake_case instead of camelCase

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -209,7 +209,7 @@
         "filename": "backend/src/core/services/auth.service.ts",
         "hashed_secret": "f114703480996b273d7e57cbd195b4ab1e70a21b",
         "is_verified": false,
-        "line_number": 65
+        "line_number": 64
       }
     ],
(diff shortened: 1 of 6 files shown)


jia1 (approved):

jia1:
@KishenKumarrrrr is sgID broken now, despite us not changing the versions?

If so, that's rather unacceptable.

jia1:
Have asked Kishen. According to Kishen, the sgID team is maintaining both arrangements within the same client version. So we're good!

KishenKumarrrrr:
We no longer need to do this.

PR #2235, opened 2023-11-16
Refactor/remove phonebook

KishenKumarrrrr:
## Problem

Phonebook is shutting down 😢 . In this PR, I remove phonebook from postman.

## Deployment Checklist

- [ ] Remove unused env vars from AWS SM (can be done after deployment)


jia1:
This method: `const getPresignedUrl = async (): Promise<{` does not seem to be in use anymore. But I don't think it's essential to remove it either.

PR #2236, opened 2023-11-23
fix: ui bug

KishenKumarrrrr:
## Problem

e2e tests are failing because it relied on a subtitle that was part of the phonebook UI. Added it back here.


jia1 (approved):
THANK YOU

PR #2237, opened 2023-11-23
Fix/UI bug

KishenKumarrrrr:
## Problem

Protected campaigns should conditionally rendered the "Step 2" subtitle


jia1 (approved):
SLAYYYY

PR #2240, opened 2024-01-04
refactor: change protected and unsubscribe url

KishenKumarrrrr:
## Problem

In this PR, I change some hardcoded variables in postman that will help facilitate our DNS migration on staging


jia1 (approved):
ty!

PR #2241, opened 2024-01-24
feat: disable tracking in transactional email

KishenKumarrrrr:
## Problem

AWS SES includes a 1x1 tracking pixel to detect whether or not a user has opened/read an email.  GSIBs block this tracking pixel which causes it to appear as a little red X at the bottom of emails. This makes the emails appear suspicious.

In this PR, I add a feature to enable transactional email users to turn off this tracking feature via a new `disable_tracking` flag. This would prevent this issue from popping up. 

## Solution

The events that are tracked by SES is determined by the configuration set that SES uses when sending out the email. This configuration set is determined by the value of the `X-SES-CONFIGURATION-SET` header. Configuration sets are defined via the AWS console. 

I've added an additional configuration set (`postman-email-no-tracking`) to `ap-southeast-2` that does not include tracking `Open` and `Read` states. This prevents the gif from being added. 

Additionally, the transactional email request body now includes an optional `disable_tracking` field that users can set to `true` if they don't want the tracking pixel to be added. When tracking is disabled, we will use the `postman-email-no-tracking` configuration set instead of the default `postman-email-open` configuration set.

### New Env Vars
- `BACKEND_SES_NO_TRACKING_CONFIGURATION_SET` : `Name of SES configuration set that does not include tracking open/read events`

### DIFF
--- a/backend/src/core/config.ts
+++ b/backend/src/core/config.ts
@@ -83,6 +83,7 @@ interface ConfigSchema {
   }
   mailFrom: string
   mailConfigurationSet: string
+  noTrackingMailConfigurationSet: string
   mailVia: string
   mailDefaultRate: number
   transactionalEmail: {
@@ -467,6 +468,11 @@ const config: Config<ConfigSchema> = convict({
     doc: 'The configuration set specified when sending an email',
     default: 'postman-email-open',
     env: 'BACKEND_SES_CONFIGURATION_SET',
+  },
+  noTrackingMailConfigurationSet: {
+    doc: 'AWS SES Configuration set that does not include open and read tracking',
+    default: 'postman-email-no-tracking',
+    env: 'BACKEND_SES_NO_TRACKING_CONFIGURATION_SET',
   },
   mailVia: {
     doc: 'Text to appended to custom sender name',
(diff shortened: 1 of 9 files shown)


backend/src/core/config.ts
```diff
@@ -468,6 +469,11 @@ const config: Config<ConfigSchema> = convict({
     default: 'postman-email-open',
     env: 'BACKEND_SES_CONFIGURATION_SET',
   },
+  noTrackingMailConfigurationSet: {
```
pallani, line 472: can add a comment somewhere that this is the only or chosen way we have decided to control the tracking.

backend/src/email/routes/email-transactional.routes.ts
```diff
@@ -62,6 +62,7 @@ export const InitEmailTransactionalRoute = (
         .items(
           Joi.string().trim().email().options({ convert: true }).lowercase()
         ),
+      disable_tracking: Joi.boolean().optional(),
```
pallani, line 65: we should define the default to be `false` somewhere.

worker/Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN apk update && apk upgrade && apk add --no-cache --virtual builds-deps build-
 
 RUN apk add jq
 
-RUN python3 -m pip install awscli
+RUN apk add --no-cache aws-cli
```
pallani, line 24: add comment on why we added this change. don't need to justify since we will throw away the code soon. just for documentation.

jia1 (approved):

PR #2242, opened 2024-02-13
feat: csv upload via be

KishenKumarrrrr:
## Problem

GSIB users on certain networks are still unable to upload CSV files. 

This PR includes a fix that uses the BE to bypass the whitelisting.

### DIFF
--- a/backend/src/core/middlewares/file-attachment.middleware.ts
+++ b/backend/src/core/middlewares/file-attachment.middleware.ts
@@ -1,7 +1,10 @@
 import { Request, Response, NextFunction } from 'express'
 import fileUpload, { UploadedFile } from 'express-fileupload'
 import config from '@core/config'
-import { ensureAttachmentsFieldIsArray } from '@core/utils/attachment'
+import {
+  ensureAttachmentsFieldIsArray,
+  removeFirstAndLastCharacter,
+} from '@core/utils/attachment'
 import { isDefaultFromAddress } from '@core/utils/from-address'
 import {
   ApiAttachmentFormatError,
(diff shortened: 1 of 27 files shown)


backend/src/core/middlewares/file-attachment.middleware.ts
```diff
@@ -156,11 +161,44 @@ async function streamCampaignEmbed(
...
+  try {
+    // 3. Upload file to presigned URL
+    const response = await axios.put(presignedUrl, uploadedFile.data, {
+      headers: {
+        'Content-Type': uploadedFile.mimetype,
+      },
+      withCredentials: false,
+    })
+    // 4. Return the etag and transactionId to the FE
+    const formattedEtag = removeFirstAndLastCharacter(response.headers.etag)
```
jia1, line 186: Why do you do this, is it because:

```
Removing the first and last character from an ETag is a common technique used to address caching issues. ETags are used in HTTP headers to help with web caching. By removing the first and last character, your friend might be attempting to modify the ETag in order to force a cache revalidation. This can be useful when there are issues with caching and the content is not being updated as expected.
```

frontend/src/services/upload.service.ts
```diff
@@ -63,19 +63,20 @@ async function getMd5(blob: Blob): Promise<string> {
 
 export async function uploadFileWithPresignedUrl(
   uploadedFile: File,
-  presignedUrl: string
-): Promise<string> {
+  _presignedUrl: string // Making this unused because the endpoint below generates its own presignedUrl and uploads the file
```
jia1, line 66: s2pid qn: Why dowan reuse `_presignedUrl`?

jia1 (approved):
LGTM except for some qns. Also understand that this fix may not be merged ultimately as the team is exploring a different solution.

PR #2249, opened 2024-02-23
refactor: bad sgid session returns 400 instead of 500

KishenKumarrrrr:
## Problem

Since Feb 8 2024, the `POST /login/sgid` endpoint is called twice whenever a user attempts to log in via a GSIB device. One of the API calls is made without `req.session.sgid` set and thus fails in the BE and results in a HTTP 500 being returned. 

However, this HTTP 500 does not affect the user's ability to log in and is not reflected in the network tab of the developer console as well. Thus, the impact of this issue is low. Instead of returning a HTTP 500, we are choosing to downgrade the error to a HTTP 400 instead to reflect the seriousness of the issue. 

Our current hypothesis is that this second call is made by SIS instead of our FE due to `*.postman.gov.sg` being whitelisted on SGProxy but the SGID URL having to go through SIS. The emergence of this issue also nicely corresponds to the date we whitelisted `*.postman.gov.sg` which lends credence to the hypothesis.

### DIFF
--- a/backend/src/core/middlewares/auth.middleware.ts
+++ b/backend/src/core/middlewares/auth.middleware.ts
@@ -253,9 +253,15 @@ export const InitAuthMiddleware = (authService: AuthService) => {
     const { code } = req.body
     const logMeta = { code, action: 'verifySgidCode' }
     try {
-      if (!req.session) {
+      /*
+        Since Feb 8 2024, this endpoint is called twice when users attempt to log in via SGID on GSIB machines.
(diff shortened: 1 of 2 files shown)


jia1 (approved):
Thanks!

Have we checked in with sgID, or is this not worth pursuing?

KishenKumarrrrr:
> Thanks!
> 
> Have we checked in with sgID, or is this not worth pursuing?

I think it is more likely a result of the jumping between SIS and the machine itself. This only happens on GSIB so likely not a SGID issue. Since this doesn't impact the user's ability to log in, probably not worth pursuing.

PR #2252, opened 2024-04-18
fix: email attachments

KishenKumarrrrr:
## Problem

Attachments broke when I performed the DNS migration to the legacy site. I think this is because we are injecting an ENV var at build time in amplify. This PR should fix it 🙏


jia1 (approved):
TY!!!!

PR #2253, opened 2024-05-02
refactor: remove SGID login

KishenKumarrrrr:
## Problem

In this PR, I remove SGID login from our application. The team came to this decision after learning that POCDEX was returning employee information of former employees. Since support for POCDEX has ceased, we decided that the best thing to do was to drop support for this login flow. 

## Solution

I've removed all mention to SGID except for in the infra folder. Deleting secrets from infra can be tricky and we should focus on preventing login via SGID first. 

When users click on the Singpass button, they will be greeted by the following modal

Screenshot 2024-05-02 at 8 48 39 PM


backend/package.json
```diff
@@ -33,7 +33,7 @@
     "@aws-sdk/client-sns": "3.370.0",
     "@aws-sdk/rds-signer": "3.370.0",
     "@aws-sdk/s3-request-presigner": "3.370.0",
-    "@opengovsg/sgid-client": "^2.1.0",
+    "@datadog/browser-rum": "^4.43.0",
```
jia1, line 36: Is this intended?

jia1 (approved):

jeantanzj:
@KishenKumarrrrr could you share more about the delay in POCDEX updates? 
I can be contacted on OGP slack at @jean if it's not convenient to share here

PR #2271, opened 2024-08-08
refactor: default from address

KishenKumarrrrr:
## Problem

In this PR, I make changes to the application such that the default mail address is `info@mail.postman.gov.sg`. Since this is a breaking change on the transactional endpoint, we will add an escape hatch to the validation flow such that we will accept both `info@mail.postman.gov.sg` and `donotreply@postman.gov.sg`. 

This change was made because of a directive by PSD.

## Deployment Checklist

- [x] Change value of `BACKEND_SES_FROM` to `Postman.gov.sg <info@mail.postman.gov.sg>`
- [x] Change value of `WORKER_SES_FROM` to `Postman.gov.sg <info@mail.postman.gov.sg>`


backend/src/email/middlewares/email.middleware.ts
```diff
@@ -223,10 +223,17 @@ export const InitEmailMiddleware = (
     const { fromName: defaultFromName, fromAddress: defaultFromAddress } =
       parseFromAddress(config.get('mailFrom'))
 
+    // As part of a PSD directive, we have changed the defaultFromAddress to info@mail.postman.gov.sg.
+    // To prevent any breaking changes, we must now support both the new and old default address
+    const allowedDefaultAddresses = [
+      defaultFromAddress,
+      'donotreply@mail.postman.gov.sg',
+    ]
```
jia1, line 231: Can we centralise this, so that we have a single source of truth?

But I won't block the PR cus of this, because we don't foresee further changes.

KishenKumarrrrr: Okie, I changed it to reuse `isDefaultFromAddress`. I'm pushing to staging now, lemme reverify it still works.

jia1: I verified your initial commit works. I'll also check again once this is up on staging. It's a quick check for me - I use the transactional email API to send 4 emails and also create campaign in UI to send email

jia1: Works for me now too

jia1 (approved):

PR #2276, opened 2024-08-28
chore: remove swagger

jia1:
We already have an up-to-date API guide [here](https://postman-v1.guides.gov.sg/email-api-guide/overview) that is maintained by our product operations team. As such, we will be removing Swagger API docs to keep to a single source of truth.


frontend/package.json
```diff
@@ -35,6 +35,7 @@
     "draft-js-import-html": "^1.4.1",
     "draftjs-filters": "^3.0.1",
     "draftjs-to-html": "0.9.1",
+    "enquirer": "^2.4.1",
```
KishenKumarrrrr, line 38: Can I check what is this for?

KishenKumarrrrr (changes requested):

frontend/package.json
jia1: It was a peer dependency that was blocking CI

jia1: As discussed, will move it to dev dep

KishenKumarrrrr (approved):

PR #2282, opened 2024-11-01
feat: add migration files for index

KishenKumarrrrr:
## Problem

This PR is part of my solution for the aggregation problem. The current explain plan hints at some inefficiencies where we are filtering through many rows in order to derive the result we need. My hypothesis is that this would perform significantly better if we add a proper index to allow efficient filtering of `status` and `error_code`. In this PR, I add the relevant index for this. 

Screenshot 2024-11-01 at 10 00 50 AM

## Deployment Checklist

- [ ] Run migration on `staging`
- [ ] Run migration on `production`


jwnicholas99 (approved):
Thanks @KishenKumarrrrr

PR #2283, opened 2024-11-01
feat: add profiler

KishenKumarrrrr:
## Problem

This PR is a continuation of #2282 . The purpose of turning on profiling is for us to collect some concrete data on what the exact bottleneck is during callbacks so we don't succumb to premature optimisation. 

There is a risk that turning this feature on in production would be detrimental to the application performance. As such, I am only doing so for 20% of all our servers.

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -24,7 +24,7 @@
         "connect-redis": "4.0.4",
         "convict": "6.2.4",
         "cors": "2.8.5",
-        "dd-trace": "^5.1.0",
+        "dd-trace": "^5.24.0",
         "dotenv": "8.2.0",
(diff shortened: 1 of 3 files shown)


jwnicholas99 (approved):
Thanks @KishenKumarrrrr

jwnicholas99:
As discussed in real life, the probability of >30 enabled profiling is 0.6% 

image

jia1 (approved):
LGTM, I don't see any breaking changes that apply to us (pertaining to dd-trace upgrade from v2 to v5). I also do not know of and cannot find a field/config inside `tracer.init({ ... })` that directly configures the sampling rate for profiling, so this is great!

PR #2285, opened 2024-11-20
chore: remove record log and turn on dbm propagation

KishenKumarrrrr:
## Problem

This PR is a continuation of #2283. The profile from the latest callback spike indicates that the logging statement in the callback flow is introducing a significant bottleneck to our processing. This could be explained by the log statement that attempts to log out the entire record that is returned to us by SNS. 

Screenshot 2024-11-20 at 10 35 18 AM

## Solution

In this PR, I avoid logging out the entire record in an attempt to remove the above-mentioned bottleneck. Additionally, I also set `dbmPropagationMode: 'full'` in our tracers so that we can retrieve explain plans of the SQL statements we are executing in the future. Based on the traces, the DB query is the next potential bottleneck in the execution and generating data on its performance will allow us to optimise in the future.

### DIFF
--- a/backend/src/email/utils/callback/parsers/ses.ts
+++ b/backend/src/email/utils/callback/parsers/ses.ts
@@ -225,7 +225,6 @@ const blacklistIfNeeded = async (message: any): Promise<void> => {
 const parseRecord = async (record: SesRecord): Promise<void> => {
   logger.info({
     message: 'Parsing SES callback record',
-    record,
   })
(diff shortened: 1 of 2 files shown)


jia1 (approved):
Thank you, @KishenKumarrrrr. There's a 2nd log statement that is taking up time too. Will you be finding that statement, or just dealing with this one alone is sufficient enough?

Issue #2140, opened 2023-08-01
Have to choose between attachment or CC/BCC, can't be both


changqing-cpf:
**Describe the bug**

In order to send attachment, have to use multipart/form-data. But when using multipart/form-data, there is no way to make cc or bcc an array.

Tried JSON.stringify, got the error message: "cc" must be an array.

Also tried cc[], got the error message: "cc[]" is not allowed

If I use application/json instead, can't send attachment.

**Expected behavior**

Parse multipart/form-data such that cc and bcc can be converted into an array. Alternatively, remove the array requirement for cc and bcc, and allow sending multiple "cc" and "bcc" form fields, like how attachments are implemented.

PR #2286, opened 2024-11-21
Fix/form data cc bcc single recipient

kevinkim-ogp:
## Problem
Unable to send a single CC/BCC recipient when calling the transactional email API with form-data. This appears to be an edge case, as there are no issues when sending multiple CC/BCC recipients using form-data.

```
// how we normally append CC/BCC with more than 1 recipient
// this is accepted
formData.append('cc', 'user@agency.gov.sg')
formData.append('cc', 'admin@agency.gov.sg')

// BUT fails when only 1 recipient, api validation interprets this as a string
// returning error message 'cc must be an array'
formData.append('cc', 'user@agency.gov.sg')


// other alternatives that were tested but not allowed by api
formData.append('cc[]', 'user@agency.gov.sg')
formData.append('cc', ['user@agency.gov.sg'])  // form-data does not accept arrays
formData.append('cc', JSON.stringify(['user@agency.gov.sg']))  // api does not accept stringified arrays
```
Note: Postman requires attachments to be sent via form-data

Related issue highlighted by another user: https://github.com/opengovsg/postmangovsg/issues/2140 

## Solution
Enhance current CC/BCC validation to accept stringified arrays.
This change ensures:
- Compatibility with direct array submissions (existing implementation).
- Compatibility with form-data, where cc can be sent as a stringified JSON array.

### DIFF
--- a/backend/src/email/routes/email-transactional.routes.ts
+++ b/backend/src/email/routes/email-transactional.routes.ts
@@ -21,6 +21,45 @@ export const InitEmailTransactionalRoute = (
   const router = Router({ mergeParams: true })
 
   // Validators
+  const emailArrayValidation = (fieldName: 'cc' | 'bcc') => {
+    return Joi.alternatives().try(
+      // array
+      Joi.array()
+        .unique()
+        .items(
+          Joi.string().trim().email().options({ convert: true }).lowercase()
+        ),
+
+      // stringified array (form-data)
+      Joi.string().custom((value: string) => {
+        let parsed
+        try {
+          parsed = JSON.parse(value)
+        } catch {
+          throw new Error(
+            `${fieldName} must be a valid array or stringified array.`
+          )
+        }
+
+        if (!Array.isArray(parsed)) {
+          throw new Error(`${fieldName} must be a valid stringified array`)
+        }
+        const { value: validatedEmails, error } = Joi.array()
+          .unique()
+          .items(
+            Joi.string().email().trim().lowercase().options({ convert: true })
+          )
+          .validate(parsed)
+
+        if (error) {
+          throw new Error(`${fieldName} ${error.message}`)
+        }
+
+        return validatedEmails
+      })
+    )
+  }
+
   const sendValidator = {
     [Segments.BODY]: Joi.object({
       recipient: Joi.string()
@@ -54,16 +93,8 @@ export const InitEmailTransactionalRoute = (
         .valid(...Object.values(TransactionalEmailClassification))
         .optional(),
       tag: Joi.string().max(255).optional(),
-      cc: Joi.array()
-        .unique()
-        .items(
-          Joi.string().trim().email().options({ convert: true }).lowercase()
-        ),
-      bcc: Joi.array()
-        .unique()
-        .items(
-          Joi.string().trim().email().options({ convert: true }).lowercase()
-        ),
+      cc: emailArrayValidation('cc'),
+      bcc: emailArrayValidation('bcc'),
       disable_tracking: Joi.boolean().default(false),
     }),
   }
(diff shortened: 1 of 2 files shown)


jia1:
Thanks @kevinkim-ogp!

I ran your branch on my local machine and specifying `["example@domain.com"]` works for me.

I'll take a look at the code right now.

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -863,6 +865,161 @@ describe(`${emailTransactionalRoute}/send`, () => {
     ])
   })
 
+  const ccValidTests = [
+    ['cc-recipient@agency.gov.sg', 'cc-recipient-2@agency.gov.sg'],
+    JSON.stringify([
+      'cc-recipient@agency.gov.sg',
+      'cc-recipient-2@agency.gov.sg',
+    ]),
```
jia1, line 873: Are we able to have a test case with just a single cc recipient?

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -863,6 +865,161 @@ describe(`${emailTransactionalRoute}/send`, () => {
...
+
+      expect(res.body).toStrictEqual({
+        code: 'api_validation',
+        message: errMsg,
+      })
+    }
+  )
+
+  test.each(ccInvalidTests)(
+    'Should throw api validation error if cc is not valid array or stringified array - JSON payload',
```
jia1, line 998: Probably copy pasta error. This is a test using form data

backend/src/email/routes/email-transactional.routes.ts
```diff
@@ -21,6 +21,45 @@ export const InitEmailTransactionalRoute = (
...
+          )
+        }
+
+        if (!Array.isArray(parsed)) {
+          throw new Error(`${fieldName} must be a valid stringified array`)
+        }
+        const { value: validatedEmails, error } = Joi.array()
+          .unique()
+          .items(
+            Joi.string().email().trim().lowercase().options({ convert: true })
```
jia1, line 50: `nit:` Could we either abstract or standardise the sequence of the standard validators for consistency?

It was `Joi.string().trim().email().options({ convert: true }).lowercase()` above.

jia1 (changes requested):
Thank you, @kevinkim-ogp! Just need a bit of tidying up on your end and then this should be good to go.

backend/src/email/routes/email-transactional.routes.ts
kevinkim-ogp: good point, have abstracted the email validation

backend/src/email/routes/tests/email-transactional.routes.test.ts
kevinkim-ogp: good catch, added test for single recipient

backend/src/email/routes/tests/email-transactional.routes.test.ts
```diff
@@ -866,14 +866,11 @@ describe(`${emailTransactionalRoute}/send`, () => {
   })
 
   const ccValidTests = [
-    ['cc-recipient@agency.gov.sg', 'cc-recipient-2@agency.gov.sg'],
-    JSON.stringify([
-      'cc-recipient@agency.gov.sg',
-      'cc-recipient-2@agency.gov.sg',
-    ]),
+    [['cc-recipient@agency.gov.sg']],
+    [['cc-recipient@agency.gov.sg', 'cc-recipient-2@agency.gov.sg']],
```
jia1, line 870: I'm confused... why did we change the value of the test parameters to 2d arrays?

kevinkim-ogp: reverted to 3d arrays, apologies for the confusion

jia1 (approved):
LGTM, will let you know once we get this deployed

PR #2287, opened 2024-11-26
feat: custom spans

KishenKumarrrrr:
## Problem

This PR is a continuation of #2285. After previously removing the logging of the entire record objects, the ELB 502s during callback spikes continued to occur. While trying to debug the issue further, we realised that the existing traces in DD did not contain sufficient information.

In this PR, I manually add custom spans to the callback endpoint that we are interested in so that we can hopefully better understand where the bottlenecks in this endpoint lie.

💡 The newly created spans are still a little wonky. My guess is that this is due to the implementation of this endpoint where there are many overlapping asynchronous operations being executed at the same time.

## Before
Screenshot 2024-11-26 at 11 24 16 AM

## After (No DB calls in the SS because this is on staging, we will have them on production)
Screenshot 2024-11-26 at 11 24 33 AM

### DIFF
--- a/backend/src/email/utils/callback/parsers/ses.ts
+++ b/backend/src/email/utils/callback/parsers/ses.ts
@@ -14,6 +14,7 @@ import config from '@core/config'
 import { compareSha256Hash } from '@shared/utils/crypto'
 import { EmailTransactionalService } from '@email/services/email-transactional.service'
 import { SesEventType, Metadata } from '@email/interfaces/callback.interface'
+import { tracer } from 'dd-trace'
 
 const logger = loggerWithLabel(module)
 const REFERENCE_ID_HEADER_V2 = 'X-SMTPAPI' // Case sensitive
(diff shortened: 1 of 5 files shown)


backend/src/email/utils/callback/parsers/ses.ts
```diff
@@ -248,15 +263,21 @@ const parseRecord = async (record: SesRecord): Promise<void> => {
...
+        {
+          timestamp: new Date(record.Timestamp),
+          bounce: message.bounce,
+          complaint: message.complaint,
+          delivery: message.delivery,
+        },
+        parseRecordSpan
+      )
     }
     return parseNotificationAndEvent(type, message, metadata)
```
jia1, line 278: Just confirming - you don't need to trace this

jia1 (approved):
LGTM

backend/src/email/utils/callback/parsers/ses.ts
KishenKumarrrrr: I think I initially removed it because it wasn't working as I expected. I realised thats because it was executing asynchronously so it doesn't have a proper parent span. But I'll just add this in first and we can reason about it when we need to.

PR #2289, opened 2024-11-27
chore: upgrade axios

jia1:


jia1:
Closing because we are only using the type in `shared/` and for `sst/`, interactions are within our network

PR #2299, opened 2025-05-26
fix: guide links

KishenKumarrrrr:
## Problem

In this PR, I update the Postman guide link. 

This is necessary because we are in the process of removing the guide CNAME records because we've mixed up the V1 and V2 guide links. 

⚠️ While I change the guide link in the API key expiry email, I will not be re-deploying the SST job for it.


jwnicholas99 (approved):
LGTM, thanks @KishenKumarrrrr

PR #2302, opened 2025-08-21
feat: remove demo campaigns

KishenKumarrrrr:
## Problem

The recipient limit validation on demo campaigns was discovered to be broken. As such, users could abuse the demo feature to send out their entire campaigns as the number of recipients was not being capped to 20.

## Solution

We've decided to remove the demo feature entirely since agencies are supposed to use postman.gov.sg for sending out SMSes.

### DIFF
--- a/shared/package.json
+++ b/shared/package.json
@@ -44,7 +44,7 @@
   },
   "dependencies": {
     "@types/nodemailer": "6.4.0",
-    "axios": "1.8.4",
+    "axios": "^1.8.4",
     "cheerio": "1.0.0-rc.10",
     "dd-trace": "2.45.1",
     "libphonenumber-js": "1.10.6",
--- a/frontend/src/components/dashboard/demo/demo-bar/DemoBar.tsx
+++ b/frontend/src/components/dashboard/demo/demo-bar/DemoBar.tsx
@@ -1,60 +1,20 @@
 import cx from 'classnames'
 
-import { useContext, useEffect, useState } from 'react'
-
-import CreateDemoModal from '../create-demo-modal'
+import { useEffect, useState } from 'react'
 
 import styles from './DemoBar.module.scss'
 
-import { TextButton } from 'components/common'
-import { ModalContext } from 'contexts/modal.context'
-
 const DemoBar = ({
-  numDemosSms,
-  numDemosTelegram,
   isDisplayed,
 }: {
   numDemosSms: number
   numDemosTelegram: number
   isDisplayed: boolean
 }) => {
-  const modalContext = useContext(ModalContext)
   const [isMenuVisible, setIsMenuVisible] = useState(isDisplayed)
-  const [hasDemo, setHasDemo] = useState(false)
-
-  useEffect(() => {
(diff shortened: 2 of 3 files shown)


shared/package.json
```diff
@@ -44,7 +44,7 @@
   },
   "dependencies": {
     "@types/nodemailer": "6.4.0",
-    "axios": "1.8.4",
+    "axios": "^1.8.4",
```
jwnicholas99, line 47: Is there a reason why we are no longer pinning the version of `axios`?

backend/src/core/routes/campaign.routes.ts
```diff
@@ -30,7 +30,12 @@ const createCampaignValidator = {
       .required(),
     name: Joi.string().max(255).trim().required(),
     protect: Joi.boolean().default(false),
-    demo_message_limit: Joi.number().default(null).min(1).max(20),
+    // NOTE: demo_message_limit is no longer used since demo campaigns were removed from the FE.
+    // Instead of allowing arbitrary values, we restrict it to null or 0 to gracefully fail requests
+    // that still attempt to set this field. The column still exists in the DB, so we can't simply
+    // drop it here without further pruning. If/when we clean up the DB schema, this validation can
+    // be removed entirely.
+    demo_message_limit: Joi.number().default(null).min(0).max(0),
```
jwnicholas99, line 38: nit: is it cleaner if we say that `demo_message_limit` must be null?

jwnicholas99, line 38: Just to clarify, this also works when you duplicate a demo campaign?

jwnicholas99 (approved):
Just have some small nits, but I'm good with blocking the demo messages at the validator level.

backend/src/core/routes/campaign.routes.ts
```diff
@@ -30,7 +30,7 @@ const createCampaignValidator = {
       .required(),
     name: Joi.string().max(255).trim().required(),
     protect: Joi.boolean().default(false),
-    demo_message_limit: Joi.number().default(null).min(1).max(20),
+    demo_message_limit: Joi.number().default(null).min(0).max(0),
```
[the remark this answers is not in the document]
KishenKumarrrrr: Added `.default(null).valid(0)` instead.

backend/src/core/routes/campaign.routes.ts
jwnicholas99: What if you do `default(null).valid(null)`? So only those that are undefined or explicit null are valid?

KishenKumarrrrr: Looks a bit strange but yea that works too. I've made that change.

KishenKumarrrrr: Actually I was mistaken, the demo duplicate in the FE does not use the actual duplication endpoint. It just creates a new demo campaign which ends up getting block due to the validation change.

jwnicholas99: Ok nice!

PR #2353, opened 2026-04-11
fix: patch axios vuln

KishenKumarrrrr:
## Problem

Patch axios vuln


backend/package.json
```diff
@@ -36,7 +36,7 @@
     "@datadog/browser-rum": "^4.43.0",
     "@sentry/node": "5.30.0",
     "async-retry": "1.3.3",
-    "axios": "0.30.0",
+    "axios": "1.15.0",
```
jia1, line 39: This jump is safe?

jia1 (approved):

PR #2356, opened 2026-04-16
refactor: replace findOrCreate with findOrCreateWithTransaction

KishenKumarrrrr:
## Problem

In this PR, I replace the sequelize `findOrCreate` function with our own implementation in order to facilitate a blue-green deployment of our DB for the engine upgrade. 

This change is necessary because DDL statements are incompatible with blue-green deployments and must be removed. The sequelize `findOrCreate` function actually creates a temporary function under the hood before executing the transaction and this is flagged as a DDL statement that causes the deployment to go into a `DEGRADED` state. To avoid this scenario, we will replace this function with our own implementation.

### DIFF
--- a/backend/src/core/utils/find-or-create.ts
+++ b/backend/src/core/utils/find-or-create.ts
@@ -1,0 +1,78 @@
+import {
+  Model,
+  ModelStatic,
+  Transaction,
+  UniqueConstraintError,
+  Sequelize,
+} from 'sequelize'
+
+type WhereOptions<T> = Partial<T> & Record<string, unknown>
+
+interface FindOrCreateOptions<T> {
+  where: WhereOptions<T>
+  defaults?: Partial<T>
+  transaction?: Transaction | null
+}
+
+/**
+ * A safe alternative to Sequelize's findOrCreate that does not use DDL statements.
+ *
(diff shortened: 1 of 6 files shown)


backend/src/core/utils/find-or-create.ts
```diff
@@ -0,0 +1,78 @@
...
+type WhereOptions<T> = Partial<T> & Record<string, unknown>
+
+interface FindOrCreateOptions<T> {
+  where: WhereOptions<T>
+  defaults?: Partial<T>
+  transaction?: Transaction | null
+}
+
+/**
+ * A safe alternative to Sequelize's findOrCreate that does not use DDL statements.
```
jia1, line 18: Was wondering if this was created from scratched or adapted from somewhere else?

Could we rely on sequelize for this change i.e. replace `findOrCreate` with `findCreateFind`?

jia1 (approved):

backend/src/core/utils/find-or-create.ts
jia1: From Kishen: my concern was that the library comment says that it won't work in a transaction and I wasn't sure what weird behaviour this would introduce. Not sure if my decision to create my own version was valid tho. IIRC GO also just created their own version

Image

jia1 (approved):
LGTM

PR #2382, opened 2026-08-11
feat: harden email OTP character set and length

KishenKumarrrrr:
## Problem

In this PR, I harden the email login OTP for Postman by increase the number of characters the OTP uses.

Linear ticket: https://linear.app/ogp/issue/SGC-1719/postman-emails-harden-email-otp-character-set

## Solution

I bumped the OTP from 6 to 8 characters and switched to a proper high-entropy alphabet — `ABCDEFGHJKLMNPQRSTUVWXYZ23456789` in `generateOtp` (`auth.service.ts`). That's a Crockford Base32-style set (32 symbols → ~32^8 ≈ 10^12 keyspace at length 8), and I deliberately dropped the lookalike characters `I`, `O`, `0` and `1` so the code stays easy to read and type when someone copy-pastes it from their email.

A few decisions worth calling out:

- **No user comms.** Officers copy-paste the OTP rather than transcribe it, so a longer alphanumeric code doesn't change their flow. This was aligned on the ticket, so I didn't add any comms or migration.
- **Kept the login copy honest.** I updated the backend Joi validation to `length(8)` and the frontend error message to "enter an 8-character code" so validation and the message the officer sees actually match the new format.
- **Reworded the error modal title** from "Singpass login is unavailable" to the more general "Unable to sign in" — the old title was misleading for a plain OTP-login failure that has nothing to do with Singpass.

**Before**
Screenshot 2026-08-11 at 17 36 29

**After**
Screenshot 2026-08-11 at 17 37 42

## Tests

- Updated the existing `POST /auth/otp` / `POST /auth/login` route tests to expect the 8-character format (email body regex, valid/invalid OTP cases, retry-exceeded case).
- Verified end-to-end that generating an OTP produces an 8-char code from the new alphabet, that the input rejects disallowed characters, and that login succeeds with a valid code.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -203,15 +203,6 @@
         "line_number": 19
       }
     ],
-    "backend/src/core/services/auth.service.ts": [
-      {
-        "type": "Base64 High Entropy String",
-        "filename": "backend/src/core/services/auth.service.ts",
-        "hashed_secret": "f114703480996b273d7e57cbd195b4ab1e70a21b",
-        "is_verified": false,
-        "line_number": 31
-      }
-    ],
     "backend/src/email/services/tests/email-template.service.test.ts": [
       {
         "type": "Secret Keyword",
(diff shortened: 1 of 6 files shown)


moocowdrinksmilk (approved):
lgtm

PR #2383, opened 2026-08-14
fix: upgrade nodemailer to 7.0.11 across backend, shared, worker

KishenKumarrrrr:
## What

Upgrades `nodemailer` to **7.0.11** in all three packages that depend on it:

| Package | Before | After |
|---------|--------|-------|
| `backend` | 6.9.9 | 7.0.11 |
| `shared` | 6.9.9 | 7.0.11 |
| `worker` | 6.6.2 | 7.0.11 |

## Why

Follow-up to the incident review ([SGC-1697](https://linear.app/ogp/issue/SGC-1697/upgrade-nodemailer-in-all-4-products)). Versions below **7.0.7** carry a lenient address-parser flaw that can misroute mail (including OTPs) to unintended recipients. 7.0.11 is the latest 7.x — it patches that CVE plus the subsequent addressparser DoS and `parseDataURI` ReDoS fixes.

Chose 7.0.11 over the latest 9.x deliberately: it satisfies the `≥ 7.0.7` requirement with the smallest breaking-change surface (no v8/v9 majors).

## Scope / risk

- **Only plain SMTP transport is used** — a single `createTransport({ host, port, auth })` in `shared/src/clients/mail-client.class/index.ts`, reused by `backend` and `worker`.
- The **only** breaking change in nodemailer 7.0.0 is the SES-SDK transport (aws-sdk SES → SESv2). This repo does **not** use that transport, so nothing to migrate.
- Node engine floor is unchanged (`>=6.0.0`).
- `@types/nodemailer` left at 6.4.0 — the code only touches stable SMTP types and it typechecks clean against v7.

Per SGC-1697, Postman Email's login validation is already sufficient (Joi `.email()` rejects the quoted-local-part attack at the boundary), so no validation change is needed here — this PR is purely the dependency bump.

## Verification

- `shared` builds clean; `backend` + `worker` `tsc --noEmit` both pass.
- Backend auth route tests: 10/10 pass.
- Runtime check: nodemailer 7.0.11 `sendMail` with the exact option shape the code builds (from/to/subject/replyTo/html/headers/attachments) — envelope correct, addresses parsed, `X-SES-CONFIGURATION-SET` + `List-Unsubscribe` headers preserved.

Ref: SGC-1697

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -39,7 +39,7 @@
         "module-alias": "2.2.2",
         "moment": "^2.29.4",
         "nanoid": "3.3.6",
-        "nodemailer": "6.9.9",
+        "nodemailer": "7.0.11",
         "papaparse": "5.2.0",
         "pg": "8.5.1",
         "pg-connection-string": "2.4.0",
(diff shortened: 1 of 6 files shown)


moocowdrinksmilk (approved):
lgtm

PR #2392, opened 2026-09-15
fix(editor): keep inline styles on linked text in exported HTML

### LINKED ISSUE #2391: Bold (and other inline styles) lost on linked text in exported HTML
Hey! Found that when text has both an inline style and a link, the exported HTML drops the style: the entity tag overwrites the style tags in renderContent (Converter.ts). So bold on part of a linked sentence disappears for the email recipient. Happy to PR a fix if that helps.

soroush5:
renderContent overwrote style tags with entity tags; now both are kept (<a><b>). Verified old drops bold, new keeps it; tsc + eslint clean.

Fixes #2391


soroush5:
Closing as stale, no activity for a while. Happy to rebase and reopen if there is interest. Thanks for your time.

